PR #164, opened 2020-01-28
Browser context tests

### LINKED ISSUE #5: Add browsercontext.spec.js tests

kblok:
closes #5


src/PlaywrightSharp.Abstractions/IFrame.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <summary>
+        /// Executes a function in browser context
+        /// </summary>
+        /// <param name="script">Script to be evaluated in browser context</param>
+        /// <param name="args">Arguments to pass to script</param>
+        /// <remarks>
+        /// TODO
+        /// </remarks>
+        /// <returns>Task that completes result of the script</returns>
+        Task<T> EvaluateAsync<T>(string script, params object[] args);
```
Meir017: maybe we should use the same approach as in puppeteer-sharp and have an `EvaluateFunction` & `EvaluateExpression` methods

src/PlaywrightSharp.Abstractions/PlaywrightSharp.Abstractions.csproj
```diff
@@ -12,4 +12,7 @@
     <Commandlineparameters></Commandlineparameters>
     <RunWithWarnings>true</RunWithWarnings>
   </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
Meir017: maybe we should use `System.Json`?

src/PlaywrightSharp.Abstractions/IFrame.cs
kblok: I had this conversation with the Playwright team https://github.com/microsoft/playwright/issues/710
We are going to try to go to one single method this time.

src/PlaywrightSharp.Abstractions/PlaywrightSharp.Abstractions.csproj
kblok: That's a good one! 
I thought that was .NET Core. It's net standard :)

PR #165, opened 2020-01-29
Dialog tests

### LINKED ISSUE #9: Add dialog.spec.js tests

kblok:
closes #9

### DIFF
--- a/src/PlaywrightSharp.Abstractions/Dialog.cs
+++ b/src/PlaywrightSharp.Abstractions/Dialog.cs
@@ -1,0 +1,51 @@
+﻿using System.Threading.Tasks;
+
+namespace PlaywrightSharp
+{
+    /// <summary>
+    /// <see cref="IDialog"/> objects are dispatched by page via the <see cref="IPage.Dialog"/> event.
+    /// </summary>
+    /// <example>
+    /// An example of using Dialog class:
+    ///<code>
+    ///<![CDATA[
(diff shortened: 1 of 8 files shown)


src/PlaywrightSharp.Abstractions/Dialog.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    /// <example>
+    /// An example of using Dialog class:
+    ///<code>
+    ///<![CDATA[
+    /// Page.Dialog += async (sender, e) =>
+    /// {
+    ///     await e.Dialog.AcceptAsync();
+    /// }
+    /// await Page.EvaluateExpressionAsync("alert('yo');");
```
Meir017: I think it's called `EvaluateAsync` 😉

src/PlaywrightSharp.Abstractions/Dialog.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// ]]>
+    /// </code>
+    /// </example>
+    public interface IDialog
+    {
+        /// <summary>
+        /// Dialog's type, can be one of alert, beforeunload, confirm or prompt.
+        /// </summary>
+        /// <value>The type of the dialog.</value>
+        public DialogType DialogType { get; set; }
```
Meir017: no public...

PR #168, opened 2020-01-31
Mouse tests

### LINKED ISSUE #23: Add mouse.spec.js tests

Meir017:
closes #23


src/PlaywrightSharp.Tests/Mouse/MouseTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            }
+            await Page.Mouse.MoveAsync(100, 100);
+            await Page.EvaluateAsync(@"() => {
+                window.result = [];
+                document.addEventListener('mousemove', event => {
+                    window.result.push([event.clientX, event.clientY]);
+                });
+            }");
+            await Page.Mouse.MoveAsync(200, 300, new MoveOptions { Steps = 5 });
+            var arr = new int[][] { new[] { 1, 2 } };
```
kblok: We don't need this variable

src/PlaywrightSharp.Tests/Mouse/MouseTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            Assert.Equal(new int[][]
+                { new[] { 120, 140 },
+                 new[] { 140, 180 },
+                 new[] { 160, 220 },
+                 new[] { 180, 260 },
+                 new[] { 200, 300 }
+            }, await Page.EvaluateAsync<int[][]>("result"));
```
kblok:
```cs
Assert.Equal(
    new int[][]
    { 
        new[] { 120, 140 },
        new[] { 140, 180 },
        new[] { 160, 220 },
        new[] { 180, 260 },
        new[] { 200, 300 }
    }, 
    await Page.EvaluateAsync<int[][]>("result"));
```

src/PlaywrightSharp.Tests/TestConstants.cs
```diff
@@ -29,6 +29,7 @@ internal static class TestConstants
         internal static bool IsWebKit = false;
         internal static bool IsFirefox = false;
         internal static bool IsChromium = true;
+        internal static bool IsMac = false;
```
kblok:
```suggestion
        internal static bool IsMacOSX = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
```

PR #172, opened 2020-02-01
emulate tests

### LINKED ISSUE #12: Add emulation.spec.js tests

kblok:
closes #12

### DIFF
--- a/src/PlaywrightSharp.Abstractions/IPage.cs
+++ b/src/PlaywrightSharp.Abstractions/IPage.cs
@@ -29,6 +29,13 @@ namespace PlaywrightSharp
         int DefaultTimeout { get; set; }
 
         /// <summary>
+        /// Setup media emulation
+        /// </summary>
+        /// <param name="options"></param>
+        /// <returns>A <see cref="Task"/> that completes when the message is confirmed by the browser.</returns>
+        Task EmulateMediaAsync(EmulateMedia options);
+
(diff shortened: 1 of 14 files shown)


src/PlaywrightSharp.Abstractions/IPage.cs
```diff
@@ -144,8 +151,16 @@ public interface IPage
...
-
+        event EventHandler<ConsoleEventArgs> Console;
+
+        /// <summary>
+        /// Waits for event to fire and passes its value into the predicate function.
+        /// </summary>
+        /// <param name="e">Event to wait for</param>
+        /// <param name="options">Extra options</param>
+        /// <returns>A <see cref="Task"/> that completes when the predicate returns truthy value. Yielding the information of the event.</returns>
+        Task<T> WaitForEvent<T>(PageEvent e, WaitForEventOptions options = null);
```
kblok: @Meir017 do you have a better/generic idea for this?
It sucks that we can't make extension methods over events.

Meir017: Maybe instead we could expose a generic event property?
```cs
event EventHandler<PageEventArgs> Event;
```
where `PageEventArgs` will be something like
```cs
public class PageEventArgs
{
    public string PageEventName { get; } // enum
    public object Data { get; } // can by anything...
}
```

PR #179, opened 2020-02-03
Navigation tests

### LINKED ISSUE #24: Add navigation.spec.js tests

kblok:
closes #24


src/PlaywrightSharp.Abstractions/IPage.cs
```diff
@@ -116,23 +137,39 @@ public interface IPage
...
-        /// - <see cref="WaitForNavigationAsync(NavigationOptions)"/>
+        /// - <see cref="WaitForNavigationAsync(WaitForNavigationOptions)"/>
         /// **NOTE** <see cref="DefaultNavigationTimeout"/> takes priority over <seealso cref="DefaultTimeout"/>.
         /// </summary>
-        int DefaultNavigationTimeout { get; set; }
-
+        int DefaultNavigationTimeout { get; set; }
+
+        /// <summary>
+        /// This completes when the page reaches a required load state, load by default.
```
Meir017:
```suggestion
        /// Completes when the page reaches a required load state, load by default.
```

src/PlaywrightSharp.Abstractions/WaitForNavigationOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Wait for this specific URL.
+        /// </summary>
+        public string Url { get; set; }
+
+        /// <summary>
+        /// Wait for an URL matching this expression.
+        /// </summary>
+        public string UrlRegEx { get; set; }
```
Meir017: shouldn't this be included in `UrlPredicate`?

kblok: Both `Url` and `UrlRegEx` will end up being a predicate, internally.

PR #195, opened 2020-02-09
BrowserFetcher in Chromium

kblok:
progress on #68


src/PlaywrightSharp.Chromium/ChromiumBrowserType.cs
```diff
@@ -1,17 +1,30 @@
...
+using System.IO;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 
 namespace PlaywrightSharp.Chromium
 {
     /// <inheritdoc cref="IBrowserType"/>
     public class ChromiumBrowserType : IBrowserType
     {
+        private readonly string _revision = "733125";
```
Meir017:
```suggestion
        private static readonly string _revision = "733125";
```

src/PlaywrightSharp.Chromium/ChromiumBrowserType.cs
```diff
@@ -25,9 +38,49 @@ public Task<IBrowser> ConnectAsync(ConnectOptions options = null)
...
+
+            var path = options?.Path ?? Path.Combine(Directory.GetCurrentDirectory(), ".local-chromium");
+            string host = options?.Host ?? "https://storage.googleapis.com";
+            var platform = options.Platform ?? GetPlatform();
+
+            Func<Platform, string, BrowserFetcherConfig> paramsGetter = (platform, revision) =>
+            {
+                var archiveName = string.Empty;
+                var executablePath = string.Empty;
+                if (platform == Platform.Linux)
```
Meir017: could we use here the new C# 8 switch expression?

src/PlaywrightSharp.Chromium/ChromiumBrowserType.cs
kblok: Playwright has that as an instance prop but I think we can make it static due to our code nature.

src/PlaywrightSharp.Chromium/ChromiumBrowserType.cs
kblok: Converted to a switch, but I have many lines to turn them into expressions.

PR #197, opened 2020-02-11
JsHandle tests

### LINKED ISSUE #20: Add jshandle.spec.js tests

Meir017:
closes #20


kblok:
Let's compare this PR with this code https://github.com/microsoft/playwright/blob/v0.10.0/test/jshandle.spec.js

kblok:
@Meir017 one code factor and we ready to go

PR #223, opened 2020-02-13
add keyboard.spec

### LINKED ISSUE #21: Add keyboard.spec.js tests

Meir017:
closes #21


src/PlaywrightSharp.Abstractions/KeyDefinitions.cs
```diff
@@ -0,0 +1,9 @@
+namespace PlaywrightSharp
+{
+    /// <summary>
+    /// Utility class for getting KeyDefinition.
+    /// </summary>
+    public static class KeyDefinitions
+    {
```
kblok: Are we able to populate this class now?

src/PlaywrightSharp.Abstractions/PressOptions.cs
```diff
@@ -0,0 +1,15 @@
+namespace PlaywrightSharp
+{
+    /// <summary>
+    /// options to use when pressing a key.
```
kblok:
```suggestion
    /// Options to use when pressing a key.

```

src/PlaywrightSharp.Abstractions/KeyDefinitions.cs
Meir017: yes :)

PR #225, opened 2020-02-13
Mark tests to run in CI

kblok:


kblok:
What do you think about this @Meir017?
This would help us having green builds and detect regressions. If everything is red until we finish we won't be able to see if we broke something.

The idea would be adding categories to the tests we start implementing.

Meir017:
sounds pretty good for now

PR #228, opened 2020-02-16
add page tests

### LINKED ISSUE #26: Add page.spec.js tests

Meir017:
closes #26


src/PlaywrightSharp.Abstractions/IPage.cs
```diff
@@ -264,9 +276,34 @@ public interface IPage : IDisposable
...
         /// <returns>A <see cref="Task"/> that completes when the predicate returns truthy value. Yielding the information of the event.</returns>
         Task<T> WaitForEvent<T>(PageEvent e, WaitForEventOptions options = null);
 
+        /// <summary>
+        /// Waits for event to fire once and return its value.
+        /// </summary>
+        /// <typeparam name="T">Return type.</typeparam>
+        /// <param name="e">Event to wait for.</param>
+        /// <returns>A <see cref="Task"/> that completes when the predicate returns truthy value. Yielding the information of the event.</returns>
+        Task<T> OnceAsync<T>(PageEvent e);
```
Meir017: @kblok since we have the `WaitForEvent` does this make sense?

src/PlaywrightSharp.Abstractions/WaitForClickOption.cs
```diff
@@ -3,7 +3,7 @@ namespace PlaywrightSharp
     /// <summary>
     /// Options for <see cref="ClickOptions.WaitFor"/>.
     /// </summary>
-    public enum WaitForClickOption
+    public enum WaitForOption
```
kblok: We should rename the file.

src/PlaywrightSharp.Tests/Page/AsyncStacksTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        [Fact]
+        public async Task ShouldWork()
+        {
+            Server.SetRoute("/empty.html", context =>
+            {
+                context.Abort(); // is this right?
+                return Task.CompletedTask;
+            });
+            var exception = await Assert.ThrowsAsync<PlaywrightSharpException>(() => Page.GoToAsync(TestConstants.EmptyPage));
+            // TODO: is this right?
```
kblok: I wouldn't worry about this right now. We'll get there when we implement this.

src/PlaywrightSharp.Tests/Page/Events/PageEventsErrorTests.cs
```diff
@@ -0,0 +1,37 @@
...
+            string error = null;
+            Page.Error += (sender, e) => error = e.Error;
+            if (TestConstants.IsChromium)
+            {
+                _ = Page.GoToAsync("chrome://crash").ContinueWith(t => { });
+            }
+            else if (TestConstants.IsWebKit)
+            {
+                // TODO: expose PageDelegate
+                // Page._delegate._session.send('Page.crash', }
```
kblok: Let's throw an assert fail.

src/PlaywrightSharp.Abstractions/IPage.cs
kblok: This shouldn't be part of the API.

src/PlaywrightSharp.Tests/Page/Events/PageEventsPopupTests.cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+        }
+
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.Events.Popup</playwright-describe>
+        ///<playwright-it>should work</playwright-it>
+        [Fact]
+        public async Task ShouldWork()
+        {
+            var popupTask = Page.OnceAsync<PopupEventArgs>(PageEvent.Popup);
```
kblok: See this pattern to implement "once"s https://github.com/kblok/puppeteer-sharp/blob/74488cf1760b2d4895a4855e9be87cb50ec3c9d8/lib/PuppeteerSharp.Tests/NetworkTests/SetRequestInterceptionTests.cs#L151

src/PlaywrightSharp.Tests/Page/PageOpenerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+        }
+
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.opener</playwright-describe>
+        ///<playwright-it>should provide access to the opener page</playwright-it>
+        [Fact]
+        public async Task ShouldProvideAccessToTheOpenerPage()
+        {
+            var (popupEvent, _) = await TaskUtils.WhenAll(
```
kblok: WhenAll doesn't return all the tasks as an array like Promise.all in javascript.

src/PlaywrightSharp.Tests/Page/PageOpenerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            Assert.Equal(Page, opener);
+        }
+
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.opener</playwright-describe>
+        ///<playwright-it>should return null if parent page has been closed</playwright-it>
+        [Fact]
+        public async Task ShouldReturnNullIfParentPageHasBeenClosed()
+        {
+            var (popupEvent, _) = await TaskUtils.WhenAll(
```
kblok: Same

src/PlaywrightSharp.Tests/Page/Events/PageEventsConsoleTests.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.Events.Console</playwright-describe>
+        ///<playwright-it>should have location for console API calls</playwright-it>
+        [Fact]
+        public async Task ShouldHaveLocationForConsoleAPICalls()
+        {
+            await Page.GoToAsync(TestConstants.EmptyPage);
+            var (messageEvent, _) = await TaskUtils.WhenAll(
```
kblok: Let's review all the WhenAlls

src/PlaywrightSharp.Tests/Page/PageWaitForRequestTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.waitForRequest</playwright-describe>
+        ///<playwright-it>should respect timeout</playwright-it>
+        [Fact]
+        public async Task ShouldRespectTimeout()
+        {
+            var exception = await Assert.ThrowsAsync<TimeoutException>(
+                () => Page.WaitForEvent(PageEvent.Request, new WaitForEventOptions<RequestEventArgs>
+                {
+                    Predicate = _ => false,
```
kblok: Can we avoid this?

src/PlaywrightSharp.Tests/Page/PageWaitForResponseTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.waitForResponse</playwright-describe>
+        ///<playwright-it>should respect timeout</playwright-it>
+        [Fact]
+        public async Task ShouldRespectTimeout()
+        {
+            var exception = await Assert.ThrowsAsync<TimeoutException>(
+                () => Page.WaitForEvent(PageEvent.Response, new WaitForEventOptions<ResponseEventArgs>
+                {
+                    Predicate = _ => false,
```
kblok: Same

src/PlaywrightSharp.Tests/Page/PageWaitForResponseTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.waitForResponse</playwright-describe>
+        ///<playwright-it>should respect default timeout</playwright-it>
+        [Fact]
+        public async Task ShouldRespectDefaultTimeout()
+        {
+            var exception = await Assert.ThrowsAsync<TimeoutException>(
+                () => Page.WaitForEvent(PageEvent.Response, new WaitForEventOptions<ResponseEventArgs>
+                {
+                    Predicate = _ => false
```
kblok: Same

src/PlaywrightSharp.Tests/Page/PageBrowserContextTests.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+        }
+
+        ///<playwright-file>page.spec.js</playwright-file>
+        ///<playwright-describe>Page.browserContext</playwright-describe>
+        ///<playwright-it>should return the correct browser instance</playwright-it>
+        [Fact]
+        public void ShouldReturnTheCorrectBrowserInstance()
+        {
+            Assert.Equal(Context, Page.BrowserContext);
```
kblok: Arrow

src/PlaywrightSharp.Tests/TaskUtils.cs
```diff
@@ -0,0 +1,21 @@
+using System.Threading.Tasks;
+
+namespace PlaywrightSharp.Tests
+{
+    internal static class TaskUtils
+    {
+        public static async Task<(T1, T2)> WhenAll<T1, T2>(Task<T1> task1, Task<T2> task2)
```
kblok: aaaaaaahhhhhh cleeeeever <3

src/PlaywrightSharp.Tests/Page/Events/PageEventsPopupTests.cs
Meir017: I was thinking of adding it since we added `WaitForEvent` and they both work in a similar way and use the same enum

src/PlaywrightSharp.Tests/Page/PageWaitForRequestTests.cs
Meir017: which part? creating an instance of `WaitForEventOptions` or the predicate that evaluates to `false`?

src/PlaywrightSharp.Tests/TaskUtils.cs
Meir017: 😀

src/PlaywrightSharp.Tests/Page/PageOpenerTests.cs
Meir017: is this relevant? we are returning a tuple here after calling `Task.WhenAll` on all of the tasks

kblok:
could you run dotnet format @Meir017?

src/PlaywrightSharp.Tests/Page/Events/PageEventsPopupTests.cs
kblok: `once` is part of the `EventEmitter` class from node, like `on`.  Which are our events.

I think that here https://github.com/microsoft/playwright/blob/v0.10.0/test/page.spec.js#L128 they use `once` as a `WaitForEvent`. We can replace that `once` with a `WaitForEvent`.

But here https://github.com/microsoft/playwright/blob/v0.10.0/test/page.spec.js#L212 they use once by the book. once attaches to the event and deataches after the first event trigger.

So, it's not part of the API, but we can use `WaitForEvent` when then wrap once inside a promise.

src/PlaywrightSharp.Tests/Page/PageOpenerTests.cs
kblok: not relevant.

src/PlaywrightSharp.Tests/Page/PageWaitForRequestTests.cs
kblok: The predicate evaluating to false. We could check for null in the implementation.

src/PlaywrightSharp.Tests/Page/Events/PageEventsPopupTests.cs
Meir017: done

src/PlaywrightSharp.Tests/Page/PageWaitForRequestTests.cs
Meir017: if it evaluates to `false` then it means that the Task never completes so checking for null won't help up because if we don't pass that property in then it acts as if it was `_ => true` and waits for any event

kblok: Got it, great job man!

PR #231, opened 2020-02-18
Chromium - Implement connect tests

kblok:
Related to #68


kblok:
@Meir017 it seems that we need to connect lots of dots just to get a single test green. Would you mind start taking a look at this guy while I make some progress?

Meir017:
No problem 😀

kblok:
We are green here @Meir017 !

src/PlaywrightSharp.Chromium/Helpers/JsonHelper.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace PlaywrightSharp.Chromium.Helpers
+{
+    internal static class JsonHelper
+    {
+        public static JsonSerializerOptions DefaultChromiumJsonSerializerOptions
```
Meir017: this could be a singleton. no need to create a new instance on every call

src/PlaywrightSharp.Chromium/Helpers/JsonExtensions.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.IO;
+using System.Text.Json;
+
+namespace PlaywrightSharp.Chromium.Helpers
+{
+    internal static partial class JsonExtensions
+    {
+        public static T ToObject<T>(this JsonElement element, JsonSerializerOptions options = null)
```
Meir017: I'm not sure this is chromium specific

kblok: Yeah, I moved the class where because the other extension uses the `DefaultChromiumJsonSerializerOptions` but we can have both classes.

PR #232, opened 2020-02-19
Protocol classes generation

### LINKED ISSUE #229: Generate protocol classes
https://github.com/kblok/playwright/blob/v0.10.0/utils/protocol-types-generator/index.js

Meir017:
closes #229


Meir017:
@kblok I might need to move the browser fetcher implementations out of the `Playwright.<BrowserType>` in order to use it in this console app. webkit & firefox generate the protocol classes from a local file that comes with the browser

src/PlaywrightSharp.sln
```diff
@@ -30,6 +30,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "PlaywrightSharp", "Playwrig
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "PlaywrightSharp.Abstractions", "PlaywrightSharp.Abstractions\PlaywrightSharp.Abstractions.csproj", "{FD41DFE2-6C0B-4717-BC60-7EC5EA224FDA}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PlaywrightSharp.ProtocolTypesGenerator", "PlaywightSharp.ProtocolTypesGenerator\PlaywrightSharp.ProtocolTypesGenerator.csproj", "{1E580718-B961-4E14-9F5A-A74BC54EB0C7}"
+EndProject
```
kblok: Our solution has a "Tools" folder, Are we putting this there?

kblok:
>@kblok I might need to move the browser fetcher implementations out of the Playwright.<BrowserType> in order to use it in this console app. webkit & firefox generate the protocol classes from a local file that comes with the browser

Why? because it's internal? if so, we can use "internals visible to".

Meir017:
The `CDPSession` class in chromium and the other ones in the other browsers could now have a strongly-typed API without any string arguments but for that we will need to run the generator as part of the build process but if the BrowserFetcher class in in the browser projects then we'll have a two-way dependency... 🤔

kblok:
@Meir017 I don't see this as something we need to run every time. I picture this more as the Devices generator you did.

Why do you need the CDPSession?

Meir017:
https://gist.github.com/Meir017/36be9e728c4784c0d2192318d8499ca8
starting to look good 😄

Meir017:
> @Meir017 I don't see this as something we need to run every time. I picture this more as the Devices generator you did.
> 
> Why do you need the CDPSession?

I was thinking that the CDPSession will depend on the generated types which is why the browserfetcher cannot be in the same csproject

kblok:
I took a look at pw code. I didn't realize that they were using the app to generate types. Maybe we could give some priority to Page Goto and Page Evaluate tests, and punt this PR till then.

Meir017:
@kblok might be overkill but it could be really cook
my idea is to have the `ChromiumSession` class have a single method
```cs
    public class ChromiumSession
    {
        public Task<TResponse> SendAsync<TRequest, TResponse>(TRequest request)
            where TRequest : ChromiumRequest<TResponse>
            where TResponse : ChromiumResponse
        {
            return null;
        }
    }

    public abstract class ChromiumRequest<TResponse> where TResponse : ChromiumResponse {}
    public abstract class ChromiumResponse { }
```

then using it will be as simple as:
```cs
ChromiumSession session = ...
var response = await session.SendAsync(new PageNavigateRequest());
// response is PageNavigateResponse
```

but for some reason the compiler does like this :( 
CS0411 C# The type arguments for method cannot be inferred from the usage. Try specifying the type arguments explicitly. 🤦‍♂ 

then you will always have IntelliSense without the constant string that represent the command

kblok:
I like the idea of getting rid of the string with the method name. But, we have many SendAsync with no arguments, that would require creating a class just to hold the method name.

C# will try to infer generics base on the arguments you pass to the function. If there is not TResponse in the argument list, it won't work.

Meir017:
too bad. I thought that having the request class have the TResponse in it's signature will be enough

kblok:
Just speaking my mind...
But this is too much boilerplate. 

```
response = client.SendAsync(new ChromiumMessage<PageNavigateRequest, PageNavigateResponse> 
    { 
        Request = new PageNavigateRequest 
        { 
            SomeProp = true
        }
    }); 
```

Meir017:
@kblok I found the solution

```cs
public class ChromiumSession
    {
        public Task<TResponse> SendAsync<TResponse>(ChromiumRequest<TResponse> request) where TResponse : ChromiumResponse
        {
            return null;
        }
    }

    public abstract class ChromiumRequest<TResponse> where TResponse : ChromiumResponse { }
    public abstract class ChromiumResponse { }

    public class PageNavigateRequest : ChromiumRequest<PageNavigateResponse>
    {
    }
    public class PageNavigateResponse : ChromiumResponse { }

    public class Demo
    {
        public static async Task Run(ChromiumSession session)
        {
            var response = await session.SendAsync(new PageNavigateRequest());
            // response is PageNavigateResponse
        }
    }
```

based on https://github.com/jbogard/MediatR/blob/master/src/MediatR/Mediator.cs#L29

kblok:
And `ChromiumRequest` would have an abstract property with the method name?

Meir017:
something like that

kblok:
@Meir017 I didn't check if you are already doing that. But I'd love to treat the first part of the method name as the namespace and folder. So PageNavigateResponse would be inside the Page folder under the Page namespace.

Meir017:
I thought that because it's generated code it could all sit in the same file called `Protocol.Generated.cs` or something like that
The types do start with the namespace

kblok:
@Meir017 I buy it.

kblok:
@Meir017 I'm wondering about 2 things.

* Are we going to have these kinds of classes https://github.com/kblok/playwright-sharp/pull/231/files#diff-ba22ad99a12fe9b9b6a0c0c29915a257R3 ?
* What if need to code some enum or something like that. How are we going to solve it?

Meir017:
with the events we should be able to write something like:
```cs
ChromiumSession session = ...
var request = await sesstion.WaitForEvent<NetworkRequestWillBeSentEventArgs>();
// request is NetworkRequestWillBeSentEventArgs
```

kblok:
I wouldn't use event args to wait for event, that wont be clear users, and we are going to have many EventArgs events. We could build a class with read-only string properties like `public string Load => ”Load”`.  We could implement WaitForEvent quite easily with that.

src/PlaywrightSharp.sln
Meir017: yes :)

Meir017:
@kblok I think this should wait for the #231 so I could test it to be sure it works

src/PlaywrightSharp.Chromium/Protocol/ChromiumEvent.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// Basic class for chromium events.
+    /// </summary>
+    public abstract class ChromiumEvent
+    {
+        /// <summary>
+        /// Gets the name of the event recieved from chromium devtools protocol.
+        /// </summary>
+        public abstract string InternalName { get; }
```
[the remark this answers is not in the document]
Meir017: I tried `Name` & `Event` but there are events with properties using that name. this is so we could deserialize automatically the event to the correct event type

Meir017:
> @Meir017 can we create an events class? something like this
> 
> 
> ```
> public class ChromiumEvents
> {
>     public string Load => "Load";
> }
> ```
> 
> This would work for `WaitForEvent`.
> 
> Let's make classes internal.

not sure how this would help since there isn't an event type and it's just a string so there isn't any enforcement on it

kblok:
> not sure how this would help since there isn't an event type and it's just a string so there isn't any enforcement on it

It will help the developer so they don’t guess the event name when they use WaitForEvent.

src/PlaywrightSharp.Chromium/Protocol/ChromiumEvent.cs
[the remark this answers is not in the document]
kblok: Wouldn't I need to know the type before instantiating a ChromoumEvent?

src/PlaywightSharp.ProtocolTypesGenerator/ChromiumProtocolTypesGenerator.cs
```diff
@@ -0,0 +1,381 @@
...
+            string directory = Path.Join("..", "..", "..", "..", "PlaywrightSharp.Chromium", "Protocol");
+            string output = Path.Join(directory, "Protocol.Generated.cs");
+            if (!Directory.Exists(directory))
+            {
+                Directory.CreateDirectory(directory);
+            }
+
+            if (revision.Local && File.Exists(output))
+            {
+                // return;
```
kblok: Are we going to return for real here?

src/PlaywightSharp.ProtocolTypesGenerator/PlaywrightSharp.ProtocolTypesGenerator.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
kblok: Let's add our analyzers here as well. This deserves our love too!

src/PlaywightSharp.ProtocolTypesGenerator/ChromiumProtocolTypesGenerator.cs
```diff
@@ -0,0 +1,381 @@
...
+
+            // work around, too lazy to solve this
+            knownTypes["Headers"] = "System.Collections.Generic.IDictionary<string, string>";
+            knownTypes["ArrayOfStrings"] = "int[]";
+            knownTypes["StringIndex"] = "int";
+        }
+
+        private void GenerateTypes(StringBuilder builder, ChromiumProtocolDomain domain)
+        {
+            if (domain.Types != null)
```
kblok: brackets

src/PlaywrightSharp.Chromium/ChromiumBrowser.cs
```diff
@@ -145,15 +145,15 @@ private async void Session_MessageReceived(object sender, MessageEventArgs e)
                 switch (e.MessageID)
                 {
                     case "Target.targetCreated":
-                        await CreateTargetAsync(e.MessageData?.ToObject<TargetCreatedResponse>()).ConfigureAwait(false);
+                        await CreateTargetAsync(e.MessageData?.ToObject<TargetTargetCreatedEventArgs>()).ConfigureAwait(false);
```
kblok: I don't like the `EventArgs` suffix for these classes.

Meir017: hmm, these classes should be used instead in the event handler.
We shouldn't need to create POCO objects, maybe 1-2 and that's it. everything else is covered in the generated protocol classes

kblok: Our events don't necessarily match browser events, in fact, those are at the Abstraction layer, which means that our events are cross-browser events.
e.g. [FileChooserEventArgs](https://github.com/kblok/playwright-sharp/blob/master/src/PlaywrightSharp.Abstractions/FileChooserEventArgs.cs#L8) that's a generic, cross-browser event we need to build. It's **cross-browser**, I think is a key concept. 

In this example, this is clearly not an event args. It's not an event we get from an event handler and it's not an argument we are going to send using an event handler.
Chromium might have called it event, but for us, it's just a message.

Meir017: Ahh, Makes perfect sense.
So maybe changing the suffix of the chromium event classes to `ChromiumEvent` is better?

kblok: Yeah let's do it!

Meir017:
I'm working on getting rid of `MessageEventArgs`. 
I want something like
```cs
switch (e.MessageID)
{
    case "Target.targetCreated":
        await CreateTargetAsync(e.MessageData?.ToObject<TargetTargetCreatedChromiumEvent>()).ConfigureAwait(false);
        return;

    case "Target.targetDestroyed":
        await DestroyTargetAsync(e.MessageData?.ToObject<TargetTargetDestroyedChromiumEvent>()).ConfigureAwait(false);
        return;

    case "Target.targetInfoChanged":
        ChangeTargetInfo(e.MessageData?.ToObject<TargetTargetCreatedChromiumEvent>());
        return;
}
```

to become something like this:
```cs
switch (e)
{
    case TargetTargetCreatedChromiumEvent targetCreated:
        await CreateTargetAsync(targetCreated).ConfigureAwait(false);
        break;
    case TargetTargetDestroyedChromiumEvent targetDestroyed:
        await DestroyTargetAsync(targetDestroyed);
        break;
    case TargetTargetInfoChangedChromiumEvent targetInfoChanged:
        ChangeTargetInfo(targetInfoChanged);
        break;
}
```

src/PlaywrightSharp.Chromium/Protocol/DeserializerHelper.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace PlaywrightSharp.Chromium.Protocol
+{
+    internal static class DeserializerHelper
```
Meir017: @kblok  this is the magic that causes the deserialization to work. something similar will be done for the event responses

src/PlaywrightSharp.Chromium/ChromiumBrowserType.cs
```diff
@@ -202,7 +202,7 @@ public async Task<IBrowserApp> LaunchBrowserAppAsync(LaunchOptions options = nul
                     }
 
                     var transport = await BrowserHelper.CreateTransportAsync(browserApp.ConnectOptions).ConfigureAwait(false);
-                    await transport.SendAsync("Browser.close", new BrowserCloseRequest { Id = ChromiumConnection.BrowserCloseMessageId }).ConfigureAwait(false);
+                    await transport.SendAsync(new BrowserCloseRequest().Command).ConfigureAwait(false);
```
kblok: Can't we make `Command` static? you could have one static and one instance for the interface.

src/PlaywrightSharp.Chromium/ChromiumPage.cs
```diff
@@ -112,27 +121,32 @@ internal async Task InitializeAsync()
 
                 if (options.Geolocation != null)
                 {
-                    tasks.Add(_client.SendAsync("Emulation.setGeolocationOverride", options.Geolocation));
+                    tasks.Add(_client.SendAsync(new EmulationSetGeolocationOverrideRequest
+                    {
+                        Accuracy = (double)options.Geolocation.Accuracy,
```
kblok: Can't we match types here?

src/PlaywrightSharp.Chromium/ChromiumPage.cs
```diff
@@ -163,17 +177,36 @@ private void OnLifecycleEvent(PageLifecycleEventResponse e)
...
+            })).ToArray()).ConfigureAwait(false);
+        }
 
         private Task EmulateTimezoneAsync(string timezoneId)
         {
             throw new NotImplementedException();
         }
 
-        private void HandleFrameTree(PageGetFrameTreeItem frameTree)
+        private void HandleFrameTree(Protocol.Page.FrameTree frameTree)
```
kblok: Can't we add usings?

src/PlaywrightSharp.Chromium/Helpers/JsonExtensions.cs
```diff
@@ -9,5 +9,8 @@ internal static partial class JsonExtensions
     {
         public static T ToObject<T>(this JsonElement element)
             => JsonSerializer.Deserialize<T>(element.GetRawText(), JsonHelper.DefaultChromiumJsonSerializerOptions);
+
+        public static object ToObject(this JsonElement element, Type type)
```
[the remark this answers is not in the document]
Meir017: for dynamic deserialization when the type isn't known in advance

src/PlaywrightSharp.Chromium/PlaywrightSharp.Chromium.csproj
```diff
@@ -1,26 +1,26 @@
...
-        <ProjectReference Include="..\PlaywrightSharp\PlaywrightSharp.csproj" />
-        <ProjectReference Include="..\PlaywrightSharp.Abstractions\PlaywrightSharp.Abstractions.csproj" />
-    </ItemGroup>
-    <Import Project="../Common/Dependencies.props" />
-    <ItemGroup>
-        <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
-            <_Parameter1>PlaywrightSharp.Tests</_Parameter1>
-        </AssemblyAttribute>
-    </ItemGroup>
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
Meir017: I think I ran `dotnet-format` and committed this file...

src/PlaywrightSharp.Chromium/Protocol/ChromiumProtocolTypes.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text.Json;
+
+namespace PlaywrightSharp.Chromium.Protocol
+{
+    internal static class ChromiumProtocolTypes
+    {
+        private static readonly IDictionary<string, Type> ChromiumResponseMapper = new Dictionary<string, Type>();
+        private static readonly IDictionary<string, Type> ChromiumEventsMapper = new Dictionary<string, Type>();
+
+        static ChromiumProtocolTypes()
```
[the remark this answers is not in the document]
Meir017: it's a static constructur...
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors#remarks

src/PlaywrightSharp.Chromium/ChromiumPage.cs
Meir017: makes sense
I think `double` is good enough here
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types

src/PlaywrightSharp.Chromium/ChromiumBrowserType.cs
Meir017: I'm pretty sure this is the only place we send a message directly to the transport abstraction. I think that only if we have another case like this then it might be a good enough excuse to have a static property

src/PlaywrightSharp.Chromium/ChromiumSession.cs
```diff
@@ -73,7 +69,8 @@ internal async Task<T> SendAsync<T>(string method, object args = null)
                 }
             }
 
-            return waitForCallback ? await callback.TaskWrapper.Task.ConfigureAwait(false) : null;
+            var result = waitForCallback ? (await callback.TaskWrapper.Task.ConfigureAwait(false)) : null;
+            return (TChromiumResponse)result;
```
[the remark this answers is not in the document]
Meir017: result is `IChromiumResponse` not `TChromiumResponse`

src/PlaywrightSharp.Chromium/Protocol/ChromiumProtocolTypes.cs
```diff
@@ -0,0 +1,50 @@
...
+using PlaywrightSharp.Chromium.Helpers;
+
+namespace PlaywrightSharp.Chromium.Protocol
+{
+    internal static class ChromiumProtocolTypes
+    {
+        private static readonly IDictionary<string, Type> ChromiumResponseMapper = new Dictionary<string, Type>();
+        private static readonly IDictionary<string, Type> ChromiumEventsMapper = new Dictionary<string, Type>();
+
+        static ChromiumProtocolTypes()
```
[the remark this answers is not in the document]
Meir017: this is how static constructors work. no access modifiers...

src/PlaywightSharp.ProtocolTypesGenerator/ChromiumProtocolTypesGenerator.cs
```diff
@@ -0,0 +1,401 @@
...
+            {
+                return property.Type switch
+                {
+                    "array" => this.ConvertItemsProperty(property.Items),
+                    _ => this.ConvertJsToCsharp(property.Type)
+                };
+            }
+            catch (Exception e)
+            {
+                throw e;
```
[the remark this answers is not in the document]
Meir017: it was for debugging...

src/PlaywrightSharp.Chromium/ChromiumPage.cs
```diff
@@ -4,26 +4,35 @@
...
 
+        private readonly ISet<string> _isolatedWorlds;
+
         public ChromiumPage(ChromiumSession client, ChromiumBrowser browser, IBrowserContext browserContext)
         {
             _client = client;
             _browser = browser;
             _browserContext = browserContext;
             _networkManager = new ChromiumNetworkManager(_client, this);
+            _isolatedWorlds = new HashSet<string>();
```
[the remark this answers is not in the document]
Meir017: this is the same way they do it in playwright. I don't this it's expensive enough to worry about

src/PlaywrightSharp.Chromium/Messaging/ConnectionResponse.cs
```diff
@@ -12,7 +12,7 @@ internal class ConnectionResponse
 
         public string Method { get; set; }
 
-        public JsonElement? Params { get; set; }
+        public JsonElement Params { get; set; }
```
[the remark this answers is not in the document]
kblok: @Meir017 ?

src/PlaywightSharp.ProtocolTypesGenerator/ChromiumProtocolTypesGenerator.cs
```diff
@@ -0,0 +1,400 @@
...
+            if (domain.Commands == null)
+            {
+                return;
+            }
+
+            foreach (var command in domain.Commands)
+            {
+                // request
+                string baseName = $"{domain.Domain}{char.ToUpper(command.Name[0])}{command.Name.Substring(1)}";
+                builder.AppendLine("/// <summary>");
```
kblok: Let's treat warnings as errors.

src/PlaywightSharp.ProtocolTypesGenerator/PlaywrightSharp.ProtocolTypesGenerator.csproj
```diff
@@ -0,0 +1,20 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" Version="4.7.1" />
+  </ItemGroup>
+
+  <Import Project="../Common/Dependencies.props" />
+
+  <ItemGroup>
+    <ProjectReference Include="..\PlaywrightSharp.Abstractions\PlaywrightSharp.Abstractions.csproj" />
```
kblok: Let's add our analyzers here as well

src/PlaywightSharp.ProtocolTypesGenerator/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+        private static async Task Main(string[] args)
+        {
+            IProtocolTypesGenerator[] generators =
+            {
+                new ChromiumProtocolTypesGenerator(),
+            };
+
+            var revision = new RevisionInfo("test")
+            {
+                ExecutablePath = @"D:\Playground\Playground.cs\ConsoleApp1\ConsoleApp1\.local-chromium\Win64-706915\chrome-win\chrome.exe",
```
kblok: Can we get rid of this?

src/PlaywightSharp.ProtocolTypesGenerator/PlaywrightSharp.ProtocolTypesGenerator.csproj
Meir017: they are included through `<Import Project="../Common/Dependencies.props" />`

kblok: We are missing these guys there right https://github.com/kblok/playwright-sharp/blob/master/src/PlaywrightSharp.Chromium/PlaywrightSharp.Chromium.csproj#L9-L11 ?

src/PlaywightSharp.ProtocolTypesGenerator/ChromiumProtocolTypesGenerator.cs
Meir017: is warning that I'm not using the result. just calling `AppendLine` modifies the StringBuilder. there's no reason to use the result here...
in this case, the Analyzers are not helping at all

src/PlaywrightSharp.Chromium/Messaging/ConnectionResponse.cs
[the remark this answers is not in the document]
Meir017: I'm pretty sure a response will either have the `Params` property or the `Result` property. I added a check to see if  `Params` is an object type (ie: event data), otherwise the result is a response for a command

src/PlaywightSharp.ProtocolTypesGenerator/PlaywrightSharp.ProtocolTypesGenerator.csproj
Meir017: yeah. just one of them though

src/PlaywrightSharp.Chromium/Messaging/ConnectionResponse.cs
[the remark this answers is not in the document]
Meir017: anyways, I reverted this

src/PlaywightSharp.ProtocolTypesGenerator/ChromiumProtocolTypesGenerator.cs
kblok: Yeah, that rule is quite silly. You can disable that one across the board.

PR #242, opened 2020-03-04
fix update generated protocol

Meir017:
Now using the chrome downloaded in the tests 😞


src/PlaywightSharp.ProtocolTypesGenerator/Program.cs
```diff
@@ -8,7 +9,11 @@ private static async Task Main(string[] args)
         {
             (IProtocolTypesGenerator, RevisionInfo)[] generators =
             {
-                (new ChromiumProtocolTypesGenerator(), new RevisionInfo()),
+                (new ChromiumProtocolTypesGenerator(), new RevisionInfo
+                {
+                    ExecutablePath = Path.Combine(Directory.GetCurrentDirectory(), "..","..","..","..", @"PlaywrightSharp.Tests\bin\Debug\netcoreapp3.1\.local-chromium\Win64-733125\chrome-win\chrome.exe"),
```
kblok: Can we treat warnings as errors in this project?

kblok: Can't we use Chromium's BrowserFetcher to get the path? this will change a lot.

kblok:
Can't we take a look at the warnings on TestConstants.cs ?

src/PlaywightSharp.ProtocolTypesGenerator/Program.cs
Meir017: the `BrowserFetcher` class is located in `PlaywrightSharp.Chromium` we would have to depend on that project.
it's possible.
the only issue is running the first time since we won't have the protocol classes (webkit & firefox)

kblok: We can make a temporal change when we need it, we would need it to be dynamic at the end of the day.

Meir017:
> Can't we take a look at the warnings on TestConstants.cs ?

this is not related at all to this PR

kblok:
>this is not related at all to this PR

It's never too late.

PR #243, opened 2020-03-05
chromium page.url tests

### LINKED ISSUE #72: Pass page.spec.js tests in Chromium
Let's turn these tests green, let's remember to check Puppeteer's code

Meir017:
closes #72

### DIFF
--- a/src/PlaywrightSharp.Chromium/ChromiumBrowserApp.cs
+++ b/src/PlaywrightSharp.Chromium/ChromiumBrowserApp.cs
@@ -10,10 +10,12 @@ namespace PlaywrightSharp.Chromium
     public class ChromiumBrowserApp : IBrowserApp
     {
         private readonly ChromiumProcessManager _processManager;
+        private readonly Func<Task> _gracefullyClose;
 
-        internal ChromiumBrowserApp(ChromiumProcessManager processManager, ConnectOptions options)
+        internal ChromiumBrowserApp(ChromiumProcessManager processManager, Func<Task> gracefullyClose, ConnectOptions options)
         {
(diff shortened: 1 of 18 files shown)


src/PlaywrightSharp.Chromium/ChromiumBrowserApp.cs
```diff
@@ -10,10 +10,12 @@ namespace PlaywrightSharp.Chromium
     public class ChromiumBrowserApp : IBrowserApp
```
kblok: FYI. My next task is making this class a generic class. There is not `crBrowserApp` un Playwright.

src/PlaywrightSharp.Chromium/ChromiumSession.cs
```diff
@@ -73,9 +72,16 @@ internal async Task<TChromiumResponse> SendAsync<TChromiumResponse>(IChromiumReq
             return (TChromiumResponse)result;
         }
 
-        internal void Close(string reason)
+        internal void OnClosed()
```
kblok: let's keep the reason and add that to the exception message.

src/PlaywrightSharp/Page.cs
```diff
@@ -496,6 +514,19 @@ internal void OnPopup(object popupPage)
 
         internal void DidDisconnected()
         {
+            _disconnected = true;
```
kblok: arrow

Meir017:
@kblok the next stage is the evaluate ability

kblok:
> @kblok the next stage is the evaluate ability

Yeah, I'll end up covering that for #68

PR #245, opened 2020-03-05
Chromium: Implement Launcher.userDataDirTests

kblok:
related to #68


src/PlaywightSharp.ProtocolTypesGenerator/Chromium/ChromiumProtocolTypesGenerator.cs
```diff
@@ -99,13 +99,13 @@ private void PrepareArrayTypes(ChromiumProtocolDomainsContainer response)
                     }
                     else if (type.Type == "integer")
                     {
-                        _knownTypes[type.Id] = "int";
-                        _knownTypes[$"{domain.Domain}.{type.Id}"] = "int";
+                        _knownTypes[type.Id] = "int?";
+                        _knownTypes[$"{domain.Domain}.{type.Id}"] = "int?";
```
kblok: Properties need to be nullable, if we don't do this we would send properties with primitives default values, such as 0 for timeout

src/PlaywrightSharp.Chromium/ChromiumExecutionContext.cs
```diff
@@ -0,0 +1,104 @@
...
+        public ChromiumExecutionContext(ChromiumSession client, ExecutionContextDescription contextPayload)
+        {
+            _client = client;
+            _contextId = contextPayload.Id.Value;
+        }
+
+        public async Task EvaluateAsync(FrameExecutionContext context, bool returnByValue, string script, object[] args)
+            => await EvaluateAsync<object>(context, returnByValue, script, args).ConfigureAwait(false);
+
+        public async Task<T> EvaluateAsync<T>(FrameExecutionContext context, bool returnByValue, string script, object[] args)
```
kblok: This code is an MVP for what we need to do in the PR.

src/PlaywrightSharp.Chromium/Protocol/RemoteObjectType.cs
```diff
@@ -0,0 +1,7 @@
+namespace PlaywrightSharp.Chromium.Protocol
+{
+    internal static class RemoteObjectType
+    {
+        public static string Bigint => "bigint";
```
kblok: We might get more here. if we don't we can move it to a const.

src/PlaywrightSharp.Tests/Launcher/UserDataDirTests.cs
```diff
@@ -70,21 +72,25 @@ public async Task UserDataDirOptionShouldRestoreState()
...
-            using (var browser2 = await Playwright.LaunchAsync(options))
-            {
-                var page2 = await browser2.DefaultContext.NewPageAsync();
-                await page2.GoToAsync(TestConstants.EmptyPage);
-                Assert.Equal("hello", await page2.EvaluateAsync<string>("localStorage.hey"));
-            }
+            var page = await browser.DefaultContext.NewPageAsync();
+            await page.GoToAsync(TestConstants.EmptyPage);
+            await page.EvaluateAsync("localStorage.hey = 'hello'");
+            await browser.CloseAsync();
```
kblok: Dispose calls are fire and forget, we need to await for the close here.

Meir017: maybe we should implement `IAsyncDisposable`?

kblok: The issue with IAsyncDisposable is that it's .NET standard 2.1
That would leave .NET Framework out. I will try to see if I can make it multi-target in another PR.

PR #246, opened 2020-03-06
Treat objects as JsonElements

kblok:
We need a more dynamic object (like `JsonElement`) to handle properties that the protocol considers "Objects".

e.g. `ExecutionContextDescription` has a `AuxData` property, which is considered an `object`. If we leave this as an object we won't be able to get its `frameId` property.

Using a `JsonElement` will help us make classes for these kinds of properties.


kblok:
cc @Meir017

src/PlaywightSharp.ProtocolTypesGenerator/Chromium/ChromiumProtocolTypesGenerator.cs
```diff
@@ -26,11 +26,6 @@ public async Task GenerateTypesAsync(RevisionInfo revision)
                 Directory.CreateDirectory(directory);
             }
 
-            if (revision.Local && File.Exists(output))
```
kblok: If it's local great. If we already have the file, we wouldn't care, we have git to protect ourselves.

PR #252, opened 2020-03-10
Chroimium: Pass Headfull tests

kblok:
Related to #68


src/PlaywightSharp.ProtocolTypesGenerator/Chromium/ChromiumProtocolTypesGenerator.cs
```diff
@@ -85,7 +85,7 @@ private void PrepareArrayTypes(ChromiumProtocolDomainsContainer response)
                 {
```
kblok: Mind these changes @Meir017

PR #254, opened 2020-03-14
Screenshot tests

### LINKED ISSUE #29: Add screenshot.spec.js tests
Based on this taghttps://github.com/microsoft/playwright/tree/v0.10.0. We don't need to get the tests green. We only need to incorporate the code. Let's also check if we can get code from Puppeteer

kblok:
closes #29


Meir017:
@kblok please rename the PR...

PR #255, opened 2020-03-15
Network tests

### LINKED ISSUE #25: Add network.spec.js tests

kblok:
closes #25


src/PlaywrightSharp.Abstractions/IResponse.cs
```diff
@@ -49,5 +50,17 @@ public interface IResponse
         /// </summary>
         /// <returns>A <see cref="Task"/> that completes when the text was processed, yielding to a text representation of response body.</returns>
         Task<string> GetTextAsync();
+
+        /// <summary>
+        /// Returns a <see cref="Task"/> which resolves to a <see cref="JsonDocument"/> representation of response body.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> that completes when the json body is parsed, yielding a <see cref="JsonDocument"/> representation of response body.</returns>
+        Task<JsonDocument> GetJsonAsync();
```
Meir017: don't we want this to return `T`?

src/PlaywrightSharp.Abstractions/IResponse.cs
```diff
@@ -49,5 +50,17 @@ public interface IResponse
...
+        /// Returns a <see cref="Task"/> which resolves to a <see cref="JsonDocument"/> representation of response body.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> that completes when the json body is parsed, yielding a <see cref="JsonDocument"/> representation of response body.</returns>
+        Task<JsonDocument> GetJsonAsync();
+
+        /// <summary>
+        /// Returns a <see cref="Task"/> which resolves to a buffer with response body.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> that completes when the response is returned by the server, yielding a <see cref="byte"/> array.</returns>
+        Task<byte[]> GetBufferAsync();
```
Meir017: can this be `Span<byte>`?

src/PlaywrightSharp.Tests/PlaywrightSharp.Tests.csproj
```diff
@@ -5,6 +5,7 @@
         <IsPackable>false</IsPackable>
         <IsTestProject>true</IsTestProject>
         <LangVersion>8.0</LangVersion>
+        <ReleaseVersion>0.0.0</ReleaseVersion>
```
Meir017: is this needed?

src/PlaywrightSharp.Abstractions/IResponse.cs
kblok: we could do both :)

src/PlaywrightSharp.Abstractions/IResponse.cs
kblok: What would be the benefit for the consumer?

src/PlaywrightSharp.Tests/PlaywrightSharp.Tests.csproj
kblok: None of the changes here

src/PlaywrightSharp.Abstractions/IResponse.cs
kblok: I just checked that you can implicitly cast T[] to Span<T> ([source](https://docs.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay).

I think a byte[] as a more pure and generic type.

PR #280, opened 2020-03-29
Chromium: Pass Evaluation tests

### LINKED ISSUE #59: Pass evaluation.spec.js tests in Chromium
Let's turn these tests green, let's remember to check Puppeteer's code

kblok:
closes #59

### DIFF
--- a/src/PlaywrightSharp.Tests/Evaluation/PageEvaluateTests.cs
+++ b/src/PlaywrightSharp.Tests/Evaluation/PageEvaluateTests.cs
@@ -1,9 +1,7 @@
-using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
+using PlaywrightSharp.Helpers;
 using PlaywrightSharp.Tests.Attributes;
 using PlaywrightSharp.Tests.BaseTests;
 using Xunit;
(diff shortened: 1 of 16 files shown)


src/PlaywrightSharp.Tests/Evaluation/PageEvaluateTests.cs
```diff
@@ -116,11 +117,11 @@ public async Task ShouldEvaluateInThePageContext()
         public async Task ShouldReturnUndefinedForObjectsWithSymbols()
         {
             Assert.Null(await Page.EvaluateAsync<object>("() => [Symbol('foo4')]"));
-            Assert.Equal(JsonDocument.Parse("{}"), await Page.EvaluateAsync<JsonDocument>(@"() => {
+            Assert.Equal(JsonDocument.Parse("{}").RootElement.ToJson(), (await Page.EvaluateAsync<JsonElement>(@"() => {
```
Meir017:
```suggestion
            Assert.Equal("{}", (await Page.EvaluateAsync<JsonElement>(@"() => {
```

src/PlaywrightSharp/Helpers/StringExtensions.cs
```diff
@@ -64,33 +64,53 @@ public static string UnQuote(this string value)
...
         }
 
         /// <summary>
-        /// Determin if the script is a javascript function and not an expression.
+        /// Determine if the script is a javascript function and not an expression.
         /// </summary>
         /// <param name="script">Script to evaluate.</param>
         /// <returns>Whether the script is a function or not.</returns>
-        public static bool IsJavascriptFunction(this string script)
+        public static bool IsJavascriptFunction(ref string script)
```
Meir017: why not use the same flow as playwright does?

kblok: Thanks for pushing here. It turns out that pw adds the parenthesis to test the function, but we don't need to send that function with parenthesis.

Reverted.

PR #286, opened 2020-03-31
Pass ElementHandleFill tests

kblok:
Part of #57

I had to add `Macross.Json.Extensions` as a dependency because `System.Text.Json"  doesn't support custom enum values yet, see https://github.com/dotnet/runtime/issues/31081.

I added it only in the Chromium project,  I think it would end up also in Firefox and WebKit, but it's not needed by the rest of the projects, that's why I didn't add it to the common props.

### DIFF
--- a/src/PlaywrightSharp.Chromium/PlaywrightSharp.Chromium.csproj
+++ b/src/PlaywrightSharp.Chromium/PlaywrightSharp.Chromium.csproj
@@ -23,4 +23,7 @@
             <_Parameter1>PlaywrightSharp.Tests</_Parameter1>
         </AssemblyAttribute>
     </ItemGroup>
+    <ItemGroup>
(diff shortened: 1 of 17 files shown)


src/PlaywrightSharp.Chromium/PlaywrightSharp.Chromium.csproj
```diff
@@ -23,4 +23,7 @@
             <_Parameter1>PlaywrightSharp.Tests</_Parameter1>
         </AssemblyAttribute>
     </ItemGroup>
+    <ItemGroup>
+      <PackageReference Include="Macross.Json.Extensions" Version="1.2.1" />
```
Meir017: what does this do?

kblok: See de description :)

PR #287, opened 2020-04-01
Chromium: Pass ElementHandleHoverTests

kblok:
part of #57

### DIFF
--- a/src/PlaywrightSharp/ElementHandle.cs
+++ b/src/PlaywrightSharp/ElementHandle.cs
@@ -127,10 +127,7 @@ namespace PlaywrightSharp
         }
 
         /// <inheritdoc cref="IElementHandle.HoverAsync"/>
-        public Task HoverAsync()
-        {
-            throw new NotImplementedException();
-        }
+        public Task HoverAsync(PointerActionOptions options = null) => PerformPointerActionAsync(point => _page.Mouse.MoveAsync(point.X, point.Y), options);
 
         /// <inheritdoc cref="IElementHandle.PressAsync"/>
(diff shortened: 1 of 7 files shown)


src/PlaywrightSharp/ElementHandle.cs
```diff
@@ -127,10 +127,7 @@ public Task<double> GetVisibleRatioAsync()
         }
 
         /// <inheritdoc cref="IElementHandle.HoverAsync"/>
-        public Task HoverAsync()
-        {
-            throw new NotImplementedException();
-        }
+        public Task HoverAsync(PointerActionOptions options = null) => PerformPointerActionAsync(point => _page.Mouse.MoveAsync(point.X, point.Y), options);
```
Meir017, line 130: shouldn't this use `IPointerActionOptions` instead?

kblok: We need `IPointerActionOptions` so internally Click and Hover can call the same methods.

I think there is no much value to expose that abstraction on the API. A user might not know what to send here. If we ask for a `PointerActionOptions` the user will know that she needs to create a new ` PointerActionOptions` class.

In PW there is no `IPointerActionOptions`, `ClickOptions` it's a composition of `WaitForOptions` and `PointerActionOptions`.

What do you think we use composition instead of inheritance? 

```
public class ClipOptions
{
   public WaitForOptions WaitFor { get; set; }
   public PointerActionsOptions PointerOptions { get; set; }
}
```

Meir017: I think inheritance is OK here

PR #299, opened 2020-04-08
Webkit: Generate protocol types

### LINKED ISSUE #237: Implement protocol generator for WebKit

Meir017:
closes #237 
related to #229


Meir017:
@kblok now I just need to refactor the common code generation in chromium & webkit

Meir017:
@kblok should I create a base class for the `IBrowserType` that will have the duplicated logic?

- `GetPlatform`
- `ResolveExecutablePath`
- `GetChromeExecutablePath` / `GetFirefoxExecutablePath`
- `SetEnvVariables`

kblok:
>@kblok should I create a base class for the IBrowserType that will have the duplicated logic?

Let's do it `internal abstract class BrowserTypeBase`

src/PlaywrightSharp/BrowserTypeBase.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+namespace PlaywrightSharp
+{
+    /// <summary>
+    /// Base type for implementing <see cref="IBrowserType"/>.
+    /// </summary>
+    public abstract class BrowserTypeBase : IBrowserType
```
kblok, line 12: With so many abstracts here... how about removing `IBrowserType` here, and moving `IBrowserType` to each class?
This class would be way more cleaner and we wouldn't have so many overrides on the final classes.

Meir017: that's what I did at first but then I noticed that `BrowserTypeBase` does implement some of the members on the interface.
maybe we could rename this to `BrowserType` and remove the interface completely?

kblok: Interfaces are a feature. It will help users to inject objects using a DI. It would even help them to Mock our classes.

As far as I can see the only functionality it's implementing is `Devices`. I could even think why these aren't extension methods... just speaking my mind.

Meir017: the `ExecutablePath` is implemented.
I don't see an issue having multiple abstract methods

kblok: fair enough.

PR #305, opened 2020-04-10
Firefox: Pass page wait for tests

Meir017:
related to #113


Meir017:
@kblok what happened to the CI? works for me locally

kblok:
@Meir017 I stopped it to rebuild it. It seems we have a recurrent fail.

PR #314, opened 2020-04-14
Remove conditional compilation around DisposeAsync

### LINKED ISSUE #272: Use Microsoft.Bcl.AsyncInterfaces package instead of conditional compilation
We could use https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/ to implement IAsyncDisposable across platforms

slang25:
Fixes #272 

The reference to  `Microsoft.Bcl.AsyncInterfaces` doesn't need to be explicit as it is pulled in via `System.Text.Json`, happy to keep it in or remove it 😃


kblok:
Let's leave that there. So the dependency is clear.

kblok:
Thanks!

PR #331, opened 2020-04-30
Chromium: Pass NetworkEventsTests

kblok:
relates to #71

### DIFF
--- a/src/PlaywrightSharp.Chromium/ChromiumNetworkManager.cs
+++ b/src/PlaywrightSharp.Chromium/ChromiumNetworkManager.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using PlaywrightSharp.Chromium.Protocol;
 using PlaywrightSharp.Chromium.Protocol.Fetch;
 using PlaywrightSharp.Chromium.Protocol.Network;
+using RequestPattern = PlaywrightSharp.Chromium.Protocol.Fetch.RequestPattern;
 
 namespace PlaywrightSharp.Chromium
 {
@@ -25,6 +26,7 @@ namespace PlaywrightSharp.Chromium
         private bool _userCacheDisabled;
         private bool _protocolRequestInterceptionEnabled = false;
         private bool _userRequestInterceptionEnabled = false;
+        private Credential _credentials = null;
 
         public ChromiumNetworkManager(ChromiumSession client, Page page)
         {
@@ -42,6 +44,12 @@ namespace PlaywrightSharp.Chromium
         {
             _userCacheDisabled = !enabled;
             return UpdateProtocolCacheDisabledAsync();
+        }
+
+        internal Task SetRequestInterceptionAsync(bool value)
+        {
+            _userRequestInterceptionEnabled = value;
+            return UpdateProtocolRequestInterceptionAsync();
         }
 
         private async void Client_MessageReceived(object sender, IChromiumEvent e)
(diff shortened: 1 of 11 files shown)


src/PlaywrightSharp.ProtocolTypesGenerator/StandardProtocolTypesGeneratorBase.cs
```diff
@@ -21,6 +20,17 @@ protected override async Task GenerateTypesAsync(StringBuilder builder, Revision
 
             PrepareArrayTypes(response);
 
+            // We first creat enums.
+            foreach (var domain in response.Domains)
+            {
+                builder.Append("namespace ").Append(NamespacePrefix).Append('.').AppendLine(domain.Domain);
+                builder.AppendLine("{");
+
+                GenerateEnums(builder, domain);
```
kblok, line 29: @Meir017 we need to write all enums first so we know which types are enums

src/PlaywrightSharp.ProtocolTypesGenerator/StandardProtocolTypesGeneratorBase.cs
```diff
@@ -77,38 +87,46 @@ private void PrepareArrayTypes(ProtocolDomainsContainer response)
...
             }
 
-            foreach (var type in domain.Types)
+            // Generate enums
+            foreach (var type in domain.Types.Where(t => t.Enum != null))
             {
-                if (type.Enum != null)
-                {
-                    _enums.Add(type.Id);
+                _enums.Add($"{domain.Domain}.{type.Id}");
```
kblok, line 100: There are references between domains. For instance, inside `Fetch` we have `Network.ResourceType`.

src/PlaywrightSharp.ProtocolTypesGenerator/StandardProtocolTypesGeneratorBase.cs
```diff
@@ -182,35 +200,38 @@ private string FormatDocs(string docs)
...
             if (property.Ref != null)
             {
-                return ConvertRefToCsharp(property.Ref, property.Optional ?? false);
+                return ConvertRefToCsharp(domain, property.Ref, property.Optional ?? false);
             }
 
             return property.Type switch
             {
-                "array" => ConvertItemsProperty(property.Items, isResponse),
+                "array" => ConvertItemsProperty(domain, property.Items, isResponse),
```
kblok, line 212: @Meir017 we need to pass the domain al the way to be able to build the fully qualified enum name

src/PlaywrightSharp.Chromium/ChromiumNetworkManager.cs
```diff
@@ -44,6 +46,12 @@ internal Task SetCacheEnabledAsync(bool enabled)
             return UpdateProtocolCacheDisabledAsync();
         }
 
+        internal Task SetRequestInterceptionAsync(bool value)
```
Meir017, line 49: rename `value` to `enabled`

src/PlaywrightSharp.Tests/Page/Network/NetworkEventsTests.cs
```diff
@@ -58,13 +62,19 @@ public async Task PageEventsResponse()
         [SkipBrowserAndPlatformFact(skipFirefox: true)]
         public async Task PageEventsRequestFailed()
         {
-            Server.SetRoute("/one-style.css", context => context.Response.WriteAsync("deadbeef"));
+            int port = TestConstants.Port + 100;
+            var disposableServer = new SimpleServer(port, TestUtils.FindParentDirectory("PlaywrightSharp.TestServer"), false);
```
Meir017, line 66: do we have to create another server?

src/PlaywrightSharp.Chromium/ChromiumNetworkManager.cs
kblok: I named that at first, but pw named that value at this level. It would be easier to track changes in this way.

src/PlaywrightSharp.Tests/Page/Network/NetworkEventsTests.cs
kblok: I'm taking down this server in line 71. I don't want to mess with some possible tests using this same server.

If you know a way we can force an `ERR_INVALID_HTTP_RESPONSE` it would be great. In this way, I was able to generate an `ERR_CONNECTION_REFUSED`.  Which kinda works to test a failed request with no response.

PR #349, opened 2020-05-02
Firefox: Pass network tests

Meir017:
relates to #112

### DIFF
--- a/src/PlaywrightSharp.ProtocolTypesGenerator/ProtocolTypesGeneratorBase.cs
+++ b/src/PlaywrightSharp.ProtocolTypesGenerator/ProtocolTypesGeneratorBase.cs
@@ -43,5 +43,8 @@ namespace PlaywrightSharp.ProtocolTypesGenerator
 
         protected void GenerateResponseDefinition(StringBuilder builder, string typeName)
             => builder.AppendLine($"internal partial class {typeName}Response: I{Project}Response");
+
+        protected void GeneratePropertyDefinition(StringBuilder builder, string type, string name)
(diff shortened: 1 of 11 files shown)


src/PlaywrightSharp.ProtocolTypesGenerator/ProtocolTypesGeneratorBase.cs
```diff
@@ -43,5 +43,8 @@ protected void GenerateRequestDefinition(StringBuilder builder, string typeName)
 
         protected void GenerateResponseDefinition(StringBuilder builder, string typeName)
             => builder.AppendLine($"internal partial class {typeName}Response: I{Project}Response");
+
+        protected void GeneratePropertyDefinition(StringBuilder builder, string type, string name)
+            => builder.AppendLine($"[System.Text.Json.Serialization.JsonPropertyName(\"{name}\")] public {type} {name.ToPascalCase()} {{ get; set; }}");
```
kblok, line 48: Shouldn't this be solved by the JSON serializer?

Meir017: for most cases yes. in firefox we have a `base64body` which is converted to `Base64Body` and deserialized as `base64Body` which is wrong. this way we won't have potential errors in the future due to this

kblok: Boooo! 
Good solution 😉

Meir017:
this can be merged. I currently have a hanging test in `ResponseTextTests` so that will be in another PR

PR #365, opened 2020-05-07
fix flaky test

Meir017:
in playwright they just verify that a simple catch is enough without checking which specific error is thrown


Meir017:
@kblok `PlaywrightSharp.Tests.Page.NetworkIdleTests.ShouldNavigateToEmptyPage` seems to be flaky too.

I think we could fix that in another PR since it's not related to this and is another flaky test and shouldn't block this PR :)

kblok:
@Meir017 NetworkIdleTests are a headache.

Which test are we trying to improve here?

PR #390, opened 2020-05-14
Fix ShouldWorkWithRequestInterception

kblok:


src/PlaywrightSharp.Tests/RequestInterception/IgnoreHttpsErrorsTests.cs
```diff
@@ -31,7 +31,7 @@ public async Task ShouldWorkWithRequestInterception()
 
             await page.SetRequestInterceptionAsync(true);
             page.Request += async (sender, e) => await e.Request.ContinueAsync();
-            var response = await Page.GoToAsync(TestConstants.EmptyPage);
+            var response = await page.GoToAsync(TestConstants.HttpsPrefix + "/empty.html");
```
Meir017, line 34: consider also fixing lines 26-29 to call the class method `NewPageAsync` like playwright does

Issue #367, opened 2020-05-07
Review NetworkIdleTests


kblok:
Those tests are quite instable in CI.

Meir017:
I found an issue here,
in some places we have
```cs
foreach (var watcher in LifecycleWatchers.ToArray())
{
    watcher.On...();
}
```
sometimes `watcher` is null, not sure if the right thing is to change it to `watcher?.On...()` but it could help

kblok:
We could try that. My main concern is whether the bug is that null check or if we have some race condition and we needed to send a message to a disposed watcher. We'll see.

PR #400, opened 2020-05-18
Null check on watchers

kblok:
Possible fix for #367


kblok:
@Meir017 it doesn't fix the issue, but I think it helps to have the same approach across the board.

PR #401, opened 2020-05-19
Chromium: Pass coverage tests

kblok:
Related to #80

### DIFF
--- a/src/PlaywrightSharp.Tests/Chromium/JSCoverageTests.cs
+++ b/src/PlaywrightSharp.Tests/Chromium/JSCoverageTests.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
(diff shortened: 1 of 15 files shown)


src/PlaywrightSharp.Tests/Chromium/JSCoverageTests.cs
```diff
@@ -13,10 +12,14 @@ namespace PlaywrightSharp.Tests.Chromium
 {
     ///<playwright-file>chromium/chromium.spec.js</playwright-file>
     ///<playwright-describe>JSCoverage</playwright-describe>
-    public class CoverageTests : PlaywrightSharpPageBaseTest
+    [Trait("Category", "chromium")]
+    [Trait("Category", "firefox")]
```
Meir017, line 16: I think we should add firefox and webkit once we actually implement `ICoverage` for them

Meir017:
why does `src/PlaywrightSharp.Chromium/Protocol/Protocol.Generated.cs` have changes?

kblok:
>why does src/PlaywrightSharp.Chromium/Protocol/Protocol.Generated.cs have changes?

dotnet format 🤷

src/PlaywrightSharp.Tests/Chromium/JSCoverageTests.cs
kblok: That's a good idea, but on the other side, these tests are ignored for firefox and webkit. There are no coverage classes on firefox and webkit.

Meir017: for me it's a bit confusing since it seems that these tests are covered in firefox & webkit. maybe these tests should only have the chromium trait

kblok: agreed

PR #412, opened 2020-05-23
Remove networkidle2 tests

### LINKED ISSUE #367: Review NetworkIdleTests

kblok:
Playwright removed networkidle2 support here https://github.com/microsoft/playwright/pull/1883

We are not removing that feature yet. But simplifying our tests will bring us more stability.

closes #367


Meir017:
this was before version 1.0.0 so it makes sense :)

PR #437, opened 2020-05-26
Chromium: Pass PageEventsCloseTests

kblok:
related to #72


src/PlaywrightSharp.Tests/Page/Events/PageEventsCloseTests.cs
```diff
@@ -1,50 +1,53 @@
-using System.Threading.Tasks;
```
Meir017, line 1: dotnet format?

kblok: It doesn't like the move and the edit. I split that into two commits but the diff is the same here.
It looks ok if you look this through a git client.

PR #438, opened 2020-05-26
Chromium: Pass PageEventsPageErrorTests

kblok:
related to #72


src/PlaywrightSharp.Tests/Page/Events/PageEventsPageErrorTests.cs
```diff
@@ -1,36 +1,39 @@
-using System.Threading.Tasks;
```
Meir017, line 1: dotnet format?

kblok: Same, te git move. It looks ok on git.

PR #449, opened 2020-05-28
Firefox: pass PageEventsConsoleTests

Meir017:
related to #113


kblok:
@Meir017 tell me if you don't love getting all greens 😂

PR #460, opened 2020-06-01
Firefox: pass PageAuthenticateTests

Meir017:
related to #106


src/PlaywrightSharp.Firefox/FirefoxPage.cs
```diff
@@ -246,7 +250,13 @@ public Task SetCacheEnabledAsync(bool enabled)
 
         public Task SetRequestInterceptionAsync(bool enabled) => throw new NotImplementedException();
 
-        public Task AuthenticateAsync(Credentials credentials) => throw new NotImplementedException();
+        public Task AuthenticateAsync(Credentials credentials)
+            => _session.SendAsync(
+                new NetworkSetAuthCredentialsRequest
+                {
+                    Username = credentials?.Username,
```
Meir017, line 257: we need to allow to serialize as `{ username: null, password: null }`

kblok: Can you explain this a little bit more?

src/PlaywrightSharp.Firefox/FirefoxPage.cs
```diff
@@ -494,7 +504,7 @@ private void OnWorkerCreated(PageWorkerCreatedFirefoxEvent e)
             string workerId = e.WorkerId;
             var worker = new Worker(e.Url);
             FirefoxSession tempWorkerSession = null;
-            var workerSession = new FirefoxSession(_session.Connection, "worker", workerId, async (id, request) =>
+            var workerSession = new FirefoxSession(_session.Connection, "worker", workerId, async (id, request, jsonOptions) =>
```
kblok, line 507: Same, could you explain this?

src/PlaywrightSharp.Firefox/FirefoxPage.cs
Meir017: we need to send a json that includes both properties with a null value, currently we ignore all null values

src/PlaywrightSharp.Firefox/FirefoxPage.cs
Meir017: this is in order for us to be able to use custom options that will allow us to handle properties in a special way

src/PlaywrightSharp.Firefox/FirefoxPage.cs
```diff
@@ -246,7 +250,13 @@ public Task SetCacheEnabledAsync(bool enabled)
...
         public Task SetRequestInterceptionAsync(bool enabled) => throw new NotImplementedException();
 
-        public Task AuthenticateAsync(Credentials credentials) => throw new NotImplementedException();
+        public Task AuthenticateAsync(Credentials credentials)
+            => _session.SendAsync(
+                new NetworkSetAuthCredentialsRequest
+                {
+                    Username = credentials?.Username,
+                    Password = credentials?.Password,
+                }, options: _authenticateJsonOptions);
```
kblok, line 259: We don't need the `_authenticateJsonOptions` variable, we could create the instance here. It won't be that expensive and we won't have that in memory all the time.

src/PlaywrightSharp.Firefox/FirefoxPage.cs
kblok: Can't we mark Username and Password to send null all the time? 
I know it's autogenerated, but can we think about a workaround there? I would rather mess with the code generator than adding a boilerplate only for this.

src/PlaywrightSharp.Firefox/FirefoxPage.cs
```diff
@@ -1,617 +1,616 @@
...
-
-        private class AuxData
-        {
-            public string FrameId { get; set; }
-
-            public string Name { get; set; }
-        }
-    }
-}
+using System;
```
kblok, line 1: Ouch

src/PlaywrightSharp.Firefox/FirefoxPage.cs
Meir017: hmm, I'm not sure how. I could try adding a `JsonConverter` for this class

Meir017: once #477 is merged I'll add the JsonConverter for this

kblok: #477 Merged.

src/PlaywrightSharp.Firefox/Helper/NetworkSetAuthCredentialsRequestJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+using PlaywrightSharp.Firefox.Protocol.Network;
+using PlaywrightSharp.Firefox.Protocol.Runtime;
+using PlaywrightSharp.Helpers;
+
+namespace PlaywrightSharp.Firefox.Helper
+{
+    /// <summary>
+    /// We need to be able to send null values in order to reset the username/password values.
+    /// </summary>
+    internal class NetworkSetAuthCredentialsRequestJsonConverter : JsonConverter<NetworkSetAuthCredentialsRequest>
```
kblok, line 13: My concern here is whether we are going to need more of these or not. I wouldn't like to have 5 classes like this.

Meir017: I think this is the last one

PR #482, opened 2020-06-04
Firefox: pass PageOpenerTests

Meir017:
related to #113

### DIFF
--- a/src/PlaywrightSharp.Firefox/FirefoxPage.cs
+++ b/src/PlaywrightSharp.Firefox/FirefoxPage.cs
@@ -259,7 +259,7 @@ namespace PlaywrightSharp.Firefox
 
         public Task<byte[]> GetPdfAsync(string file, PdfOptions options) => throw new NotImplementedException();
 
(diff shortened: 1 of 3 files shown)


src/PlaywrightSharp.Firefox/FirefoxPage.cs
```diff
@@ -259,7 +259,7 @@ public Task SetEmulateMediaAsync(MediaType? mediaType, ColorScheme? colorScheme)
 
         public Task<byte[]> GetPdfAsync(string file, PdfOptions options) => throw new NotImplementedException();
 
-        public Task<IPage> GetOpenerAsync() => throw new NotImplementedException();
+        public async Task<IPage> GetOpenerAsync() => await _openerResolver().ConfigureAwait(false);
```
kblok, line 262: Why do we need the async/await?

Meir017: `_openerResolver` returns `Task<Page>` and we need to return as `IPage`

PR #609, opened 2020-06-11
Firefox: Pass Worker tests

### LINKED ISSUE #119: Pass workers.spec.js tests in Firefox

kblok:
closes #119


src/PlaywrightSharp.Firefox/FirefoxPage.cs
```diff
@@ -401,7 +405,7 @@ private void OnNavigationCommitted(PageNavigationCommittedFirefoxEvent e)
         {
             foreach (var pair in _workers)
             {
-                if (pair.Key == e.FrameId)
+                if (pair.Value.FrameId == e.FrameId)
```
Meir017, line 408: so this is what I missed... 😄

PR #616, opened 2020-06-17
Firefox: Pass IgnoreHttpsErrorsTests

kblok:
related to #106


kblok:
@Meir017 I had the feeling this was the one

kblok:
@Meir017 `IgnoreHTTPSErrors` is at the browser level in Firefox. So this test was affecting other tests coming after that.

PW doesn't have this issue because they run this test after HTTPS tests :p  

Create a browser for this test should fix the issue.

Meir017:
@kblok damn, that's sneeky

PR #618, opened 2020-06-17
Firefox: Pass InterceptionSetOfflineModeTests

kblok:
related to #106


Meir017:
this can be merged as it's not supported yet on firefox and the tests are skipped :)

kblok:
@Meir017 I will remove this in the next PR. We agree on not adding categories if the test suite doesn't apply.

PR #623, opened 2020-06-18
Skip ShouldFireDisconnectedWhenClosingWithWebSocket on Firefox linux

kblok:
we will fix this on #622

### DIFF
--- a/src/PlaywrightSharp.Tests/Launcher/LaunchWebSocketOptionTests.cs
+++ b/src/PlaywrightSharp.Tests/Launcher/LaunchWebSocketOptionTests.cs
@@ -1,4 +1,5 @@
 using System.Threading.Tasks;
+using PlaywrightSharp.Tests.Attributes;
 using PlaywrightSharp.Tests.BaseTests;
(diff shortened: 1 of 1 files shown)


src/PlaywrightSharp.Tests/Launcher/LaunchWebSocketOptionTests.cs
```diff
@@ -48,7 +49,7 @@ public async Task ShouldSupportTheWebSocketOption()
         ///<playwright-file>launcher.spec.js</playwright-file>
         ///<playwright-describe>Playwright.launch |webSocket| option</playwright-describe>
         ///<playwright-it>should fire "disconnected" when closing with webSocket</playwright-it>
-        [Retry]
+        [SkipBrowserAndPlatformFact(skipFirefox: true, skipLinux: true)]
```
Meir017, line 52: this should have a comment mentioning it would be fixed in #622

kblok: Not necessary. We can track that here in GitHub.

PR #634, opened 2020-06-20
Update launcher tests

### LINKED ISSUE #514: Review changes on launcher.spec.js
This PR might help https://github.com/kblok/playwright/pull/2/files. This issue is not only about the test itself, but also see what was changed under the hood

kblok:
closes #514


Meir017:
@kblok I think that adding `Microsoft.Extensions.Logging` should be a separate PR

kblok:
@Meir017 agree.

kblok:
Takings this back @Meir017. The logger is not just a method call in pw (https://github.com/microsoft/playwright/blob/master/src/server/browserType.ts#L108) is part of the API. 
I won't implement all the logging but I'm going to implement this API wise, so we don't need to go back to this a change all the API again.

PR #803, opened 2020-09-01
VS Code setup

kblok:
Created two launch settings to launch the PDF and Screenshot Demos. 
I also added a task to run tests
image

image


kblok:
@dgozman 😉

PR #844, opened 2020-09-11
Remove Microsoft.AspNetCore.StaticFiles dependency

slang25:
Reused existing code to replace the need for `Microsoft.AspNetCore.StaticFiles`.
Also removed old `InternalsVisibleTo` attributes to the legacy projects.


kblok:
Thank you @slang25. I'll merge it in a few days when I get green tests back. Just in case we don't break anything with this :)

kblok:
@slang25 would you mind pulling from master?

slang25:
Sure, done 🙂

Issue #157, opened 2020-01-24
Build docfx site


kblok:

Meir017:
@kblok can this be closed?

kblok:
Yes Sir!

PR #893, opened 2020-10-02
Initial docs using docfx

Meir017:
related to #157


Meir017:
something is working https://meir017.github.io/playwright-sharp/api/index.html
image

Meir017:
in order for this to work I had to do the steps of creating a secret described here - https://github.com/maxheld83/ghpages#secrets

arjun27:
This is awesome! I'm wondering if we can set it up like in the Python repo: https://github.com/microsoft/playwright-python/blob/master/.github/workflows/deploy-docs.yml#L27

GITHUB_TOKEN is pre-configured on Actions agents.

kblok:
I would follow this recipe to push to the gh_pages branch https://github.com/dotnetconfig/dotnet-config/blob/dev/.github/workflows/pages.yml

kblok:
Another thing @Meir017. Only if it's a small flag, if not we can leave that for later.
http://www.puppeteersharp.com/ doesn't have a global search as https://dotnetconfig.org/
Could you check if that's easy to add?

Meir017:
@kblok adding 
```json
"globalMetadata": {
      "_appTitle": "Playwright Sharp",
      "_enableSearch": true
    }
```
solved this :)
image

kblok:
This is awesome!

kblok:
I think that scope wise this is great.
I would make it build only on `tag`. So we show the docs of the published app. But I would merge it as is so we get the first doc up and running.

@Meir017 do you want to create a PR on https://github.com/kblok/playwright-sharp  so we make sure that the gh-pages branch works?

We would need to work on the home page @arjun27

Meir017:
@kblok I created the PR on https://github.com/Meir017/playwright-sharp to make sure it works.
to verify that it works I added 
```yml
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```
so it will run for each commit :)
and you can see it here - https://meir017.github.io/playwright-sharp

kblok (approved):

Meir017:
I think it's better to have the docs be built even without tags, this way even typo fixes will be deployed and won't wait for a version release

kblok:
@Meir017 that's a fair point, but pushing an unshipped API to the official site won't be ok. No one will care about the docs from the master branch.
Can we add a manual trigger (https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) in case of a red alert?

>not sure it's possible with docfx but maybe we could add a way to view previous version docs similar to playwright

https://playwright.dev/ rocks. The search is awesome. It feels great. I would create markdowns based on the XML and use the same tool. But my ultimate dream is hosting our docs in docs.microsoft.com.

Meir017:
@kblok now it can run manually too.
I copied the release part from playwright, is it correct?
image

kblok:
@Meir017 sounds perfect!

PR #894, opened 2020-10-02
Fix PDF generation

### LINKED ISSUE #884: Error when generating pdf
When trying to generate a pdf after updating to latest nuget package (0.111.2), I receive the following error:
`PlaywrightSharpException: options.format.toLowerCase is not a function`

after calling it like this:
```
var result = await page.GetPdfAsync(null, pdfConversionOptions.Scale, pdfConversionOptions.DisplayHeaderFooter, pdfConversionOptions.HeaderTemplate, pdfConversionOptions.FooterTemplate,
  pdfConversionOptions.PrintBackground, pdfConversionOptions.Landscape, pdfConversionOptions.PageRanges, pdfConversionOptions.Format, pdfConversionOptions.Width, pdfConversionOptions.Height,
  pdfConversionOptions.Margin, pdfConversionOptions.PreferCSSPageSize);
```

I tried also specifying directly `PaperFormat.Letter` for example but I receive the same error...

kblok:
closes #884

### DIFF
--- a/src/PlaywrightSharp/Page.cs
+++ b/src/PlaywrightSharp/Page.cs
@@ -701,6 +701,11 @@ namespace PlaywrightSharp
             Margin margin = null,
             bool preferCSSPageSize = false)
         {
+            if (BrowserContext.BrowserName != "chromium")
+            {
+                throw new NotSupportedException($"{BrowserContext.BrowserName} doesn't support this action.");
+            }
+
             byte[] result = Convert.FromBase64String(await _channel.GetPdfAsync(
                 scale,
                 displayHeaderFooter,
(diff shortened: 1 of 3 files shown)


src/PlaywrightSharp/Page.cs
```diff
@@ -701,6 +701,11 @@ public async Task<IResponse> WaitForResponseAsync(Func<IResponse, bool> predicat
             Margin margin = null,
             bool preferCSSPageSize = false)
         {
+            if (BrowserContext.BrowserName != "chromium")
```
Meir017, line 704: do we have a constant for this?

PR #901, opened 2020-10-04
devops: run tests on all operating systems

### LINKED ISSUE #833: Add Windows and Linux to the CI

mxschmitt:
it can only go wrong

Closes #833

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,9 +9,14 @@ on:
       - master
 
 jobs:
-  chromium-mac:
+  test:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -9,9 +9,14 @@ on:
       - master
 
 jobs:
-  chromium-mac:
+  test:
     name: Chromium on MacOS
```
kblok, line 13: We need a new name :)

mxschmitt: yep

PR #970, opened 2020-10-23
doc(page): add download example

### LINKED ISSUE #968: Download file
Hi again, can you please create a simple example of how to process downloads? From button click to save as filename if possible?

Many thanks,
-M

kblok:
closes #968


kababoom (commented):
Excelent...
One note, acceptDownloads is not a member of LaunchPersistentContextAsync. There it must be set with the LaunchPersistentOptions. But both versions are working fine though. Thanks

kblok:
Good point @kababoom. I will add the persistent example.

PR #976, opened 2020-10-28
driver: update to v0.152.0

kblok:


Meir017:
do we need `src/PlaywrightSharp/Drivers/0.152.0`?

kblok:
Good catch  @Meir017 :)

PR #981, opened 2020-11-02
feat(playwright): install browsers by default

### LINKED ISSUE #977: Add ability to run the install from the CreateAsync method
If we have a parameter `bool installIfNeeded` we could avoid having two lines with the `InstallAsync` and `CreateAsync`.

kblok:
closes #977 

This will be a breaking change for users using the dotnet tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ dotnet tool install playwright-sharp-tool -g
 playwright-sharp install-browsers
 ```
 
-If you runthese two commands, you can call `CreateAsync` passing `installIfNeeded` in false.
+If you run these two commands, you can call `CreateAsync` passing `installIfNeeded` in false.
 
 ```cs
  using var playwright = await Playwright.CreateAsync(installIfNeeded: false);
--- a/README.md
+++ b/README.md
@@ -16,14 +16,12 @@ Headless execution is supported for all browsers on all platforms.
 Playwright Sharp relies on two external components: The browsers and the Playwright driver.
 
 ```cs 
-await Playwright.InstallAsync();
 var playwright = await Playwright.CreateAsync();
 ```
 
-`Playwright.InstallAsync()` will download the required browsers. `Playwright.CreateAsync()` will create and launch the playwright driver.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -38,15 +36,17 @@ dotnet tool install playwright-sharp-tool -g
 playwright-sharp install-browsers
 ```
 
-By running these two commands, you can avoid having the `await Playwright.InstallAsync();` line in your code.
+If you runthese two commands, you can call `CreateAsync` passing `installIfNeeded` in false.
```
Meir017, line 39:
```suggestion
If you run these two commands, you can call `CreateAsync` passing `installIfNeeded` in false.
```

docfx_project/documentation/index.md
```diff
@@ -36,15 +34,17 @@ dotnet tool install playwright-sharp-tool -g
 playwright-sharp install-browsers
 ```
 
-By running these two commands, you can avoid having the `await Playwright.InstallAsync();` line in your code.
+If you runthese two commands, you can call `CreateAsync` passing `installIfNeeded` in false.
```
Meir017, line 37:
```suggestion
If you run these two commands, you can call `CreateAsync` passing `installIfNeeded` in false.
```

src/PlaywrightSharp/Playwright.cs
```diff
@@ -100,20 +100,27 @@ internal Playwright(IChannelOwner parent, string guid, PlaywrightInitializer ini
         /// <param name="debug">Enables the playwright driver log. Pass `pw:api` to get the Playwright API log.
         /// It is a shortcut to the DEBUG=pw:api environment variable.
         /// </param>
+        /// <param name=installIfNeeded">Call <see cref="InstallAsync(string, string)"/> before starting the driver</param>
```
Meir017, line 103:
```suggestion
        /// <param name=installIfNeeded">Call <see cref="InstallAsync(string, string)"/> before starting the driver.</param>
```

kblok:
We are going to invest in a better dotnet tool.

PR #991, opened 2020-11-12
Drivers as runtimes content

kblok:
This PR is taking playwright drivers out of the PlaywrightSharp.dll. These drivers will now be included in the nugget package inside the [runtimes folder](https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks?WT.mc_id=DT-MVP-5003814#architecture-specific-folders).

The dotnet build process will take a platform-specific driver based on the `RuntimeIdentifier` set in the `csproj` file, or based on the `runtime` used in the [dotnet publish](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish?WT.mc_id=DT-MVP-5003814#options) command.
If the `RuntimeIdentifier` is not set the entire `runtimes` folder is being copied. I'm checking that scenario as a fallback. 

This change has the following benefits:
 * We avoid writing files in the bin folder in runtime.
 * The PlaywrightSharp.dll will be under 500kb.
 * We will be able to ship .NET 5 specific versions without having huge dlls.
 * The driver (CLI) will be "visible" at build time and also at deploy time. So users will be able to run the `install` command without a dotnet tool.

This PR also includes a `targets` file. This will add a target to the build process, which will **download the browser locally at build time**.

That means that if PlaywrightSharp is used as an end-to-end testing tool, **users won't need to run `InstallAsync`**.  
And by making the CLI visible, users using PlaywrightSharp in WebApps **won't need to run `InstallAsync` as long as they run `playwright-cli.exe install` (in windows) or `./playwright-cli install` in OSX/Linux as part of the deploy process**. So with this PR **we will discourage the usage of `InstallAsync`**.


kblok:
cc @arjunattam @pavelfeldman @yury-s

kblok:
Would you mind taking a look at this @Meir017?

.github/workflows/nuget-package-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: dotnet build ./src/PlaywrightSharp/PlaywrightSharp.csproj -c Debug
+      - name: Delete browsers on Mac
+        if: ${{ matrix.os == 'macos-latest' }}
+        run: rm -R /Users/runner/Library/Caches/ms-playwright
+      - name: Delete browsers on Linux
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: rm -R /home/runner/.cache/ms-playwright
+      - name: Delete browsers on windows
+        if: ${{ matrix.os == 'windows-latest' }}
+        run: Remove-Item -path C:\Users\runneradmin\AppData\Local\ms-playwright -recurse
```
mxschmitt, line 35: simplified, but not clue about Windows since ~ works probably not there
```suggestion
      - name: Delete browsers on Mac
        if: matrix.os == 'macos-latest'
        run: rm -r ~/Library/Caches/ms-playwright
      - name: Delete browsers on Linux
        if: ${{ matrix.os == 'ubuntu-latest' }}
        run: rm -r ~/.cache/ms-playwright
      - name: Delete browsers on windows
        if: matrix.os == 'windows-latest'
        run: Remove-Item -path C:\Users\runneradmin\AppData\Local\ms-playwright -recurse
```

mxschmitt (commented):
added nits to the CI config, to be more independent.

arjunattam:
This is exciting! I'm assuming the dotnet-tool will not be required anymore, and we can recommend cli for manual installs (if required) and codegen. How would users know where the cli is? Is the path trivial?

kblok:
> I'm assuming the dotnet-tool will not be required anymore, and we can recommend cli for manual installs (if required)

Exactly :)

>  and codegen. 

I'm thinking that we could wrap the CLI into the current dotnet tool so users can do `playwright-sharp codegen` no matter the current directory. What do you think?


> How would users know where the cli is? Is the path trivial?

It's trivial, it's the `bin` folder where the app is compiled.

ScottHuangZL:
It is work if I do not publish as single file format.

However, if I try below:
**dotnet publish -r win-x64 -c Release /p:PublishSingleFile=true /p:PublishTrimmed=true**

The publish itself process is work, but when I click the main exe, it would show error **"The path is empty. (Parameter 'path')"**

Can you help to advise solution? Or else, I have to publish without single file format, that means lots of dlls ... and seems messy...

Btw, it try to set driverExecutablePath parameters, but not sure which correct path, can you advise?

ScottHuangZL:
> It is work if I do not publish as single file format.
> 
> However, if I try below:
> **dotnet publish -r win-x64 -c Release /p:PublishSingleFile=true /p:PublishTrimmed=true**
> 
> The publish itself process is work, but when I click the main exe, it would show error **"The path is empty. (Parameter 'path')"**
> 
> Can you help to advise solution? Or else, I have to publish without single file format, that means lots of dlls ... and seems messy...
> 
> Btw, it try to set driverExecutablePath parameters, but not sure which correct path, can you advise?

I find can fix by set driver path as "./playwright-cli.exe", issue closed.

PR #1001, opened 2020-11-17
feat(screencast): new recordVideo option

### LINKED ISSUE #994: v0.161.0: Implement recordVideo
In newContext, newPage, launch and launchPersistent

kblok:
closes #994


kblok:
@pavelfeldman @yury-s I'm dropping the obsoletes `videoPath` and `videoSize` options from .NET in favor of the new `recordVideo`.
I think we don't need to leave obsoletes fields in .NET and Java in a v0 stage. What do you think?

yury-s:
> I think we don't need to leave obsoletes fields in .NET and Java in a v0 stage. What do you think?

Agreed. I'm planning to drop obsolete fields in Java as well.

yury-s (approved):

PR #1020, opened 2020-11-24
feat: record HAR

### LINKED ISSUE #993: v0.161.0: Implement recordHar
In newContext, newPage, launch and launchPersistent

kblok:
closes #993 

HAR classes are not required. But users can take advantage of this class to deserialize the HAR file.
I also update the assets folder.


src/PlaywrightSharp/LaunchPersistentOptions.cs
```diff
@@ -90,6 +90,13 @@ public class LaunchPersistentOptions : LaunchOptions
         /// </summary>
         public Dictionary<string, string> ExtraHttpHeaders { get; set; }
 
+        /// <summary>
+        /// Enables HAR recording for all pages into recordHar.path file. If not specified, the HAR is not recorded.
+        /// Make sure to await <see cref="IPage.CloseAsync(bool)"/> for the HAR to be saved.
+        /// You can use <see cref="Har.HarResult"/> to deserialize the generated JSON file.
```
dgozman, line 96: How can we make sure that HarResult actually matches the implementation upstream? I am worried they may diverge over time. Do we use HarResult to serialize har in the tests?

kblok: This class is not being used in the seriaization process.
I first created that class in the test code. But I then realised that it might help the user to process the file.
There are fields that I grabbed from the source code which are not in the tests.
I think that as a user I would feel more comfortable having an async method that gives me the HAR result (even if I ask that to a closed context) instead of having to open a file and process that myself.
That API would have an output type we could enforce.
Another solution would be adding that to the api.json and generate these classes based on that.

dgozman: I agree that having types is nice for the user. However, if we ship them, we should test them. Can we use them to parse our har files in the tests? This way we'll catch issues if the types say one thing, but in reality we produce a different json. Otherwise, we'd better remove the types, because we cannot be sure they are actually correct.

kblok: > we should test them. Can we use them to parse our har files in the tests?

I'm using it to deserialize the har files :)

dgozman: Oh, I missed that! It's all good then.

kblok: @dgozman I added a test to check the consistency of the HarResult class, if we get data changes in the har file, that test will fail.

PR #1029, opened 2020-11-27
devops: update browsers table on driver update

### LINKED ISSUE #1006: Update browsers table on DriverDownloader
When we update the drivers we could also update the browser table from upstream

kblok:
closes #1006


avodovnik (approved):
LGTM

PR #1037, opened 2020-12-07
chore: remove internals visible to tests

### LINKED ISSUE #1034: Websocket property in WebSocketEventArgs should probably be public?
I was trying to use page.Websocket with 1.162.0 and it seems it's not possible since the property is marked as internal. I believe this was just an oversight and should be public instead.

https://github.com/microsoft/playwright-sharp/blob/5b0ad490548016bf2d7e4da8efec5a8600236ff6/src/PlaywrightSharp/WebSocketEventArgs.cs#L13

kblok:
I'm removing the InternalsVisibleTo  PlaywrightSharp.Tests.
I think there are some costs in doing this. 
 * There are some selector tests we won't be able to perform. But those tests are upstream so we can rely on them.
 * There are a few internal classes that I have to make public. I didn't see a risk there.

The value here is writing code close to the real usage.

closes #1034


src/PlaywrightSharp.Tests/SelectorsRegisterTests.cs
```diff
@@ -38,6 +38,7 @@ public async Task ShouldWork()
             var page = await context.NewPageAsync();
             await page.SetContentAsync("<div><span></span></div><div></div>");
 
+            /* I'd rather ignore these tests instead of exposing a CreateSelectorForTestAsync function
```
avodovnik, line 41: Can we remove them entirely? Having stale code in the code base is probably not helpful.

src/PlaywrightSharp/Playwright.cs
```diff
@@ -23,6 +23,11 @@ public sealed class Playwright : ChannelOwnerBase, IPlaywright, IChannelOwner<Pl
         /// </summary>
         public const int DefaultTimeout = 30_000;
 
+        /// <summary>
+        /// Environment variable use to set the browsers path.
+        /// </summary>
+        public const string BrowsersPathEnvironmentVariable = "PLAYWRIGHT_BROWSERS_PATH";
```
avodovnik, line 29: Is there any nicer place to put this? I'm just nit-picking here, but having a `const` feels _dirty_ to me.

src/PlaywrightSharp/PlaywrightSharpException.cs
```diff
@@ -8,8 +8,15 @@ namespace PlaywrightSharp
     /// </summary>
     public class PlaywrightSharpException : Exception
     {
-        internal const string BrowserClosedExceptionMessage = "Browser has been closed";
-        internal const string BrowserOrContextClosedExceptionMessage = "Target page, context or browser has been closed";
+        /// <summary>
+        /// Message used when the browser gets closed.
+        /// </summary>
+        public const string BrowserClosedExceptionMessage = "Browser has been closed";
```
avodovnik, line 14: Same as above (can we maybe have these consts somewhere else, would that make any sense?).

avodovnik (approved):
I'd remove the commented code, but the rest is optional - maybe worth a conversation.

src/PlaywrightSharp/Playwright.cs
kblok: Why does it feel dirty?
These constants are mapping constants in upstream.

src/PlaywrightSharp/PlaywrightSharpException.cs
kblok: I have two things in mind. A `Constants` class or a `CommonErrors` class. wdyt?

avodovnik: So, I'd probably be inclined to move it to a `Resources` somewhere. Maybe not `resx` but... If we have strings that are actual resources, why not, right?

kblok: I feel like resources are way too much overload for just a const string.

src/PlaywrightSharp/Playwright.cs
avodovnik: That's fine then, let's leave it where it is. Maybe as we grow & add more env variables we can have a struct or something to _separate_/group them together somewhere.

src/PlaywrightSharp/PlaywrightSharpException.cs
avodovnik: Yeah, maybe too much. How about something like `CommonErrorMessages`, then?  And just put them in a folder/namespace `Resources`? How's that for a compromise?

kblok: I like it! 👍

kblok:
How about now @avodovnik ?

avodovnik (approved):

PR #1060, opened 2020-12-17
feat(context): introduce storage state

### LINKED ISSUE #1052: v0.170.0 Add new arguments to Browser.NewContext, Browser.NewPage
Let's also add Context.StorageState because it's part of the same.

kblok:
closes #1052


src/PlaywrightSharp/BrowserContext.cs
```diff
@@ -306,6 +307,21 @@ public Task UnrouteAsync(Func<string, bool> url, Action<Route, IRequest> handler
...
         public Task SetExtraHttpHeadersAsync(Dictionary<string, string> headers) => Channel.SetExtraHttpHeadersAsync(headers);
 
+        /// <inheritdoc />
+        public async Task<StorageState> GetStorageStateAsync(string path = null)
+        {
+            var state = await Channel.GetStorageStateAsync().ConfigureAwait(false);
+
+            if (!string.IsNullOrEmpty(path))
+            {
+                File.WriteAllText(
```
avodovnik, line 317: Do we want to have any sort of checking for this? i.e. that it doesn't already exists, etc.? Otherwise, this might throw an exception.

src/PlaywrightSharp/BrowserContextOptions.cs
```diff
@@ -236,6 +254,21 @@ public BrowserContextOptions Clone()
...
             }
 
+            if (Proxy != null)
+            {
+                args["proxy"] = Proxy;
+            }
+
+            if (!string.IsNullOrEmpty(StorageStatePath))
+            {
+                StorageState = JsonSerializer.Deserialize<StorageState>(File.ReadAllText(StorageStatePath), JsonExtensions.DefaultJsonSerializerOptions);
```
avodovnik, line 264: Same comment as above applies. Should we be more defensive here? Or do we just "ignore it" and let the use handle it?

src/PlaywrightSharp/NameValueEntry.cs
```diff
@@ -0,0 +1,50 @@
+using System;
+using System.Collections.Generic;
+
+namespace PlaywrightSharp
+{
+    /// <summary>
+    /// Name value entry.
+    /// </summary>
+    public class NameValueEntry : IEquatable<NameValueEntry>
```
avodovnik, line 9: Is there a benefit of using this vs `KeyValuePair`?

src/PlaywrightSharp/BrowserContext.cs
kblok: I thought about that. If we don't validate the user gets an exception. If we do validate we will need to throw an exception. 
In both scenarios, the user gets an exception. Does it worth to validate that?

src/PlaywrightSharp/NameValueEntry.cs
kblok: Let's drop it :)

src/PlaywrightSharp/BrowserContext.cs
avodovnik: Yeah, that's a fair point. Let's leave it as is, then - at least for now. :-)

src/PlaywrightSharp/NameValueEntry.cs
kblok: ok. we can drop it. We need to send to the driver an object with a Name/Value pair, not a Key/Value pair

avodovnik: Ah, so it's a JSON serialization problem... Well, leave it, then. This might change if/when we auto-generate the channels. And we can figure it out then =) How's that?

kblok: It will definitely change :)

avodovnik (approved):
Looks good to me.

PR #1086, opened 2021-01-04
fix(browserType): remove firefoxUserPrefs from launch persistent options

### LINKED ISSUE #1085: LaunchOptions.FirefoxUserPrefs not working with existing firefox profile
Hello! FirefoxUserPrefs not working with existing firefox profile, please check, thank you!
**Not working**
`LaunchOptions option = new LaunchOptions();`
`option.FirefoxUserPrefs = new Dictionary<string, object>() { { "media.peerconnection.enabled", false } }; `
`var browserContext = _player.Firefox.LaunchPersistentContextAsync(_userDataDir, option);`

Screenshot_5

**Work ok**

`LaunchOptions option = new LaunchOptions();`
`option.FirefoxUserPrefs = new Dictionary<string, object>() { { "media.peerconnection.enabled", false } }; `
`var browser = _player.Firefox.LaunchAsync(option);`

Screenshot_6

kblok:
closes #1085


src/PlaywrightSharp.Tests/BrowserTypeLaunchTests.cs
```diff
@@ -171,5 +171,20 @@ public async Task ShouldWorkWithEnvironmentVariables()
 
             await using var browser = await BrowserType.LaunchAsync(options);
         }
+
+        /// <summary>
+        /// PuppeteerSharp test. It's not in upstream
```
avodovnik, line 176: Will it be? Should it be?

avodovnik (approved):
I would still personally prefer to somehow split the browser specific options away, but I understand that's there to keep in sync with upstream API, sort of. So, looks fine for now.

src/PlaywrightSharp.Tests/BrowserTypeLaunchTests.cs
kblok: We have this method https://github.com/microsoft/playwright-sharp/blob/afb06a5a3c69f48f524260ad382aab59efd13e77/src/PlaywrightSharp/BrowserType.cs#L268
Which internally converts a `LaunchOption` to a `LaunchPersistentOption`.
So I'm validating that if you want to use a `LaunchOption` you didn't set the `FirefoxUserPrefs`. Typescript doesn't have this feature. wdyt?

PR #1104, opened 2021-01-15
chore(test): Implement ability to identify missing Tests

### LINKED ISSUE #158: Console app to check missing tests
Our test library will create an XML file pointing to Playwright tests. We need a console app to check that XML against the tests we have in some Playwright checkout.

avodovnik:
Closes #158 (once done).


avodovnik:
You can run this utility by calling:

```powershell
apichecker testtests --specfilelocations "C:\Users\anvod\src\playwright\test\"
```

it will produce an output similar to this:

```
Total matching tests: 357/1330.
Total tests found by name, but not by file: 907/1330.
browser: should create new page
        browser: should create new page
        browser: should create new page
launcher: should fire close event for all contexts
        browsertype-launch: should fire close event for all contexts
launcher: should be callable twice
        browsertype-launch: should be callable twice
        page-basic: should be callable twice
dispatchevent: should dispatch click event
        page-dispatchevent: should dispatch click event
        page-dispatchevent: should dispatch click event
dispatchevent: should dispatch click event properties
        page-dispatchevent: should dispatch click event properties
dispatchevent: should dispatch click svg
        page-dispatchevent: should dispatch click svg
dispatchevent: should dispatch click on a span with an inline element inside
        page-dispatchevent: should dispatch click on a span with an inline element inside
dispatchevent: should not fail when element is blocked on hover
        page-dispatchevent: should not fail when element is blocked on hover
dispatchevent: should dispatch click when node is added in shadow dom
        page-dispatchevent: should dispatch click when node is added in shadow dom
dispatchevent: should be atomic
        page-dispatchevent: should be atomic

...

Total missing tests: 66/1330.
dispatchevent.spec.js: should dispatch click after navigation
dispatchevent.spec.js: should dispatch click after a cross origin navigation
download.spec.js: should report download path within page.on('download', .) handler for Files
download.spec.js: should report download path within page.on('download', .) handler for Blobs
elementhandle-wait-for-element-state.spec.ts: should throw waiting for hidden when detached
page-screenshot.spec.js: path option should detect joeg
page-set-input-files.spec.js: should emit addListener/removeListener
screencast.spec.js: should scale frames down to the requested size
wait-for-selector-1.spec.js: Page.$ waitFor is shortcut for main frame
interception.spec.js: should work with request interception
interception.spec.js: should work with request interception
interception.spec.js: should support cors with different methods
queryselector.spec.js: should not throw in case of missing selector
queryselector.spec.js: should should support * capture
queryselector.spec.js: should return empty array for non-existing element
queryselector.spec.js: should return multiple elements
queryselector.spec.js: query
queryselector.spec.js: create
web.spec.js: should navigate
web.spec.js: should receive events
navigation.spec.js: should work with HistoryAPI
navigation.spec.js: should work for file urls
navigation.spec.js: should throw if networkidle0 is passed as an option
navigation.spec.js: should throw if networkidle is passed as an option
navigation.spec.js: should not leak listeners during navigation of 11 pages
navigation.spec.js: extraHTTPHeaders should be pushed to provisional page
page.spec.js: should return the correct browser instance
page.spec.js: page should be client stub
click.spec.js: should click the button with px border with relative point
click.spec.js: should retry when element is animating from outside the viewport with force
page.spec.js: should *not* run beforeunload by default
page.spec.js: should fill time input after clicking
page.spec.js: should throw when element is not an &lt;input&gt;, &lt;textarea&gt; or [contenteditable]
page.spec.js: should be able to fill the input[type=number] with empty string
page.spec.js: should respect name
page.spec.js: should respect url
page.spec.js: should throw when element is not a &lt;select&gt;
page.spec.js: should return the page title
page.spec.js: should include hashes
page.spec.js: should be a sane user agent
network.spec.js: should work with compression
dialog.spec.js: should log prompt actions
page.spec.js: should work with window.close
page.spec.js: should work with page.close
page.spec.js: should fire when expected
page.spec.js: should fire when expected
keyboard.spec.js: should press the metaKey
keyboard.spec.js: should send a character with sendCharacter
jshandle.spec.js: should return empty map for non-objects
jshandle.spec.js: should return even non-own properties
fixtures.spec.js: should close the browser on double SIGINT
fixtures.spec.js: should close the browser on SIGINT + SIGERM
evaluation.spec.js: should allow cross-frame js handles
evaluation.spec.js: should work with browser context scripts with path
evaluation.spec.js: should not throw an error when evaluation does a synchronous navigation and returns an undefined
emulation.spec.js: should throw for invalid timezone IDs
elementhandle-convenience.spec.js: 'innerText should throw
elementhandle.spec.js: should throw for &lt;br&gt; elements with force
elementhandle.spec.js: should work when Node is removed
chromium/chromium.spec.js: should NOT report stylesheets across navigations
chromium/chromium.spec.js: shouldn't ignore eval() scripts if reportAnonymousScripts is true
browsercontext.spec.js: sshould fail without credentials
browsercontext.spec.js: should fail if wrong credentials
permissions.spec.js: should grant geolocation permission when listed
autowaiting.spec.js: calling window.stop
accessibility.spec.js: filtering children of leaf nodes
Found/Mismatched/Missing: 357/907/66 out of 1330
```

avodovnik:
Also, @kblok to make this work, I had to remove the download of browsers target. I'm not sure if that was a problem *on my machine only*, but worth noting that this can happen. If it happens on another machine, we should investigate further and/or remove the build targets like we talked about.

kblok:
Can we stop generating the XML doc file so we don't get all the missing XML comment warnings?

kblok:
I conflicted this PR sorry @avodovnik

src/tools/ApiChecker/ScaffoldTest.cs
```diff
@@ -135,5 +173,21 @@ private static void AddTest(CodeCompileUnit @class, string testDescribe, string
                         new CodePrimitiveExpression("This test is not yet implemented.")),
                 }));
         }
+
+        /// <summary>
+        /// Describes the options for scaffolding the tests.
+        /// </summary>
+        [Verb("scaffold-test", HelpText = "Takes a spec.ts file and scaffolds the C# test.")]
+        internal class ScaffoldTestOptions
```
kblok, line 181: Any reason to move the class here?

src/tools/ApiChecker/ScaffoldTest.cs
```diff
@@ -14,6 +39,31 @@ internal static class ScaffoldTest
...
+            }
+        }
+
+        /// <summary>
+        /// Generates a clean name from the test name.
+        /// </summary>
+        /// <param name="testDescribe">The original test name.</param>
+        /// <returns>Returns a "clean" string, suitable for C# method names.</returns>
+        public static string CleanName(string testDescribe) =>
+            new string(Array.FindAll(_textInfo.ToTitleCase(testDescribe).ToCharArray(), c => char.IsLetterOrDigit(c)));
```
kblok, line 65:
```suggestion
        public static string CleanName(string testDescribe)
            => new string(Array.FindAll(_textInfo.ToTitleCase(testDescribe).ToCharArray(), c => char.IsLetterOrDigit(c)));
```

src/tools/ApiChecker/ScaffoldTest.cs
```diff
@@ -14,6 +39,31 @@ internal static class ScaffoldTest
     {
         private static readonly TextInfo _textInfo = CultureInfo.InvariantCulture.TextInfo;
 
+        public static void FindTestsInFile(string path, Action<string> callback)
+        {
+            Regex rx = new Regex(@"it\(\'(.*)\',");
```
kblok, line 44:
```suggestion
            var rx = new Regex(@"it\(\'(.*)\',");
```

src/tools/ApiChecker/IdentifyMissingTests.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using CommandLine;
+
+namespace ApiChecker
+{
+#pragma warning disable SA1201 // Elements should appear in the correct order
```
kblok, line 36: Why don't we honor this rule?

kblok (changes requested):

src/tools/ApiChecker/ScaffoldTest.cs
avodovnik: So, they are completely tied to each other and it makes no sense to have them anywhere outside. While it seems a bit weird to have a nested class, I don't see any good reason to have them as separate files - i.e. for code readability, this makes it easier, for portability (when we move this piece somewhere else) it will make it easier too, etc. :-)

src/tools/ApiChecker/IdentifyMissingTests.cs
avodovnik: It's kind of _stupid_. But that `Options` class is `internal` so if we followed that rule, it'd have to be _beneath_ the `Run` method, which doesn't really make sense.

src/tools/ApiChecker/ScaffoldTest.cs
avodovnik: So, are you OK if I resolve this? :-)

src/tools/ApiChecker/IdentifyMissingTests.cs
kblok: I would follow that rule...

avodovnik: Well, rules are rules :-).

I moved it to the bottom, same as with `ScaffoldTest`. Consistency.

src/tools/ApiChecker/IdentifyMissingTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            if (!directoryInfo.Exists)
+            {
+                throw new ArgumentException($"The location ({directoryInfo.FullName}) specified does not exist.");
+            }
+
+            // let's map the test cases from the spec files
+            MapTestsCases(directoryInfo, options);
+
+            // now, let's load the DLL and use some reflection-fu
+            var assembly = Assembly.Load(options.TestDLLName);
```
kblok, line 58: Do we need to load the assembly if you are referencing the project?

src/tools/ApiChecker/IdentifyMissingTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            string attributeTypeName = typeof(PlaywrightSharp.Tests.PlaywrightTestAttribute).Name;
```
kblok, line 59: It seems this variable is unused.

src/tools/ApiChecker/IdentifyMissingTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            var attributes = assembly.DefinedTypes.SelectMany(
+                type => type.GetMethods().SelectMany(
+                    method => method.GetCustomAttributes<PlaywrightSharp.Tests.PlaywrightTestAttribute>()));
+
+            int potentialMatches = 0;
+            int fullMatches = 0;
+            int noMatches = 0;
+            int totalTests = 0;
+
+            List<(string, string)> missingTests = new();
```
kblok, line 70: No named here?

src/tools/ApiChecker/IdentifyMissingTests.cs
```diff
@@ -0,0 +1,164 @@
...
+                Console.WriteLine($"{invalidTest.Item1}: {invalidTest.Item2}");
+            }
+
+            Console.WriteLine($"Found/Mismatched/Missing: {fullMatches}/{potentialMatches}/{noMatches} out of {totalTests}");
+        }
+
+        private static void MapTestsCases(DirectoryInfo directoryInfo, IdentifyMissingTestsOptions options)
+        {
+            // get the sub-directories
+            if (options.Recursive)
```
kblok, line 128: When wouldn't we want to go recursive?

kblok (changes requested):

src/tools/ApiChecker/IdentifyMissingTests.cs
avodovnik: Fair enough, I've replaced it with `GetAssembly` from the attribute type.

src/tools/ApiChecker/IdentifyMissingTests.cs
avodovnik: I could see a scenario of having tests for new features in a subfolder with subfolders. I figured it was "cheap" enough to implement.

kblok (approved):

PR #1106, opened 2021-01-20
chore: prepare for v0.180.0

kblok:


src/tools/ApiChecker/IdentifyMissingTests.cs
```diff
@@ -1,6 +1,6 @@
 /*
  * MIT License
- * 
+ *
```
kblok, line 3: @avodovnik can we remove the trailing space from the doc template? 🤓

PR #1112, opened 2021-01-28
Fix Mobile Geolocation snippet

### LINKED ISSUE #1111: Mobile Geolocation snippet does not work as expected
When you copy-paste the snippet provided on README.md file it doesn't wait for map to get loaded. The screenshot generated by the code is mostly depend on the network speed of you machine.

Snippet on README.md file.
```csharp
using var playwright = await Playwright.CreateAsync();
await using var browser = await playwright.Webkit.LaunchAsync(false);

var contextOptions = playwright.Devices["iPhone 11 Pro"].ToBrowserContextOptions();
contextOptions.Locale = "en-US";
contextOptions.Geolocation = new Geolocation { Longitude = 12.492507m, Latitude = 41.889938m };
contextOptions.Permissions = new[] { ContextPermission.Geolocation };

var context = await browser.NewContextAsync(contextOptions);
var page = await context.NewPageAsync();
await page.GoToAsync("https://maps.google.com");
page.ClickAsync("text='Your location'"); //
await page.ScreenshotAsync("colosseum-iphone.png");
```

**Actual Output:**

**Expected Output:**

KoditkarVedant:
Fix #1111 

When you copy-paste the Mobile Geolocation snippet provided on README.md file it doesn't wait for map to get loaded. The screenshot generated by the code is mostly depend on the network speed of you machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,16 +69,12 @@ var context = await browser.NewContextAsync(contextOptions);
 var page = await context.NewPageAsync();
 await page.GoToAsync("https://www.google.com/maps");
 
-const string MY_LOCATION_BTN_SELECTOR = ".ml-button-my-location-fab";
-await page.ClickAsync(MY_LOCATION_BTN_SELECTOR);
-
+await page.ClickAsync(".ml-button-my-location-fab");
 await page.WaitForLoadStateAsync(LifecycleEvent.Networkidle);
 
-const string STAY_ON_WEB_BTN_SELECTOR = ".ml-promotion-no-thanks";
-if ((await page.QuerySelectorAsync(STAY_ON_WEB_BTN_SELECTOR)) != null)
+if ((await page.QuerySelectorAsync(".ml-promotion-no-thanks")) != null)
 {
-    await page.ClickAsync(STAY_ON_WEB_BTN_SELECTOR);
-    await page.WaitForTimeoutAsync(500);
+    await page.ClickAsync(".ml-promotion-no-thanks");
 }
 
 await page.ScreenshotAsync("colosseum-iphone.png");
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ This snippet emulates Mobile Safari on a device at a given geolocation, navigate
 
 ```cs 
 using var playwright = await Playwright.CreateAsync();
-await using var browser = await playwright.Webkit.LaunchAsync(false);
+await using var browser = await playwright.Webkit.LaunchAsync(headless: false);
 
 var contextOptions = playwright.Devices["iPhone 11 Pro"].ToBrowserContextOptions();
 contextOptions.Locale = "en-US";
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,7 +68,16 @@ contextOptions.Permissions = new[] { ContextPermission.Geolocation };
 var context = await browser.NewContextAsync(contextOptions);
 var page = await context.NewPageAsync();
 await page.GoToAsync("https://maps.google.com");
-page.ClickAsync("text='Your location'"); //
+await page.ClickAsync("text='Your location'");
+await page.WaitForLoadStateAsync(LifecycleEvent.Networkidle);
+
+try {
+    await page.ClickAsync("text=STAY ON WEB");
```
kblok, line 75: We could check if `QuerySelectorAsync("text=STAY ON WEB") != null` instead of using a try catch

KoditkarVedant: Done. removed try-catch block. Noticed that modal can alternatively have button with test NOT RIGHT NOW 😄.

KoditkarVedant: Do you think it is worth to commit these examples in the repository?

kblok: If you can find a more stable example of geolocalization that would be awesome. Those texts in that page are localized, so it will work only in English.
Do you think you can replace it with a better example?

KoditkarVedant: I will work on this over the weekend and see if I can come up with something.

KoditkarVedant: I've used the class query selector instead of text.

README.md
```diff
@@ -67,8 +67,20 @@ contextOptions.Permissions = new[] { ContextPermission.Geolocation };
...
+
+const string MY_LOCATION_BTN_SELECTOR = ".ml-button-my-location-fab";
+await page.ClickAsync(MY_LOCATION_BTN_SELECTOR);
+
+await page.WaitForLoadStateAsync(LifecycleEvent.Networkidle);
+
+const string STAY_ON_WEB_BTN_SELECTOR = ".ml-promotion-no-thanks";
+if ((await page.QuerySelectorAsync(STAY_ON_WEB_BTN_SELECTOR)) != null)
+{
+    await page.ClickAsync(STAY_ON_WEB_BTN_SELECTOR);
```
kblok, line 80: I like how it looks :)
Just to make this example simpler. How about removing the consts?

README.md
```diff
@@ -67,8 +67,20 @@ contextOptions.Permissions = new[] { ContextPermission.Geolocation };
...
+    await page.WaitForTimeoutAsync(500);
```
kblok, line 81: Let's remove this. I think we don't need this for the purpose of this example.

README.md
KoditkarVedant: Remove it. I just didn't want to type twice to avoid typo. 😄

README.md
KoditkarVedant: I've removed it. But It was required cause the modal was taking some time to fade out.

kblok (approved):

PR #1113, opened 2021-01-29
experiment(API): Experimental API

avodovnik:
This is a draft PR where some experiments will land, including the API that's generated from the upstream model.


experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IAccessibility.cs
```diff
@@ -0,0 +1,67 @@
...
+	/// <a href="https://en.wikipedia.org/wiki/Switch_access">switches</a>.
+	/// Accessibility is a very platform-specific thing. On different platforms, there are different screen readers that might have
+	/// wildly different output.
+	/// Rendering engines of Chromium, Firefox and Webkit have a concept of "accessibility tree", which is then translated into different
+	/// platform-specific APIs. Accessibility namespace gives access to this Accessibility Tree.
+	/// Most of the accessibility tree gets filtered out when converting from internal browser AX Tree to Platform-specific AX-Tree
+	/// or by assistive technologies themselves. By default, Playwright tries to approximate this filtering, exposing only the "interesting"
+	/// nodes of the tree.
+	/// </summary>
+	public interface IAccessibility
```
kblok, line 58: how about creating `partial interface`?

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace PlaywrightSharp
+{
+    /// <summary>
+	/// <seealso cref="[EventEmitter]"/>
+	/// A Browser is created via <see cref="IBrowserType.LaunchAsync"/>. An example of using a <see cref="IBrowser"/> to create a
+	/// <see cref="IPage"/>:
+	/// </summary>
+	public interface IBrowser
+	{
+		event EventHandler<IBrowser> OnDisconnected;
```
kblok, line 53: missing comment here?

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
```diff
@@ -0,0 +1,86 @@
...
+		/// <summary>
+		/// In case this browser is obtained using <see cref="IBrowserType.LaunchAsync"/>, closes the browser and all of its pages (if
+		/// any were opened).
+		/// In case this browser is connected to, clears all created contexts belonging to this browser and disconnects from the browser
+		/// server.
+		/// The <see cref="IBrowser"/> object itself is considered to be disposed and cannot be used anymore.
+		/// </summary>
+		Task CloseAsync();
+		/// <summary>
```
kblok, line 62: can we add spaces between members?

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
```diff
@@ -0,0 +1,86 @@
...
+		/// </summary>
+		dynamic GetContexts();
+		/// <summary>
+		/// Indicates that the browser is connected.
+		/// </summary>
+		bool IsConnected();
+		/// <summary>
+		/// Creates a new browser context. It won't share cookies/cache with other browser contexts.
+		/// </summary>
+		Task<IBrowserContext> NewContextAsync(bool acceptDownloads, bool bypassCSP, ColorScheme colorScheme, float deviceScaleFactor, IEnumerable<KeyValuePair<string, string>> extraHTTPHeaders, object geolocation, bool hasTouch, object httpCredentials, bool ignoreHTTPSErrors, bool isMobile, bool javaScriptEnabled, string locale, bool offline, string[] permissions, object proxy, BrowserStorageState storageState, string timezoneId, string userAgent);
```
kblok, line 73: shouldn't we honor the `require` bit in arguments?

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
```diff
@@ -0,0 +1,176 @@
...
+		Task ClearPermissionsAsync();
+		/// <summary>
+		/// Closes the browser context. All the pages that belong to the browser context will be closed.
+		/// </summary>
+		Task CloseAsync();
+		/// <summary>
+		/// If no URLs are specified, this method returns all cookies. If URLs are specified, only cookies that affect those URLs are
+		/// returned.
+		/// </summary>
+		Task<BrowserContextCookiesResult[]> CookiesAsync(string[] urls);
```
kblok, line 94: I would love this to be resolved as `GetCookiesAsync`

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
```diff
@@ -0,0 +1,176 @@
...
+		/// <summary>
+		/// The method adds a function called {PARAM} on the `window` object of every frame in every page in the context. When called,
+		/// the function executes {PARAM} and returns a [Promise] which resolves to the return value of {PARAM}. If the {PARAM} returns a [Promise], it will be awaited.
+		/// The first argument of the {PARAM} function contains information about the caller: `{ browserContext: BrowserContext, page:
+		/// Page, frame: Frame }`.
+		/// See <see cref="IPage.ExposeBindingAsync"/> for page-only version.
+		/// An example of exposing page URL to all frames in all pages in the context:
+		/// An example of passing an element handle:
+		/// </summary>
+		Task ExposeBindingAsync(string name, Action callback, bool handle);
```
kblok, line 104: We will need a few overloads for this callback

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserType.cs
```diff
@@ -0,0 +1,83 @@
...
+		/// <summary>
+		/// Returns the persistent browser context instance.
+		/// Launches browser that uses persistent storage located at {PARAM} and returns the only context. Closing this context will
+		/// automatically close the browser.
+		/// </summary>
+		Task<IBrowserContext> LaunchPersistentContextAsync(path userDataDir, bool acceptDownloads, string[] args, bool bypassCSP, bool chromiumSandbox, ColorScheme colorScheme, float deviceScaleFactor, bool devtools, path downloadsPath, IEnumerable<KeyValuePair<string, string>> env, path executablePath, IEnumerable<KeyValuePair<string, string>> extraHTTPHeaders, object geolocation, bool handleSIGHUP, bool handleSIGINT, bool handleSIGTERM, bool hasTouch, bool headless, object httpCredentials, bool ignoreDefaultArgs, string[] ignoreDefaultArgsValues, bool ignoreHTTPSErrors, bool isMobile, bool javaScriptEnabled, string locale, bool offline, string[] permissions, object proxy, float slowMo, float timeout, string timezoneId, string userAgent);
+		/// <summary>
+		/// Returns browser name. For example: `'chromium'`, `'webkit'` or `'firefox'`.
+		/// </summary>
+		string GetName();
```
kblok, line 81: I think we talk about this. I think that non-async methods like this should be resolve to Properties.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/ICDPSession.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.IO;
+using System.Runtime.Serialization;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace PlaywrightSharp
+{
+    /// <summary>
+	/// <seealso cref="[EventEmitter]"/>
```
kblok, line 47: We won't have EvenEmitters

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IPage.cs
```diff
@@ -0,0 +1,629 @@
...
+	/// The Page class emits various events (described below) which can be handled using any of Node's native 
+	/// <a href="https://nodejs.org/api/events.html#events_class_eventemitter">`EventEmitter`</a> methods, such as `on`, `once` or
+	/// `removeListener`.
+	/// This example logs a message for a single page `load` event:
+	/// To unsubscribe from events use the `removeListener` method:
+	/// </summary>
+	public interface IPage
+	{
+		event EventHandler<IPage> OnClose;
+		event EventHandler<IConsoleMessage> OnConsole;
```
kblok, line 60: I think we will need to auto-generate EvenHandles to wrap up the values.
Here, I would expect a CloseEvenHandler with a ConsoleMessage Property.

I think we talked about removing the `On`.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IPage.cs
```diff
@@ -0,0 +1,629 @@
...
+		event EventHandler<IPage> OnCrash;
```
kblok, line 61: `IPage` where is self-referencing. Maybe we could have one PageEvenHandler. and we put in the page property the same page..

Meir017:
@avodovnik how is this code generated? shouldn't the generator be included in this PR?

avodovnik:
@Meir017 so, the generator is at https://github.com/microsoft/playwright/pull/5089, but it's a WIP. That said, I wanted to make sure there's a PR opened here so that I can solicit at least some comments and showcase the results.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IChromiumBrowserContext.cs
```diff
@@ -47,7 +47,7 @@ namespace PlaywrightSharp
 	/// <seealso cref="<see cref="IBrowserContext"/>"/>
 	/// Chromium-specific features including background pages, service worker support, etc.
 	/// </summary>
-	public interface IChromiumBrowserContext
+	public interface IChromiumBrowserContext : IBrowserContext
```
kblok, line 50: Nice!

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IPage.cs
```diff
@@ -67,7 +67,7 @@ public interface IPage
 		event EventHandler<IFrame> FrameDetached;
 		event EventHandler<IFrame> FrameNavigated;
 		event EventHandler<IPage> OnLoad;
-		event EventHandler<Error> PageError;
+		event EventHandler<Exception> PageError;
```
kblok: I don't if we should use an Exception to return page error data.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/ITimeoutError.cs
```diff
@@ -48,7 +48,7 @@ namespace PlaywrightSharp
 	/// TimeoutError is emitted whenever certain operations are terminated due to timeout, e.g. <see cref="IPage.WaitForSelectorAsync"/> or
 	/// <see cref="IBrowserType.LaunchAsync"/>.
 	/// </summary>
-	public interface ITimeoutError
+	public interface ITimeoutError : Exception
```
kblok, line 51: This won't compile 🤓

avodovnik: I know :-) The project doesn't compile at the moment, I'm working through the problems. I'm moving `ITimeoutError` to be `TimeoutException`. We can reconsider the approach to exceptions once the first "draft" lands, I think.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IAccessibility.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.IO;
+using System.Runtime.Serialization;
+using System.Text.Json;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace PlaywrightSharp
+{
+    /// <summary>
```
kblok, line 47: It seems we have a missing space here.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IAccessibility.cs
```diff
@@ -0,0 +1,90 @@
...
+		/// <remarks>
+		/// <para>
+		/// The Chromium accessibility tree contains nodes that go unused on most platforms
+		/// and by most screen readers. Playwright will discard them as well for an easier to
+		/// process tree, unless <paramref name="interestingOnly"/> is set to <c>false</c>
+		/// </para>
+		/// </remarks>
+		/// <param name="interestingOnly">Prune uninteresting nodes from the tree. Defaults to <c>true</c></param>
+		/// <param name="root">The root DOM element for the snapshot. Defaults to the whole page.</param>
+		Task<string> SnapshotAsync(bool interestingOnly, IElementHandle root);
```
kblok, line 88: We should honor the required field making non-required arguments optional.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
```diff
@@ -0,0 +1,224 @@
...
+		/// <para>
+		/// Emitted when Browser gets disconnected from the browser application. This might
+		/// happen because of one of the following:
+		/// </para>
+		/// <list type="bullet">
+		/// <item><description>Browser application is closed or crashed.</description></item>
+		/// <item><description>The <see cref="IBrowser.CloseAsync"/> method was called.</description></item>
+		/// </list>
+		/// </summary>
+		event EventHandler<IBrowser> Disconnected;
```
kblok, line 65: EvenHandlers should implement EventArgs

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
```diff
@@ -0,0 +1,224 @@
...
+		/// </summary>
+		Task CloseAsync();
+	
+		/// <summary>
+		/// <para>
+		/// Returns an array of all open browser contexts. In a newly created browser, this
+		/// will return zero browser contexts.
+		/// </para>
+		/// </summary>
+		dynamic GetContexts();
```
kblok, line 89: This could have a better return type
image

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
```diff
@@ -0,0 +1,224 @@
...
+	
+		/// <summary><para>Indicates that the browser is connected.</para></summary>
+		bool IsConnected();
```
kblok, line 92: I think it would be great if sync getters like `Contexts` and `IsConnected` can be generated as read-only properties.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
```diff
@@ -0,0 +1,224 @@
...
+		/// <param name="offline">Whether to emulate network being offline. Defaults to <c>false</c></param>
+		/// <param name="permissions">
+		/// A list of permissions to grant to all pages in this context. See <see cref="IBrowserContext.GrantPermissionsAsync"/>
+		/// for more details.
+		/// </param>
+		/// <param name="proxy">
+		/// Network proxy settings to use with this context. Note that browser needs to be launched
+		/// with the global proxy for this option to work. If all contexts override the proxy,
+		/// global proxy will be never used and can be any string, for example <c>launch({ proxy:
+		/// { server: 'per-context' } })</c>
```
kblok, line 134: This example is typescript. Can we get around that?

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
```diff
@@ -0,0 +1,411 @@
...
+		/// </summary>
+		/// <remarks>
+		/// <para>
+		/// The order of evaluation of multiple scripts installed via <see cref="IBrowserContext.AddInitScriptAsync"/>
+		/// and <see cref="IPage.AddInitScriptAsync"/> is not defined.
+		/// </para>
+		/// </remarks>
+		/// <param name="aScript">Script to be evaluated in all pages in the browser context.</param>
+		/// <param name="sScript">Script to be evaluated in all pages in the browser context.</param>
+		/// <param name="bScript">Script to be evaluated in all pages in the browser context.</param>
```
kblok, line 128: Maybe we could suffix the type `scriptAction`, `scriptString`,  `scriptObject`.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
```diff
@@ -0,0 +1,411 @@
...
+		/// <item><description>`'gyroscope'`</description></item>
+		/// <item><description>`'magnetometer'`</description></item>
+		/// <item><description>`'accessibility-events'`</description></item>
+		/// <item><description>`'clipboard-read'`</description></item>
+		/// <item><description>`'clipboard-write'`</description></item>
+		/// <item><description>`'payment-handler'`</description></item>
+		/// </list>
+		/// </param>
+		/// <param name="origin">The [origin] to grant permissions to, e.g. "https://example.com".</param>
+		Task GrantPermissionsAsync(string[] permissions, string origin);
```
kblok, line 233: I created an enum for these permissions.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
```diff
@@ -0,0 +1,411 @@
...
+		/// <summary><para>Creates a new page in the browser context.</para></summary>
+		Task<IPage> GetNewPageAsync();
+	
+		/// <summary>
+		/// <para>
+		/// Returns all open pages in the context. Non visible pages, such as <c>"background_page"</c>
+		/// will not be listed here. You can find them using <see cref="IChromiumBrowserContext.BackgroundPages"/>.
+		/// </para>
+		/// </summary>
+		dynamic GetPages();
```
kblok, line 244: I think we should never get to the point to return a dynamic. I would make it fail.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/ICDPSession.cs
kblok: This comment is still valid.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/ICDPSession.cs
```diff
@@ -0,0 +1,73 @@
...
+	{
+		/// <summary>
+		/// Detaches the CDPSession from the target. Once detached, the CDPSession object won't emit any events and can't be used to
+		/// send messages.
+		/// </summary>
+		Task DetachAsync();
+	
+		/// <param name="method">To be generated.</param>
+		/// <param name="@params">To be generated.</param>
+		Task<dynamic> SendAsync(string method, object @params);
```
kblok, line 71: I would generic this

kblok (changes requested):

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
avodovnik: Yeah, I've not rendered comment on all of them yet, but they will all end up having them.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
avodovnik: We might, yes. I propose we add those overloads *when* we need them - i.e. if users _actually_ ~need~ use them :-).

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/ICDPSession.cs
avodovnik: This was removed in the docs side, IIRC.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IPage.cs
avodovnik: While I agree, in theory, this might bite us in the behind later. I've removed them for now - simplification is still a good goal!

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IPage.cs
avodovnik: It's `event EventHandler<string> PageError;` for now, but we might consider introducing some special object later down the line, if we feel like we want to.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IAccessibility.cs
avodovnik: I don't think there should be any blank lines at the start, but also, we'll run `dotnet format` anyway, right? So some things will get cleared out automatically _after_ they get generated.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
avodovnik: I disagree with this. The new .NET guidelines also mention that this is no longer required or suggested. I want to make it simpler and having the event argument be the same object as in other languages is beneficial for people moving between them.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
avodovnik: Yup, I agree. Done.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
avodovnik: Yes, all documentation will eventually be migrated to be C# for the .NET port.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
avodovnik: Hm, yeah, the values are specified in the documentation, but not as an enum. This *might* be because we want to keep some flexibility in here and not make it a part of the API, but I'm not 100% sure. Still, while this is the case, let's keep the _generated_ method like this, and, as previously stated, we can add an overload/extension method to make the user's life simpler.

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IAccessibility.cs
kblok: I'm talking about this

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowser.cs
kblok: Are we going to create an issue for this?

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IBrowserContext.cs
```diff
@@ -0,0 +1,410 @@
...
+		/// <summary><para>Creates a new page in the browser context.</para></summary>
+		Task<IPage> GetNewPageAsync();
+	
+		/// <summary>
+		/// <para>
+		/// Returns all open pages in the context. Non visible pages, such as <c>"background_page"</c>
+		/// will not be listed here.
+		/// </para>
+		/// </summary>
+		dynamic Pages { get; }
```
kblok, line 243: The output shouldn't be `dynamic`

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IConsoleMessage.cs
```diff
@@ -0,0 +1,77 @@
...
+	
+		/// <summary>
+		/// <para>
+		/// URL of the resource followed by 0-based line and column numbers in the resource
+		/// formatted as <c>URL:line:column</c>
+		/// </para>
+		/// </summary>
+		string Location { get; }
+	
+		string Text { get; }
```
kblok, line 65: The compiler won't like having these props without comments, would that mean that we need to go upstream and add some docs?

experiments/PlaywrightSharp/PlaywrightSharp/ApiSurface/IAccessibility.cs
kblok: Can we solve this tiny thing @avodovnik?

experiments/PlaywrightSharp/PlaywrightSharp/Microsoft.PlaywrightSharp.xml
```diff
@@ -0,0 +1,6217 @@
+<?xml version="1.0"?>
```
kblok, line 1: I think we shouldn't commit this file

avodovnik:
Closing this, as we've now started the work to migrate to this API.

PR #1115, opened 2021-02-03
chore: Move build task to the tooling project

kblok:
* ApiChecker project was renamed to PlaywrightTooling
* Both build tasks were moved to the tooling project:
  * To download the drivers now you'll need to run ` dotnet run -p ./src/tools/PlaywrightSharp.Tooling/PlaywrightSharp.Tooling.csproj -- download-drivers --basepath .`
  * To check the API you'll need to run `dotnet run -p ./src/tools/PlaywrightSharp.Tooling/PlaywrightSharp.Tooling.csproj -- check-api --basepath .`
* The dependency with the test project was removed. That caused a "functional" circular dependency. You need tooling to download the drivers to build the library, so then you can build the tests.
* As we don't have a dependency with the test project, a new Xunit library and moved the `PlaywrightTestAttribute` there. This project might help to package and ship Xunit utilities.


avodovnik:
image

👀

avodovnik (approved):
Looks good to me, and it builds on Windows. Progress.

PR #1154, opened 2021-02-16
use async in tool

SimonCropp:
note i dont know how to smoke test this. but i figured async all the way would be better


kblok:
I will test this locally @SimonCropp. Thanks!

kblok (approved):

PR #1171, opened 2021-02-20
cache json options for perf

SimonCropp:
> If you use JsonSerializerOptions repeatedly with the same options, don't create a new JsonSerializerOptions instance each time you use it. Reuse the same instance for every call. 

https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-configure-options

Also some of the side effects:

 * https://www.meziantou.net/avoid-performance-issue-with-jsonserializer-by-reusing-the-same-instance-of-json.htm
 * https://github.com/dotnet/runtime/issues/38982#issuecomment-656491611

Note that net5 has [copy semantics for better performance](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-configure-options?pivots=dotnet-5-0#copy-jsonserializeroptions). but that would require conditional compilation, or dropping older runtimes

### DIFF
--- a/src/PlaywrightSharp/Helpers/JsonExtensions.cs
+++ b/src/PlaywrightSharp/Helpers/JsonExtensions.cs
@@ -82,6 +82,7 @@ namespace PlaywrightSharp.Helpers
                 {
                     new FlexibleStringEnumConverter<ResourceType>(ResourceType.Unknown),
                     new JsonStringEnumMemberConverter(JsonNamingPolicy.CamelCase),
+                    new EvaluateArgumentConverter(),
                 },
             };
     }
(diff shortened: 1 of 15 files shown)


src/PlaywrightSharp/Helpers/JsonExtensions.cs
```diff
@@ -73,6 +82,7 @@ internal static JsonSerializerOptions GetNewDefaultSerializerOptions(bool ignore
                 {
                     new FlexibleStringEnumConverter<ResourceType>(ResourceType.Unknown),
                     new JsonStringEnumMemberConverter(JsonNamingPolicy.CamelCase),
+                    new EvaluateArgumentConverter(),
```
kblok, line 85: We can't add this to all serializer calls.

SimonCropp: why? are there any side effect?

kblok: Yes, tests are failing because of that.
The EvaluateArgumentCoverter should be used only when we need to serialize evaluate arguments. That folk does crazy stuff.

PR #1172, opened 2021-02-21
use the range indexor

SimonCropp:
thoughts on this syntax?


kblok:
I would change only in the places where there is a proved bump in performance.
I think that `Substring` is way way more readable than an indexer, and I bet Substring must be a quite optimized function.

SimonCropp:
fair enough. TBH i am on the fence as well. will close this one

PR #1177, opened 2021-02-24
chore: update to final 1.9 driver

kblok:


src/PlaywrightSharp/PlaywrightSharp.csproj
```diff
@@ -60,8 +60,6 @@
     </ContentWithTargetPath>
   </ItemGroup>
   <Target Name="EnsurePrerequisitsRan" BeforeTargets="InstallBrowsers">
-    <Error
-      Text="Playwright prerequisites are missing. Ensure you've ran `dotnet run -p ./src/tools/PlaywrightSharp.Tooling/PlaywrightSharp.Tooling.csproj -- download-drivers --basepath .`"
-      Condition="!Exists('$(MSBuildProjectDirectory)\Drivers\unix$(NativeFolder)')" />
+    <Error Text="Playwright prerequisites are missing. Ensure you've ran `dotnet run -p ./src/tools/PlaywrightSharp.Tooling/PlaywrightSharp.Tooling.csproj -- download-drivers --basepath .`" Condition="!Exists('$(MSBuildProjectDirectory)\Drivers\unix$(NativeFolder)')" />
```
avodovnik, line 63: I don't necessarily like this change - it reduces readability.

src/tools/PlaywrightSharp.Tooling/DriverDownloader.cs
```diff
@@ -90,12 +91,7 @@ private static Task<string> GetUpstreamReadmeAsync(string playwrightVersion)
         private async Task DownloadDriverAsync(DirectoryInfo destinationDirectory, string driverVersion, string platform, string runtime)
         {
             Console.WriteLine("Downloading driver for " + platform);
-            string cdn = "https://playwright.azureedge.net/builds/driver";
-
-            if (driverVersion.Contains("next"))
-            {
-                cdn += "/next";
-            }
+            string cdn = "https://playwright.azureedge.net/builds/driver/next";
```
avodovnik, line 94: Is this right, for a non-next version of `<DriverVersion>1.9.0-1614037901000</DriverVersion>`?

src/PlaywrightSharp/PlaywrightSharp.csproj
kblok: yeah, IDE fight :(

src/tools/PlaywrightSharp.Tooling/DriverDownloader.cs
kblok: I took that from this change https://github.com/microsoft/playwright-python/pull/522/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L59

src/PlaywrightSharp/PlaywrightSharp.csproj
avodovnik: I'll let you decide, but I'd discard it and rebase 👼🏻

src/tools/PlaywrightSharp.Tooling/DriverDownloader.cs
avodovnik: Hm, @aslushnikov can you clarify? Are we now supposed to use `/next` for all driver downloads?

avodovnik: @kblok Andrey just confirmed this is right, so we're fine!

avodovnik (approved):

PR #1230, opened 2021-03-02
refactor(generator): accessibility

avodovnik:


src/PlaywrightSharp.Tests/PageAccessibilityContentEditableTests.cs
```diff
@@ -1,6 +1,8 @@
+using System.Linq;
 using System.Threading.Tasks;
 using PlaywrightSharp.Tests.Attributes;
 using PlaywrightSharp.Tests.BaseTests;
+using PlaywrightSharp.Tests.Internals;
```
kblok, line 5: Do we need this?

src/PlaywrightSharp/Contracts/Enums/MixedState.generated.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Text.Json;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace PlaywrightSharp
+{
+    public enum MixedState
+    {
+        Undefined = 0,
```
kblok, line 50: Don't you think that `None` could be more csharpy?

kblok (approved):

PR #1257, opened 2021-03-04
refactor(generator): dialog

### LINKED ISSUE #1239: Use generated IDialog

kblok:
closes #1239


src/PlaywrightSharp/DialogType.cs
```diff
@@ -6,27 +6,26 @@ namespace PlaywrightSharp
     /// Dialog type.
     /// </summary>
     /// <seealso cref="IDialog"/>
-    public enum DialogType
+    public static class DialogType
```
kblok, line 9: What do you think about this @avodovnik?

avodovnik: Looks good to me. I would consider moving it into `/Contracts/Models` though, despite not being generated.

avodovnik (approved):

PR #1258, opened 2021-03-04
refactor(generator): touchscreen

### LINKED ISSUE #1249: Use generated ITouchScreen

kblok:
closes #1249


src/PlaywrightSharp/Touchscreen.cs
```diff
@@ -36,6 +36,6 @@ public Touchscreen(PageChannel channel)
             _channel = channel;
         }
 
-        public Task TapAsync(Point point) => _channel.TouchscreenTapAsync(point);
+        public Task TapAsync(float x, float y) => _channel.TouchscreenTapAsync(x, y);
```
kblok, line 39: @avodovnik do you think we could add a Point overload or should we leave it this "pure".

avodovnik: Honestly, I don't see the value in writing `new Point(1,2)` vs `(1,2)`. This is the same discussion we had around `TimeZoneInfo`, and `FileInfo` etc. I'd keep it pure and if people *want* it, we can add it (or accept their PR ;-) ).

avodovnik (approved):

PR #1259, opened 2021-03-04
chore(generator): Mouse

### LINKED ISSUE #1244: Use generated IMouse

avodovnik:
Closes #1244


src/PlaywrightSharp/Mouse.cs
```diff
@@ -13,16 +12,26 @@ public Mouse(PageChannel channel)
...
 
-        public Task MoveAsync(decimal x, decimal y, int? steps = 1) => _channel.MouseMoveAsync(x, y, steps);
+        public Task MoveAsync(float x, float y, int? steps)
+            => _channel.MouseMoveAsync(x, y, steps ?? 1);
 
-        public Task UpAsync(MouseButton button = MouseButton.Left, int clickCount = 1) => _channel.MouseUpAsync(button, clickCount);
+        public Task UpAsync(MouseButton button, int? clickCount)
+            => _channel.MouseUpAsync(EnsureDefaultValue(button), clickCount ?? 1);
+
+        private static MouseButton EnsureDefaultValue(MouseButton button) =>
```
kblok, line 30: We could make it generic to check any enum.

src/PlaywrightSharp/Mouse.cs
```diff
@@ -13,16 +12,26 @@ public Mouse(PageChannel channel)
...
+            button switch
+            {
+                MouseButton.Undefined => MouseButton.Left,
```
kblok, line 33: Maybe fail with an `InvalidArgumentException`?

src/PlaywrightSharp/Mouse.cs
avodovnik: I can do that when I encounter the next one. :-)

src/PlaywrightSharp/Mouse.cs
avodovnik: Why would we fail? If it's `Undefined` (which is `default`) we want to default back to `left`. This is technically a problem caused by the fact that we add the `Undefined` in there, but we don't specify the actual default in the API specs. This is something we can potentially think of doing in the future. But either way, I wouldn't throw, *just yet*.

PR #1263, opened 2021-03-05
chore(generator): IWorker

### LINKED ISSUE #1252: Use generated IWorker

avodovnik:
Closes #1252


avodovnik:
We can't merge this, unless https://github.com/microsoft/playwright/pull/5730 merges, because we're expecting the method to become `async`.

src/PlaywrightSharp/Contracts/IWorker.generated.cs
```diff
@@ -0,0 +1,133 @@
...
+		/// function declaration, it is interpreted as a function. Otherwise, evaluated as an
+		/// expression.
+		/// </param>
+		/// <param name="arg">Optional argument to pass to <paramref name="expression"/>.</param>
+		Task<IJSHandle> EvaluateHandleAsync(string expression, object arg = null);
+	
+		string Url { get; }
+
+
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
```
kblok, line 123: Is this a bug in the generator?

src/PlaywrightSharp/Contracts/Models/UndefinedEvaluationArgument.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Represents an `Undefined` argument.
+    /// </summary>
+    public sealed class UndefinedEvaluationArgument : IEquatable<UndefinedEvaluationArgument>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UndefinedEvaluationArgument"/> class.
+        /// </summary>
+        private UndefinedEvaluationArgument()
+        {
+            // NOOP
```
kblok, line 17:
```suggestion
```

src/PlaywrightSharp/Worker.cs
```diff
@@ -53,48 +52,35 @@ public Worker(IChannelOwner parent, string guid, WorkerInitializer initializer)
...
+        }
 
-        /// <inheritdoc/>
-        public async Task<JsonElement?> EvaluateAsync(string expression, object arg)
-            => ScriptsHelper.ParseEvaluateResult<JsonElement?>(await _channel.EvaluateExpressionAsync(
-                script: expression,
-                isFunction: expression.IsJavascriptFunction(),
-                arg: arg,
-                serializeArgument: true).ConfigureAwait(false));
+        private object TranslateObject(object obj)
```
kblok, line 76: Why do we need this? Can't we use `EvaluateArgument.Undefined` and that's it?

src/PlaywrightSharp/Contracts/IWorker.generated.cs
avodovnik: No, it's due to the fact we're migrating one by one. The reference *will* be there, when we move all the interfaces over.

src/PlaywrightSharp/Worker.cs
avodovnik: We _could_, but it would mean making a whole bunch of the other stuff public, that's currently `internal` and I would much much rather have a separate object. It also doesn't mess with the logic of where it's used (i.e. it's not used for transport, but it gets translated) and that will bring it inline to where we want to be, mid-term future. Hence my decision to go with a separate one.

kblok (approved):

PR #1282, opened 2021-03-14
Write locations in "Driver not found in any of the locations"

SimonCropp:
To help debug when something goes wrong.

### DIFF
--- a/src/PlaywrightSharp/Transport/Connection.cs
+++ b/src/PlaywrightSharp/Transport/Connection.cs
@@ -258,20 +258,23 @@ namespace PlaywrightSharp.Transport
             }
 
             string executableFile = GetPath(driversPath);
-            if (!File.Exists(executableFile))
+            if (File.Exists(executableFile))
             {
-                // fallback to the bin path
-                executableFile = Path.Combine(
-                    driversPath,
-                    RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? "playwright.sh" : "playwright.cmd");
-
-                if (!File.Exists(executableFile))
-                {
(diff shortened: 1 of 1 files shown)


SimonCropp:
note that we had this today. the reason was we were pulling in `<PackageReference Include="Verify.Playwright" Version="0.11.0" />` but not an explicit  `<PackageReference Include="PlaywrightSharp" Version="0.192.0" />`. so the `.playwright` directory wasnt created. adding the explicit PlaywrightSharp ref fixed it. this wasnt required in v 0.180. not sure if it a bug or a qwerk of nuget. but just a heads up

SimonCropp:
guessing it has something to do with no `PlaywrightSharp.props` in 192

kblok:
I'm good with this change. Regarding your comment about the `.playwright`.  That's a known limitation. The output project needs to reference Playwright in order to get the `.playwright`folder.
I'm open to seeing a PR improving that.

kblok (approved):

src/PlaywrightSharp/Transport/Connection.cs
```diff
@@ -258,20 +258,23 @@ private static string GetExecutablePath()
...
+                driversPath,
+                RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? "playwright.sh" : "playwright.cmd");
+
+            if (File.Exists(fallbackBinPath))
+            {
+                return fallbackBinPath;
             }
 
-            return executableFile;
+            throw new PlaywrightSharpException($@"Driver not found in any of the locations. Tried:
```
avodovnik, line 275: I'm personally not a huge fan of having exceptions be in multiple lines.

avodovnik (approved):

PR #1284, opened 2021-03-15
chore(generator): IPlaywright

### LINKED ISSUE #1246: Use generated IPlaywright

avodovnik:
closes #1246


src/PlaywrightSharp/Selectors.cs
```diff
@@ -19,21 +16,10 @@ internal Selectors()
...
-        /// <param name="script">Script that evaluates to a selector engine instance.</param>
-        /// <param name="path">Path to the JavaScript file. If path is a relative path, then it is resolved relative to current working directory.</param>
-        /// <param name="content">Raw script content.</param>
-        /// <param name="contentScript">Whether to run this selector engine in isolated JavaScript environment. This environment has access to the same DOM, but not any JavaScript objects from the frame's scripts. Defaults to false. Note that running as a content script is not guaranteed when this engine is used together with other registered engines.</param>
-        /// <returns>A <see cref="Task"/> that completes when the engine is registered.</returns>
-        public async Task RegisterAsync(string name, string script = null, string path = null, string content = null, bool? contentScript = null)
+        /// <inheritdoc/>
+        public async Task RegisterAsync(string name, string script, string scriptPath, bool? contentScript = null)
```
kblok, line 20: Can we have some overloads?

src/PlaywrightSharp/Playwright.cs
```diff
@@ -55,23 +55,28 @@ internal Playwright(IChannelOwner parent, string guid, PlaywrightInitializer ini
         /// <summary>
         /// Returns a list of devices to be used with <see cref="IBrowser.NewContextAsync(BrowserContextOptions)"/>.
         /// </summary>
-        public IReadOnlyDictionary<string, DeviceDescriptor> Devices => _devices;
+        public IReadOnlyDictionary<string, DeviceDescriptor> Devices { get => _devices; set => throw new NotSupportedException(); }
```
kblok, line 58: hmmmm

src/PlaywrightSharp/Contracts/IPlaywright.generated.cs
```diff
@@ -0,0 +1,102 @@
...
+		public IBrowserType Chromium { get; set; }
+
+        /// <summary>
+        /// <para>
+        /// This object can be used to launch or connect to Firefox, returning instances of
+        /// <see cref="FirefoxBrowser" />.
+        /// </para>
+        /// </summary>
+        [JsonPropertyName("firefox")]
+		public IBrowserType Firefox { get; set; }
```
kblok, line 73: Why do we have a different indentation here?

kblok (changes requested):

src/PlaywrightSharp/Playwright.cs
avodovnik: I wonder if *all* of our generated properties should just be read-only?

kblok: I think so. The SetTimeout is still a method right? That would be the only one.

src/PlaywrightSharp/Selectors.cs
avodovnik: As discussed, I've changed the defaults in the interface, and for now, I want to keep the API as minimal as possible.

src/PlaywrightSharp/Playwright.cs
avodovnik: I will have to revisit this _later_. I need the properties to be generated when we have a serializable object (i.e. all the models). I don't want to block this PR in the mean time, though. As we've talked about a few times, when we finish the fist pass, we'll probably just re-add all the generated interfaces to fix the small discrepancies that are introduced as we change stuff on the generator. Let's keep this as is, then, for now, and we'll fix it before release.

src/PlaywrightSharp/Contracts/IPlaywright.generated.cs
avodovnik: User error :-)

src/PlaywrightSharp.Tests/TestConstants.cs
```diff
@@ -63,5 +63,23 @@ public static LaunchOptions GetHeadfulOptions()
             "        http://localhost:<PORT>/frames/frame.html (dos)",
             "    http://localhost:<PORT>/frames/frame.html (aframe)"
         };
+
+        internal static readonly BrowserContextOptions _iPhone6 = new BrowserContextOptions()
```
kblok, line 67: How about removing the `_`?

src/PlaywrightSharp.Tests/TestUtils.cs
```diff
@@ -83,7 +83,7 @@ internal static async Task RegisterEngineWithPathAsync(IPlaywright playwright, s
         {
             try
             {
-                await playwright.Selectors.RegisterAsync(name, path: path);
+                await playwright.Selectors.RegisterAsync(name, null, path);
```
kblok, line 86: We could do `scriptPath: path` right?

src/PlaywrightSharp/BrowserContextOptions.cs
```diff
@@ -273,5 +273,16 @@ public BrowserContextOptions Clone()
 
             return args;
         }
+
+        private BrowserContextOptions CopyFrom(BrowserContextOptions options)
```
kblok, line 277: If this is private we could just assign the values in the constructor, what do you think?

src/PlaywrightSharp/Playwright.cs
```diff
@@ -52,26 +44,26 @@ internal Playwright(IChannelOwner parent, string guid, PlaywrightInitializer ini
...
         IChannel<Playwright> IChannelOwner<Playwright>.Channel => _channel;
 
-        /// <summary>
-        /// Returns a list of devices to be used with <see cref="IBrowser.NewContextAsync(BrowserContextOptions)"/>.
-        /// </summary>
-        public IReadOnlyDictionary<string, DeviceDescriptor> Devices => _devices;
-
         /// <inheritdoc/>
-        public IChromiumBrowserType Chromium => _initializer.Chromium;
+        public IBrowserType Chromium { get => _initializer.Chromium; set => throw new NotSupportedException(); }
```
kblok, line 48: Can you see if we can assume read-only?

kblok (changes requested):

src/PlaywrightSharp.Tests/TestConstants.cs
avodovnik: Someone's style guides kicked in and made me rename :P 

But yeah, good catch. I'll make it nicer.

src/PlaywrightSharp/Playwright.cs
avodovnik: Hm, I don't think so, otherwise deserialization will start failing. I'm going to fix it, just not in this PR :)

src/PlaywrightSharp/BrowserContextOptions.cs
avodovnik: No, because the way it works right now is, it'll return the `options` instance, not the current one. I didn't want to mess with that behaviour too much *yet*, until I actually get around to `IBrowserContext` and related classes.

src/PlaywrightSharp/Playwright.cs
kblok: https://github.com/microsoft/playwright-sharp/issues/1286

src/PlaywrightSharp.Tests/Chromium/OopifTests.cs
```diff
@@ -129,7 +129,7 @@ public async Task ShouldEmulateOffline()
         [SkipBrowserAndPlatformFact(skipFirefox: true, skipWebkit: true)]
         public async Task ShouldSupportContextOptions()
         {
-            await using var context = await _browser.NewContextAsync(new BrowserContextOptions(Playwright.Devices["iPhone 6"])
+            await using var context = await _browser.NewContextAsync(new BrowserContextOptions(TestConstants.iPhone6)
```
kblok, line 132: Do we need this? can't we just pass `TestConstants.iPhone6`?

src/PlaywrightSharp.Tests/TestUtils.cs
```diff
@@ -104,7 +104,7 @@ internal static async Task RegisterEngineAsync(IPlaywright playwright, string na
         {
             try
             {
-                await playwright.Selectors.RegisterAsync(name, script, contentScript: contentScript);
+                await playwright.Selectors.RegisterAsync(name, script, null, contentScript);
```
kblok, line 107: The previous code should work, Shouldn't it?

kblok (approved):

PR #1289, opened 2021-03-22
Add installation notes

timheuer:
Added step to install using NuGet package


kblok:
Don't you think that it will be quite painful keeping the `--version` updated?

timheuer:
> Don't you think that it will be quite painful keeping the `--version` updated?

As long as the the released version is the one you want folks to have, then yeah you could drop the version argument.

kblok:
> As long as the the released version is the one you want folks to have, then yeah you could drop the version argument.

Yeah, let's remove it.

kblok (approved):

PR #1314, opened 2021-04-22
feat(page): introduce WaitForUrlAsync

### LINKED ISSUE #1299: Implement WaitForURLAsync feature

kblok:
closes #1299


src/Playwright/Contracts/Extensions/Converters.cs
```diff
@@ -0,0 +1,15 @@
+namespace Microsoft.Playwright
+{
+    internal static class Converters
+    {
+        internal static LoadState ToLoadState(this WaitUntilState waitUntilState)
```
avodovnik, line 5: Should we think about having an implicit converter instead?

src/Playwright/Contracts/IFrame.cs
```diff
@@ -61,5 +61,14 @@ public partial interface IFrame
 
         /// <inheritdoc cref="SelectOptionAsync(string, IEnumerable{SelectOptionValue}, bool?, float?)" />
         Task<IReadOnlyCollection<string>> SelectOptionAsync(string selector, params IElementHandle[] values);
+
+        /// <inheritdoc cref="WaitForURLAsync(string, Regex, Func{string, bool}, float?, WaitUntilState)"/>
+        Task WaitForURLAsync(string urlString, float? timeout = default, WaitUntilState waitUntil = default);
```
avodovnik, line 66: Just for context, is there a part of this that should be generated that I'm missing?

src/Playwright/Contracts/Extensions/Converters.cs
avodovnik: Blargh, you're right - it's an `enum`. No implicit conversion possible.

src/Playwright/Contracts/IFrame.cs
kblok: I would merge this folk. I created #1330 based on a conversation with @pavelfeldman, so we can generate these folks.

avodovnik (approved):

avodovnik (changes requested):
The new generator will generate additional overloads. Once that gets approved, we should also update this PR.

kblok:
We should close this one, right @avodovnik?

PR #1322, opened 2021-04-26
chore: format generated code

### LINKED ISSUE #1262: Can we generate code with spaces instead of tabs?
Let the war begin

kblok:
closes #1262


avodovnik:
I think this is a problem that we need to solve with the generator. Not in this project.

avodovnik:
Yeah, let's solve this in the generator step, upstream, and make sure these files are *correct* though. Also, as we do that, we'll update the interfaces anyway so we'll make sure all of them are freshly generated.

PR #1337, opened 2021-05-01
chore: use latest generator

kblok:
I closed a few PRs using the old directory location and replaced them with this one.


avodovnik (commented):
What version of the md files did you use for this? There's at least one [PR](https://github.com/microsoft/playwright/pull/6198) that wasn't merged yet, that has changes that are currently in the API that _we_ merged.

src/Playwright/Contracts/IBrowserType.generated.cs
```diff
@@ -197,12 +200,12 @@ public partial interface IBrowserType
         /// <param name="ignoreHTTPSErrors">Whether to ignore HTTPS errors during navigation. Defaults to <c>false</c>.</param>
         /// <param name="bypassCSP">Toggles bypassing page's Content-Security-Policy.</param>
         /// <param name="viewportSize">
-        /// Emulates consistent viewport for each page. Defaults to an 1280x720 viewport. Use
-        /// <c>ViewportSize.NoViewport</c> to disable the default viewport.
+        /// Emulates consistent viewport for each page. Defaults to an 1280x720 viewport. <c>null</c>
```
avodovnik, line 203: Upstream docs aren't right for this yet, either my PR gets merged, or we'll need to create a separate one with a separate field of `ignoreViewport` or something similar - which I'm strongly against. Either way, let's wait for that PR decision, and re-run this generator.

avodovnik (changes requested):

src/Playwright/Contracts/IBrowserType.generated.cs
kblok: Let's wait

PR #1338, opened 2021-05-04
chore(docs): run generator directly from msbuild

avodovnik:
Adds a target to allow for faster running of interface generation.


kblok:
When is this target going to be called?

avodovnik:
This is purely convenience. I'm torn if I really want to see it in the csproj or not.

kblok (approved):

PR #1339, opened 2021-05-05
chore: bump driver to 1.11

kblok:


src/Playwright.Tests/ScreencastTests.cs
```diff
@@ -11,6 +11,7 @@
 
 namespace Microsoft.Playwright.Tests
 {
+    /*
```
avodovnik, line 14: Should we leave the test, but skip it, and comment the code, so we don't forget?

avodovnik (commented):
Also, remove the binary file (FIREFOX/test.png).

src/Playwright.Tests/ScreencastTests.cs
kblok: I have an issue for that. It will be my next task

kblok:
Images removed and ignored @avodovnik

src/Playwright.Tests/ScreencastTests.cs
avodovnik: Ok, that's fine then.

.gitignore
```diff
@@ -352,3 +352,6 @@ MigrationBackup/
 src/Playwright.sln.DotSettings
 src/Playwright/Drivers
 src/Playwright/playwright
+src/Playwright.Tests/Screenshots/CHROMIUM/test.png
+src/Playwright.Tests/Screenshots/FIREFOX/test.png
+src/Playwright.Tests/Screenshots/WEBKIT/test.png
```
avodovnik, line 357: I think we could also just say `src/Playwright.Tests/Screenshots/**/*.png` to be a bit more generic here?

kblok: Done!
I'm leaving the `test.png` because it's the only one we need to ignore there.

avodovnik: Yeah, that's fine. I'm just thinking of how we can minimize changing this in the future :)

avodovnik (changes requested):
Let's look at the failing tests, then we should be ready to go!

avodovnik (approved):

PR #1347, opened 2021-05-05
chore(download): implement artifacts

### LINKED ISSUE #1328: Implement new Artifacts object

### LINKED ISSUE #1340: Restore download tests
Ignored on #1339

kblok:
based on #1339 
closes #1328 
closes #1340


src/Playwright/Transport/Protocol/ArtifactInitializer.cs
```diff
@@ -0,0 +1,6 @@
+namespace Microsoft.Playwright.Transport.Protocol
+{
+    internal class ArtifactInitializer
```
avodovnik, line 3: Do we need this for deserialization purposes? Any way we can avoid having an empty class?

avodovnik (approved):

src/Playwright/Transport/Protocol/ArtifactInitializer.cs
kblok: We don't need it :)

kblok: removed

avodovnik (approved):

PR #1369, opened 2021-05-12
chore: remove non-generic WaitForEventAsync

kblok:
generated using https://github.com/microsoft/playwright/pull/6535


kblok:
@pavelfeldman, the actual non-generic version of this method was approved by Dmitry back in the `v0 something` days.

I could show you why we need this `PlaywrightEvent` class if you want to. But on the other hand, I'm wondering if the bindings need a `WaitForEvent` method if we make sure that they will have a `WaitFor` for each event.

PR #1385, opened 2021-05-16
@pavelfeldman chore(api): roll new API w/ docs, remove inheritdocs from impl

pavelfeldman:


src/Playwright/WebSocketFrame.cs
```diff
@@ -1,9 +1,10 @@
 using System;
 using System.Text;
 
+#pragma warning disable CS1591, SA1600
```
mxschmitt, line 4: something like that might work for the entire solution/project instead of adding it to every file: https://docs.microsoft.com/de-de/previous-versions/visualstudio/visual-studio-2015/ide/how-to-suppress-compiler-warnings?view=vs-2015&redirectedfrom=MSDN#suppressing-warnings-for-visual-basic

mxschmitt (approved):

src/Playwright/Artifact.cs
```diff
@@ -6,6 +6,8 @@
 using Microsoft.Playwright.Transport.Channels;
 using Microsoft.Playwright.Transport.Protocol;
 
+#pragma warning disable CS1591, SA1600
```
avodovnik, line 9: This should go into https://github.com/microsoft/playwright-sharp/blob/main/src/.editorconfig#L178

avodovnik (changes requested):
1) I do not see the value of removing `<inheritdoc />`. @pfeldman, can you help me understand? 
2) The pragmas should not be included in every file but can be put in the `.editorconfig` as per the comment.

src/Playwright/WebSocketFrame.cs
pavelfeldman: Yeah, but I think I only want it for internal namespaces :/

src/Playwright/Artifact.cs
pavelfeldman: Let me try this...

pavelfeldman:
> I do not see the value of removing <inheritdoc />. @pfeldman, can you help me understand?

I don't see the value of adding them there at the first place. Also, in some cases (LaunchOptions) they hardcode the signature via the link, hence making rolling impossible.

> The pragmas should not be included in every file but can be put in the .editorconfig as per the comment.

I did not find a way to only disable those in Playwright/* - we should probably move all those classes into the Impl subfolder and have a dedicated .editorconfig for them.

PR #1390, opened 2021-05-18
test: fix race condition in wait for url test

kblok:


src/Playwright/Waiter.cs
```diff
@@ -53,7 +55,7 @@ internal void Log(string log)
         {
             _logs.Add(log);
             var logArgs = new { info = new { waitId = _waitId, phase = "log", message = log } };
-            Forget(_channel.Connection.SendMessageToServerAsync(_channel.Guid, "waitForEventInfo", logArgs));
+            _ = _channel.Connection.SendMessageToServerAsync(_channel.Guid, "waitForEventInfo", logArgs);
```
pavelfeldman, line 58: waitForResponse: false?

src/Playwright/Transport/Connection.cs
```diff
@@ -137,29 +140,19 @@ internal static async Task InstallAsync(string driverPath = null, string browser
...
-                TaskCompletionSource = tcs,
-                TreatErrorPropertyAsError = treatErrorPropertyAsError,
-            };
-
-            _callbacks.TryAdd(id, callback);
 
-            var st = new StackTrace(true);
-            var stack = new List<object>();
-            for (int i = 0; i < st.FrameCount; ++i)
+            if (waitForResponse)
```
pavelfeldman, line 144: What does this change?

src/Playwright/Waiter.cs
kblok: We don't want the connection to keep track of the response because we don't care about it.

pavelfeldman (approved):

PR #1392, opened 2021-05-18
chore: remove selected overrides

pavelfeldman:


src/Playwright/Request.cs
```diff
@@ -72,7 +72,7 @@ internal Request(IChannelOwner parent, string guid, RequestInitializer initializ
 
         public async Task<IResponse> ResponseAsync() => (await _channel.GetResponseAsync().ConfigureAwait(false))?.Object;
 
-        public JsonDocument PostDataJSON(JsonDocumentOptions documentOptions = default)
```
kblok, line 75: I consider this important to users. Without this argument, this method will be useless  to many times.

src/Playwright/Contracts/IPage.cs
```diff
@@ -8,9 +8,6 @@ namespace Microsoft.Playwright
 {
     public partial interface IPage
     {
-        /// <inheritdoc cref="EmulateMediaAsync(Media?, ColorScheme?)"/>
-        Task EmulateMediaAsync(ColorScheme? colorScheme);
```
kblok, line 12: Are tests running without this?

pavelfeldman: They compiled...

src/Playwright/Request.cs
pavelfeldman: I'm looking at the properties and they seem to be for lax parsing, which we don't need. MaxDepth - we should hard-code to -1, i'll do it.

PR #1398, opened 2021-05-19
chore: cleaning up optional arguments

avodovnik:


src/Playwright/Contracts/IElementHandle.cs
```diff
@@ -6,21 +6,21 @@ namespace Microsoft.Playwright
 {
     public partial interface IElementHandle
     {
-        Task SetInputFilesAsync(string files, bool? noWaitAfter = default, float? timeout = default);
```
pavelfeldman, line 9: Why are thee not generated?

src/Playwright/Contracts/IFrame.cs
```diff
@@ -14,22 +14,22 @@ public partial interface IFrame
 
         Task<JsonElement?> EvalOnSelectorAsync(string selector, string expression, object arg = default);
 
-        Task SetInputFilesAsync(string selector, string files, bool? noWaitAfter = default, float? timeout = default);
+        Task SetInputFilesAsync(string selector, string files, FrameSetInputFilesOptions options = default);
```
pavelfeldman, line 17: Same question here...

pavelfeldman:
I removed overrides in .md and it generated proper API for me: https://github.com/microsoft/playwright-sharp/pull/1400

PR #1402, opened 2021-05-20
chore: update external NuGet Packages

avodovnik:


src/Playwright/Playwright.csproj
```diff
@@ -65,6 +65,16 @@
...
+  <ItemGroup>
+    <PackageReference Update="Microsoft.VisualStudio.Threading.Analyzers" Version="16.9.60">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+    <PackageReference Update="Roslynator.Analyzers" Version="3.1.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
```
kblok, line 77: Why are we adding this? I love them, just curious, didn't we have them before?

kblok (approved):

src/Playwright/Playwright.csproj
avodovnik: I think the "new" bit is actually just the `<PrivateAssets>` tags, rather than the package reference :) Well, the version changed, but the include assets, etc., I haven't seen before... :-) They basically [ensure](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets) that "These assets will be consumed but won't flow to the parent project"

PR #1403, opened 2021-05-20
chore: roll driver to 1.11.1-1621490832000

avodovnik:


src/tools/PlaywrightSharp.Tooling/DriverDownloader.cs
```diff
@@ -155,10 +156,12 @@ private async Task<bool> ExecuteAsync()
 
                 var tasks = new List<Task>();
 
+                /*
```
avodovnik, line 159: I had to comment this out for now, because the release isn't actually created. We might have more of these in the future, so I'll think more about how to sync this (if at all).

kblok: How about default to master on `UpdateBrowserVersionsAsync`, if we don't find a matching tag?

avodovnik:
The README "update" part was broken due to https://github.com/microsoft/playwright-sharp/commit/556f1e71e02115682751ab18e95ae212787f5819. I've added the `<!-- GEN -->` tags back.

src/tools/PlaywrightSharp.Tooling/DriverDownloader.cs
avodovnik: Good point, I've added a fallback to `.0` and exception catching with a warning printout, rather than failing the whole process.

avodovnik:
@kblok I'm fairly certain this is a matter of flaky tests, so I'm OK with rolling...

What likely happens is that due to parallelization (cc: @dgozman) the threads get stuck/slow and the driver cancels the select operation, which for us, results in the task actually getting marked as `IsCompleted = true` but its `status = Faulted`. 

What I don't yet understand is if there's a _proper_ fix for this. I'm fairly sure setting the timeout to a higher number will "fix" it, but... Any thoughts?

kblok (approved):

PR #1405, opened 2021-05-20
fix(page): fix Page.PauseAsync

kblok:


dgozman (approved):
Let's add a test where we do `await page.Pause()` in headless that does nothing except for sending some protocol messages.

dgozman (approved):

PR #1406, opened 2021-05-21
feat(dialogs): auto-dismiss dialogs without event handlers

dgozman:


src/Playwright/Page.cs
```diff
@@ -94,7 +94,17 @@ internal Page(IChannelOwner parent, string guid, PageInitializer initializer) :
             _channel.Route += Channel_Route;
             _channel.FrameAttached += Channel_FrameAttached;
             _channel.FrameDetached += Channel_FrameDetached;
-            _channel.Dialog += (_, e) => Dialog?.Invoke(this, e);
+            _channel.Dialog += (_, e) =>
+            {
+                if (Dialog == null)
+                {
+                    _ = e.DismissAsync().ConfigureAwait(false);
```
kblok, line 101: No need of a `ConfigureAwait` without an `await` :)
```suggestion
                    _ = e.DismissAsync();
```

src/Playwright/Page.cs
```diff
@@ -94,7 +94,17 @@ internal Page(IChannelOwner parent, string guid, PageInitializer initializer) :
...
+            {
+                if (Dialog == null)
+                {
+                    _ = e.DismissAsync();
+                }
+                else
+                {
+                    Dialog?.Invoke(this, e);
+                }
+            };
```
avodovnik, line 107: You could also write it as:

```suggestion
           _channel.Dialog += (_, e) => (Dialog ?? new EventHandler<IDialog>((_, ex) => ex.DismissAsync()))?.Invoke(this, e);
```

to keep it condensed, but it's a preference :-).

Do we need similar behaviour anywhere else?

avodovnik (approved):

src/Playwright/Page.cs
dgozman: > Do we need similar behaviour anywhere else?

I don't think so.

PR #1424, opened 2021-05-23
chore: nuke isPage, fix bugs

pavelfeldman:


dgozman (approved):

src/Playwright.Tests/PageNetworkResponseTests.cs
```diff
@@ -63,7 +64,9 @@ public async Task ShouldReturnJson()
         public async Task ShouldWorkWithGenerics()
         {
             var response = await Page.GotoAsync(TestConstants.ServerUrl + "/simple.json");
-            Assert.Equal("bar", (await response.JsonAsync<TestClass>()).Foo);
+            JsonDocument doc = await response.JsonAsync();
```
kblok, line 67: The generic JsonAsync would be way more useful than this. I don't think people would use this with all these required boilerplate.

src/Playwright.Tests/PageNetworkResponseTests.cs
```diff
@@ -159,7 +162,7 @@ public async Task ShouldWaitUntilResponseCompletes()
 
         class TestClass
```
kblok, line 163: We can drop this class if we are not going to use it for tests.

src/Playwright/BrowserContext.cs
```diff
@@ -203,13 +203,14 @@ public Task SetExtraHTTPHeadersAsync(IEnumerable<KeyValuePair<string, string>> h
...
+            var options = new JsonSerializerOptions { PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
             string state = JsonSerializer.Serialize(
                 await Channel.GetStorageStateAsync().ConfigureAwait(false),
-                Channel.Connection.GetDefaultJsonSerializerOptions());
+                options);
 
             if (!string.IsNullOrEmpty(path))
             {
-                File.WriteAllText(path, state);
+                await File.WriteAllTextAsync(path, state).ConfigureAwait(false);
```
kblok, line 213: I think WriteAllTextAsync is net 5 only.

dgozman (approved):

src/Playwright.Tests/PageNetworkResponseTests.cs
pavelfeldman: > The generic JsonAsync would be way more useful than this. 

I don't this force-converting to camel case is going to have a single use, default one can at least be used...

src/Playwright/BrowserContext.cs
pavelfeldman: Yup, figured that!

src/Playwright.Tests/PageNetworkResponseTests.cs
pavelfeldman: Will follow-up

PR #1426, opened 2021-05-24
chore: add support for dotnet tool/CLI

avodovnik:


src/Playwright.CLI/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+            var pwProcess = new Process()
+            {
+                StartInfo = playwrightStartInfo,
+            };
+
+            playwrightStartInfo.EnvironmentVariables.Add("PW_CLI_TARGET_LANG", "csharp");
+            pwProcess.Start();
+
+            pwProcess.WaitForExit();
+            Console.WriteLine(pwProcess.StandardOutput.ReadToEnd());
```
dgozman, line 35: I think we should inherit stdio streams to get incremental input/output/errors, instead of writing something at the end.

src/Playwright.CLI/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        }
+
+        private static string GetFullPath()
+        {
+            string envPath = Environment.GetEnvironmentVariable(DriverEnvironmentPath);
```
dgozman, line 41: Who sets this env variable?

README.md
```diff
@@ -101,6 +101,29 @@ await page.RouteAsync("**", (route, _) =>
...
+Then, you can simply call:
+
+```powershell
+playwrightcli codegen https://www.microsoft.com
+```
+
+### Driver Discoverability
+
+By default, the tool will attempt to look at `%USERPROFILE%\.nuget\packages\microsoft.playwright` for the location of the driver. If that behaviour
+is undesired, you can set the `PW_CLI_DRIVERPATH` environment variable, to specify the location. The location needs to be *specific* and point to the
```
dgozman, line 124: I don't think anyone should set this variable that points to internal Playwright file.

README.md
```diff
@@ -101,6 +101,29 @@ await page.RouteAsync("**", (route, _) =>
...
+
+Then, you can simply call:
+
+```powershell
+playwrightcli codegen https://www.microsoft.com
+```
+
+### Driver Discoverability
+
+By default, the tool will attempt to look at `%USERPROFILE%\.nuget\packages\microsoft.playwright` for the location of the driver. If that behaviour
```
dgozman, line 123: Will the tool pick up Playwright from my project, if I am at `src/`?

src/Playwright.CLI/Program.cs
kblok: Do we need that having `RedirectStandardOutput` set to `true`?

avodovnik: @kblok the default is `false` - so I think we do, if we want to see the input/output

src/Playwright.CLI/Program.cs
avodovnik: This would be the user calling the tool. If there's none set, we get a `null` and do _fallback_.

src/Playwright.CLI/Program.cs
kblok: sorry. I mean, if we have `RedirectStandardOutput = true` do we need to write the output to the console. Doesn't `RedirectStandardOutput` do that for us?

avodovnik: So, I assumed that too, but it didn't. I'll give it another go, though.

avodovnik: Heh, look at that, turns out you have to call `BeginOutputReadLine`, to get anything on the event handlers :D

avodovnik: This is done, now. For now, I'm _assuming_ that what I get is a direct `WriteLine`, so I don't support some scenarios, like the app clearing the output. But I think this will be fine for v1.

README.md
avodovnik: I've changed the behaviour and doc now:

- traverse current folder and find `.playwright` 
- NuGet cache

src/Playwright.CLI/Program.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            using var outputWaitHandle = new AutoResetEvent(false);
+            using var errorWaitHandle = new AutoResetEvent(false);
+
+            pwProcess.OutputDataReceived += (_, e) =>
+            {
+                if (e.Data == null) outputWaitHandle.Set();
+                else
+                {
+                    Console.WriteLine(e.Data);
```
dgozman, line 44: I think data should come with all the new lines, so WriteLine -> Write?

src/Playwright.CLI/Program.cs
```diff
@@ -0,0 +1,164 @@
...
+                }
+            };
+
+            pwProcess.ErrorDataReceived += (_, e) =>
+            {
+                if (e.Data == null) errorWaitHandle.Set();
+                else
+                {
+                    Console.Error.WriteLine(e.Data);
```
dgozman, line 53: Same here.

src/Playwright.CLI/Program.cs
```diff
@@ -0,0 +1,164 @@
...
+                }
+            };
+
+            pwProcess.Start();
+
+            pwProcess.BeginOutputReadLine();
+            pwProcess.BeginErrorReadLine();
+
+            pwProcess.WaitForExit();
```
dgozman, line 62: [Stackoverflow](https://stackoverflow.com/questions/8808663/get-live-output-from-process) says this is going to block, and so we won't see live updates. Perhaps we should await for `process.Exited` event to unblock the thread so it can perform live streaming of the output? I am also not sure what `WaitOne(5000)` is for?

src/Playwright.CLI/Program.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        private static string GetFullPath()
+        {
+            string envPath = Environment.GetEnvironmentVariable(DriverEnvironmentPath);
+            if (!string.IsNullOrEmpty(envPath))
+            {
+                envPath = Path.Join(envPath, "Drivers");
+                if (!Directory.Exists(envPath))
+                {
+                    Console.Error.WriteLine($"The path specified in the environment variable ({envPath}) does not contain the Drivers folder.");
```
dgozman, line 75: Please mention the env var name here as well.

src/Playwright.CLI/Program.cs
avodovnik: It doesn't.

src/Playwright.CLI/Program.cs
avodovnik: It'll block the calling thread, yes, but it launches the process from a different thread and the callbacks happen on other threads:

image

The `WaitOne` is to ensure that we don't make the same mistake as our tests did initially - disposing stuff before we get all the content. We either wait for `null` to be sent as the payload, when the stream is done, or 5 seconds after the process exited which likely means there was a problem in the stream.

avodovnik: FWIW, the Process implementation [uses](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.Unix.cs#L214) `AsyncStreamReader` underneath, and that queues the work using [this](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs#L247). 

So, it's possible that we get a piece of text dispatched when the call to `WaitForExit`  already returns - hence the `AutoResetEvent` :-).

pavelfeldman (approved):

PR #1433, opened 2021-05-25
chore: simplify wait for options

kblok:
This is more to open the discussion @pavelfeldman 
I think we wouldn't need those extra classes. And if we would need them in the future we can solve that with an overload.


src/Playwright/BrowserContext.cs
```diff
@@ -253,7 +253,7 @@ public async Task<T> InnerWaitForEventAsync<T>(PlaywrightEvent<T> playwrightEven
                 waiter.RejectOnEvent<IBrowserContext>(this, ContextEvent.Close.Name, new PlaywrightException("Context closed"));
             }
 
-            var result = waiter.WaitForEventAsync<T>(this, playwrightEvent.Name, null);
+            var result = waiter.WaitForEventAsync<T>(this, playwrightEvent.Name, predicate);
```
kblok, line 256: We do need this @pavelfeldman

pavelfeldman: Yes! That's a bug!

PR #1442, opened 2021-05-26
Fix sample typo

martincostello:
The sample uses Firefox but it was labelled as Webkit.


pavelfeldman:
Note that these are overall obsolete in favor of https://playwright.dev/dotnet/docs/intro

martincostello:
Cool thanks, I'm literally just about to try this out so I'll use that to get started.

PR #1443, opened 2021-05-26
api: specialize waitForEvent

pavelfeldman:


src/Playwright.Tests/PageBasicTests.cs
```diff
@@ -204,8 +204,11 @@ public async Task ShouldWorkWithPageClose()
         [Fact(Timeout = TestConstants.DefaultTestTimeout)]
         public async Task ShouldFireLoadWhenExpected()
         {
+            var loadEvent = new TaskCompletionSource();
```
kblok, line 207: Let's make the tests .NET 3.1 friendly. If we don't want to be 3.1 friendly let's remove the build step.

```suggestion
            var loadEvent = new TaskCompletionSource<bool>();
```

src/Playwright.Tests/PageBasicTests.cs
```diff
@@ -204,8 +204,11 @@ public async Task ShouldWorkWithPageClose()
...
+            Page.Load += (_, _) => loadEvent.TrySetResult();
```
kblok, line 208: Let's make the tests .NET 3.1 friendly. If we don't want to be 3.1 friendly let's remove the build step.

```suggestion
            Page.Load += (_, _) => loadEvent.TrySetResult(true);
```

src/Playwright/API/Generated/Types/BrowserContextCookiesResult.cs
```diff
@@ -37,7 +37,7 @@
 namespace Microsoft.Playwright
 {
     /// <summary>
-    /// Result of calling <see cref="IBrowserContext.CookiesAsync"/>.
+    /// Result of calling <see cref="IBrowserContext.Cookies"/>.
```
dgozman, line 40: Why this change?

src/Playwright/API/Generated/Types/ElementHandleBoundingBoxResult.cs
```diff
@@ -37,7 +37,7 @@
 namespace Microsoft.Playwright
 {
     /// <summary>
-    /// Result of calling <see cref="IElementHandle.BoundingBoxAsync"/>.
+    /// Result of calling <see cref="IElementHandle.BoundingBox"/>.
```
dgozman, line 40: Why this change?

src/Playwright.Tests/PageBasicTests.cs
kblok: We get the same in other places. Let's make a decision and apply it across the board.

dgozman (approved):

src/Playwright.Tests/PageEventPopupTests.cs
```diff
@@ -79,27 +79,15 @@ public async Task ShouldEmitForImmediatelyClosedPopupsWithLocation()
         [Fact(Timeout = TestConstants.DefaultTestTimeout)]
         public async Task ShouldBeAbleToCaptureAlert()
```
kblok, line 80:
```suggestion
        public void ShouldBeAbleToCaptureAlert()
```

src/Playwright/API/Generated/Types/BrowserContextCookiesResult.cs
pavelfeldman: Ooh, this is not right!

src/Playwright.Tests/FrameHierarchyTests.cs
```diff
@@ -191,8 +191,11 @@ public async Task ShouldReportDifferentFrameInstanceWhenFrameReAttaches()
             }");
             Assert.True(frame1.IsDetached);
 
+            var frameEvent = new TaskCompletionSource<IFrame>();
+            Page.FrameNavigated += (_, frame) => frameEvent.TrySetResult(frame);
```
kblok, line 195: Comparing the two changes in this file. How would a customer know when to use `WaitForNavigationAsync` and when they would need to implement that mannualy?

PR #1460, opened 2021-05-30
Update BrowserContextStorageStateTests.cs

interiv:
bug fix


kblok (approved):
Thank you!

avodovnik (changes requested):
`ShouldRoundTripThroughTheFile` is failing in tests - not sure this change is correct

kblok:
@avodovnik. Yeah. I wouldn't merge this without the fix. But I'm working on that :)

avodovnik:
@kblok yeah, I think this change just surfaced a problem in the test, but regardless, we need to fix it before merging/approving it

avodovnik (approved):
We can merge this after #1465 merges!

Thanks for the contribution and spotting a problem!

PR #1463, opened 2021-05-31
chore: CLI cleanup

avodovnik:
- Remove redundant code
- Use the NUGET_PACKAGES env variable, if one exists (as per .NET docs)
- dotnet format & warning cleanups
- remove var vs explicit type


avodovnik:
Since we removed the fallback to the current directory, I added the option to override the `NUGET_PACKAGES` folder (see [docs](https://docs.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders)).

pavelfeldman (approved):

PR #1465, opened 2021-05-31
fix: storage state not handled correctly

avodovnik:


avodovnik:
This issue was surfaced by https://github.com/microsoft/playwright-sharp/pull/1460, once we merge it, we should merge that test change as well!

src/Playwright/Transport/Channels/BrowserChannel.cs
```diff
@@ -91,8 +91,21 @@ internal override void OnMessage(string method, JsonElement? serverParams)
...
+                    throw new PlaywrightException($"The specified storage state file does not exist: {storageStatePath}");
+                }
+
+                storageState = File.ReadAllText(storageStatePath);
+            }
+
+            if (!string.IsNullOrEmpty(storageState))
+            {
+                args.Add("storageState", JsonSerializer.Deserialize<StorageState>(storageState, Helpers.JsonExtensions.DefaultJsonSerializerOptions));
+            }
```
kblok, line 107: I would move this our of the channel. When we auto-generate the channel, it won't do this.

pavelfeldman: This looks good, but we need a test!

avodovnik: @pavelfeldman yeah, (a small) test was ~contributed~ fixed to surface this problem: https://github.com/microsoft/playwright-sharp/pull/1460 - I'm merging it now!

PR #1150, opened 2021-02-16
remove some redundant variables

SimonCropp:

PR #1466, opened 2021-06-01
fix(tests): test code cleanup, redundant variables

avodovnik:
Rebased and updated #1150


kblok:
It looks like InnerTextShouldThrow is failing across the board.

avodovnik:
@kblok thanks for looking at this - I've pushed the change we talked about

kblok (approved):

PR #1468, opened 2021-06-01
[Proposal] chore: clean-up tests

avodovnik:


avodovnik:
This was an attempt to see why we wouldn't want to use the existing `Assert.ThrowsAsync`, which is a bit weirdly name, and is actually [a sync method](https://github.com/nunit/nunit/issues/2843).

src/Playwright.NUnitTest/PlaywrightTest.cs
```diff
@@ -45,20 +45,6 @@ public async Task PlaywrightSetup()
             BrowserType = Playwright[BrowserName];
         }
 
-        public static async Task<T> AssertThrowsAsync<T>(Func<Task> action) where T : Exception
```
pavelfeldman, line 48: Let's add a note on why it was needed instead.

avodovnik:
This experiment only proved that we shouldn't be using the built-in functions. I'm closing this PR and a new one is open to add some docs and a link for future reference to the helper. See #1472.

PR #1469, opened 2021-06-01
chore(tests): drop net48 from tests and fix worker contention

avodovnik:
At the moment, we kept getting port contention issues with our http server implementation. The current implementation used `workerIndex` which was mathematically correct, however, it wouldn't handle the use case of something already running on the agent using a port in that range (i.e. 8084). This PR attempts to replace it with incrementing the port number directly, and retrying (up to 5 times) on an already used port.


src/Playwright.Tests/BaseTests/HttpService.cs
```diff
@@ -1,25 +1,50 @@
...
-            var workerIndex = test.WorkerIndex;
             return await test.RegisterService("Http", async () =>
             {
-                var http = new HttpService();
-                http.Server = SimpleServer.Create(8081 + workerIndex * 2, TestUtils.FindParentDirectory("Playwright.Tests.TestServer"));
-                http.HttpsServer = SimpleServer.CreateHttps(8081 + workerIndex * 2 + 1, TestUtils.FindParentDirectory("Playwright.Tests.TestServer"));
-                await Task.WhenAll(http.Server.StartAsync(), http.HttpsServer.StartAsync());
-                return http;
+                int attempt = 0;
+                while (attempt <= 5)
```
pavelfeldman, line 22: I think we have a worker-related bug that we should fix instead.

avodovnik: I don't think so - the tests are configured to run as [jobs](https://docs.github.com/en/actions/learn-github-actions/introduction-to-github-actions#jobs), and GitHub will run them in parallel. Specifically, there are two jobs, the .NET 5 tests for everything, and .NET 3.1 for Chromium on Windows. The workers will still have the same index, so there would reasonably be contention for that port number.

I'm confident this PR is the way to go, but we _should_ also invest some time into housecleaning for the tests, reconfiguring the run parameters, making them faster, etc.

avodovnik: On that note, @kblok - can you help me understand the reasoning for having that separate .NET 3.1 job for Windows & Chromium? That sounds like a legacy thing and I want to make sure I understand it before I remove it :D

kblok: There some breaking changes from .NET 3.1 to .NET 5 that I wanted to test. But I considered testing 3.1 in all browsers and all platworm too much waste of time. I just wanted to confirm that we work on 3.1

pavelfeldman (approved):

PR #1472, opened 2021-06-02
chore: remove duplicate code, add docs to helper

avodovnik:


kblok (approved):

pavelfeldman:
Sorry, my intention was the opposite - I don't think we should offer our customers this patched method. So I would just delete the one in PlaywrightTest.

avodovnik:
> Sorry, my intention was the opposite - I don't think we should offer our customers this patched method. So I would just delete the one in PlaywrightTest.

Oh! All the code right now uses that (in our tests). I'll move this back to the other file, then, and mark internal. :-)

kblok:
conflicted @avodovnik

avodovnik:
> conflicted @avodovnik

Yes, not ready to be merged yet. I'll mark it as draft for now, to avoid a mishap.

pavelfeldman (approved):

src/Playwright.Tests/BaseTests/PlaywrightAssert.cs
```diff
@@ -46,7 +49,7 @@ public static class TestExtensions
             }
         }
 
-        public static void DebugLog(this PlaywrightTest test, string text)
+        internal static void DebugLog(string text)
         {
             TestContext.Progress.WriteLine(text);
         }
```
kblok, line 55:
```suggestion
        internal static void DebugLog(string text) => TestContext.Progress.WriteLine(text);
```

kblok (approved):

src/Playwright.Tests/.runsettings
```diff
@@ -0,0 +1,5 @@
+<RunSettings>
+   <NUnit>
+       <NumberOfTestWorkers>24</NumberOfTestWorkers>
```
pavelfeldman, line 3: I have 5! Let's leave defaults that count number of cores.

pavelfeldman (approved):

src/Playwright.Tests/.runsettings
avodovnik: This one shouldn't have been commited. Removing...

PR #1473, opened 2021-06-02
chore: push NUnit as NuGet

avodovnik:
- rename the namespace back to NUnit
- introduce a common version props file
- enable signing & strong name in the new project
- add the new project to pipeline YML


src/Common/Version.props
```diff
@@ -0,0 +1,21 @@
+<Project>
+  <PropertyGroup>
+    <AssemblyVersion>1.11.1</AssemblyVersion>
```
kblok, line 3: I think we don't need to keep those two projects in sync. We might need to update Nunit only when we change something there or there is a breaking change in PW.

src/Playwright.NUnitTest/Playwright.NUnit.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <AssemblyName>Microsoft.Playwright.NUnit</AssemblyName>
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+    <IsPackable>true</IsPackable>
+    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+  </PropertyGroup>
+  <Import Project="../Common/Version.props" />
+  <Import Project="../Common/SignAssembly.props" />
+  <ItemGroup>
+    <ProjectReference Include="..\Playwright\Playwright.csproj" />
```
kblok, line 27: One think we should do is declaring this as a nuget dependency.

pavelfeldman:
I called it NUnitTest originally because when I used NUnit.Foo in the code, compiler yelled at me for clash between NUnit.Foo and Microsoft.Playwright.NUnit.Foo, effectively making the use of the full-qualified NUnit.Foo impossible. Was that a glitch on my side?

avodovnik:
@pavelfeldman Yes, it yelled at me too. You can tell it to be [quiet](https://github.com/microsoft/playwright-sharp/pull/1473/files#diff-5a8526cb33dc7409785f36da0c6a464bca2829cac0fa48a8767801ce11e8402aR30) by using `global::` :-). 

Having a `.NUnit` suffix seems more natural to me. There are also packages that do the same, like [SpecFlow](https://www.nuget.org/packages/SpecFlow.NUnit/), I don't want to call it "Test" because it implies it's an actual _test_. 

I'd rather we keep it `.NUnit`. The clash, like above is a common thing, and it's usually not a problem. But, if we wanted to play it safe, we could call ourselves `Microsoft.Playwright.NUnitExtensions` or something to that tune.

src/Playwright.NUnitTest/Playwright.NUnit.csproj
avodovnik: That happens automatically apparently. Looking at the produced nuget package:

image

src/Common/Version.props
avodovnik: Yes, that was my thought too, *but* it has a dependency on Playwright. While it's technically a `>=` dependency, it felt like keeping them in-sync would be the right bet.

pavelfeldman (approved):

PR #1478, opened 2021-06-04
chore: remove esprima dependency

pavelfeldman:


src/Playwright/Transport/Connection.cs
```diff
@@ -195,6 +195,7 @@ internal static async Task InstallAsync(string driverPath = null, string browser
 
                 string messageString = JsonSerializer.Serialize(message, GetDefaultJsonSerializerOptions());
                 _logger?.LogInformation($"pw:channel:command {messageString}");
+                Console.WriteLine(messageString);
```
avodovnik, line 198:
```suggestion
                System.Diagnostics.Debug.WriteLine(messageString);
```

If we want to keep it, we can use the `Debug` output,  which gets ignored for `Release` builds, but it can offer insight for us. Otherwise, we should remove it.

src/Playwright.Tests/Helpers/DoubleExtensions.cs
```diff
@@ -0,0 +1,17 @@
+using System;
```
avodovnik, line 1: I assume this is here because you moved the one in `Playwright` to `internal`, which makes sense. I'd almost prefer we use `InternalsVisibleTo` and we can avoid code duplication. I can fix that separately after we merge this one though.

avodovnik (approved):
Just remove the `Console.WriteLine` before merging, and we're good to go!

src/Playwright.Tests/Helpers/DoubleExtensions.cs
kblok: We removed the `InternalsVisibleTo` flag because we were testing internal APIs that were supposed to be public.

src/Playwright/Transport/Connection.cs
kblok: Let's drop it.

PR #1482, opened 2021-06-04
chore: use language features to make code nicer

avodovnik:


src/Playwright.Tests/BeforeUnloadTests.cs
```diff
@@ -21,7 +21,7 @@ public async Task ShouldRunBeforeunloadIfAskedFor()
             var dialogEvent = new TaskCompletionSource<IDialog>();
             newPage.Dialog += (_, dialog) => dialogEvent.TrySetResult(dialog);
 
-            var pageClosingTask = newPage.CloseAsync(new PageCloseOptions { RunBeforeUnload = true });
+            var pageClosingTask = newPage.CloseAsync(new() { RunBeforeUnload = true });
```
kblok, line 24: I am just wondering whether it might be too new for the users to be able to learn from our tests.

pavelfeldman (approved):

src/Playwright.Tests/BeforeUnloadTests.cs
avodovnik: I had that same thought initially, but then... We should encourage usage of new language features. And if users care about what's happening, they'll have a look at the method signature and see that it's options.

PR #1493, opened 2021-06-06
chore: drop xunit remainder

pavelfeldman:


src/Playwright.Tests/Playwright.Tests.csproj
```diff
@@ -7,21 +7,23 @@
         <NoWarn>1701;1702</NoWarn>
         <AssemblyName>Microsoft.Playwright.Tests</AssemblyName>
         <RootNamespace>Microsoft.Playwright.Tests</RootNamespace>
+        <GenerateProgramFile>false</GenerateProgramFile>
     </PropertyGroup>
     <Import Project="../Common/SignAssembly.props" />
     <ItemGroup>
+		<PackageReference Include="CommandLineParser" Version="2.9.0-preview1" />
```
avodovnik, line 14: Formatting.

avodovnik (changes requested):
The `PlaywrightTestAttribute` needs to be accessible from the tooling, otherwise it doesn't work. I've got a PR that I'm already working on, that will fix this correctly. I'd almost suggest you close this one, and I'll integrate the rest of your suggested changes (the pupeteer mention removals).

src/Playwright.Tests/Playwright.Tests.csproj
pavelfeldman: I copied it from one file to another, looks like our formatting is inconsistent.

pavelfeldman:
> The PlaywrightTestAttribute needs to be accessible from the tooling, otherwise it doesn't work. 

I don't think you followed my change!

src/Playwright.Tests/Playwright.Tests.csproj
avodovnik: Yeah, I think that's VS Mac vs VS inconsistencies.

avodovnik (approved):

PR #1498, opened 2021-06-07
feat(cli): move cli code into core

pavelfeldman:


src/Playwright/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        public static void Main(string[] args)
+        {
+            var p = new Program();
+            p.Run(args);
+        }
+
+        public void Run(string[] args)
+        {
+            string path = Directory.GetCurrentDirectory();
```
dgozman, line 43: I don't see this `path` being used anywhere. Am I missing something?

src/Playwright/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            string pwPath = null;
+            try
+            {
+                pwPath = Paths.GetExecutablePath();
+            }
+            catch (Exception)
+            {
+                Console.WriteLine("Please install Playwright:");
+                Console.WriteLine("   dotnet add package Microsoft.Playwright");
```
dgozman, line 65: How could this be the case? We are running code from the Microsoft.Playwright already.

src/Playwright/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+                }
+            };
+
+            pwProcess.Start();
+
+            pwProcess.BeginOutputReadLine();
+            pwProcess.BeginErrorReadLine();
+
+            pwProcess.WaitForExit();
+            outputWaitHandle.WaitOne(5000);
```
dgozman, line 119: I keep wondering about these 5000...

dgozman (approved):

src/Playwright/Program.cs
pavelfeldman: Ouch, I moved this to CLI and forgot about it.

src/Playwright/Program.cs
pavelfeldman: This can only be the case if we found a Microsoft.Playwright.dll with no .playwright. I should ask to build instead.

src/Playwright/Program.cs
pavelfeldman: Oh, I did not look at the code itself, I assumed it just worked. Let me follow up on this :)

PR #1501, opened 2021-06-07
feat(browser): do not install browsers by default

pavelfeldman:


dgozman (approved):

martincostello:
Just curious, what was the rationale for this change?

Playwright used to Just Work™️, but dependabot updates to 1.12.1 this morning have broken all my projects because now the drivers are missing due to this change in behaviour.

Needing to install the global tool and then run it between build and test seems a step backwards for usability and ease of use.

Issue #1504, opened 2021-06-08
Add nullable reference annotations.


Jason5Lee:
The playwright library has no clear documentation about whether an argument or return type can be null. In the original TypeScript library it's not a problem since the type system of TypeScript supports nullable annotation. For example, the `page.contentText` has a `string | null` type, which indicates that it may be null. With `strictNullCheck` compiler option, you can't call its methods without proper null check.

![image](https://user-images.githubusercontent.com/25118271/121101918-592be880-c82f-11eb-9e5d-4af50bc73fcb.png)

However, in this project, the nullable annotation is absent, neither the documentation points out the fact that it can be null.

![Screen Shot 2021-06-08 at 8 00 40 AM](https://user-images.githubusercontent.com/25118271/121102134-c63f7e00-c82f-11eb-9fc1-4e8862f99e5a.png)

Actually [C# has a nullable annotation](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies) in its new version. Please properly annotate them like the TypeScript library, or at least document them.

pavelfeldman:
We can definitely generate nullable annotations. Do you have experience consuming them? In our setting, they only were in our way due to the testing nature of the client. It is easier to make "Dereference of a possibly null reference" while testing - it it was in fact null, test would fail. What is your client's nature?

Another question is that Options bags are 95% of our API, but sadly nullable references don't improve the nullability. There is no way to make compile error when required property is missing while using late property initialization construction (See https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies#late-initialized-properties-data-transfer-objects-and-nullability).

pavelfeldman:
ok, let's see how this works!

PR #1505, opened 2021-06-08
api: generate nullable annotations

pavelfeldman:
#1504

### DIFF
--- a/src/Playwright/API/Generated/Enums/ColorScheme.cs
+++ b/src/Playwright/API/Generated/Enums/ColorScheme.cs
@@ -24,6 +24,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Drawing;
 using System.Globalization;
 using System.IO;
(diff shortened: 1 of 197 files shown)


dgozman (approved):

src/Playwright.Tests/Playwright.Tests.csproj
```diff
@@ -8,6 +8,7 @@
         <AssemblyName>Microsoft.Playwright.Tests</AssemblyName>
         <RootNamespace>Microsoft.Playwright.Tests</RootNamespace>
         <GenerateProgramFile>false</GenerateProgramFile>
+        <Nullable>enable</Nullable>
```
avodovnik, line 11: If we enable this on the main project too, are there too many errors? That way we could eliminate the pragma directives in each file...

avodovnik (commented):
We should get rid of the warnings in the tests before merging this one!

src/Playwright.Tests/Playwright.Tests.csproj
pavelfeldman: Hundreds. Those are all warnings btw.

PR #1510, opened 2021-06-09
fix(cli): don't traverse fs if started in a random folder

pavelfeldman:


src/Playwright.CLI/Program.cs
```diff
@@ -32,21 +32,41 @@ static class Program
...
+                    path = Path.Combine(path, "..");
+                }
+
                 args = args[2..];
             }
 
-            var file = Traverse(new DirectoryInfo(root));
+            if (File.GetAttributes(path).HasFlag(FileAttributes.Directory))
+            {
+                var solutions = Directory.GetFiles(path, "*.sln");
```
avodovnik, line 55: I'm not sure about this... It's possible you'd clone a repo with a `src` folder, and you'd be in the root folder.

src/Playwright.CLI/Program.cs
```diff
@@ -32,21 +32,41 @@ static class Program
...
+                var projects = Directory.GetFiles(path, "*.*proj");
+                if (solutions.Length == 0 && projects.Length == 0)
+                {
+                    return PrintError($"Couldn't find project using Playwright. Ensure a project or a solution exists in {path}, or provide another path using -p.");
```
avodovnik, line 59: Nit: it's the same string as above, maybe unify the check/error somewhere?

src/Playwright/Program.cs
```diff
@@ -47,8 +47,7 @@ public int Run(string[] args)
             }
             catch (Exception)
             {
-                Console.WriteLine("Microsoft.Playwright is missing required assets. Please ensure to build your project before running.");
-                return 1;
+                return PrintError("Microsoft.Playwright build was found, but is missing required assets. Please ensure to build your project before running Playwright tool.");
```
avodovnik, line 50: Not sure what `build was found` means in this context.  Maybe package?
```suggestion
                return PrintError("Microsoft.Playwright package was found, but is missing required assets. Ensure your project is built, before running Playwright tool.");
```

src/Playwright.CLI/Program.cs
pavelfeldman: None of the dotnet build / test / run would work in that case, so I guess it is expected that you pass -p...

src/Playwright.CLI/Program.cs
pavelfeldman: M: Maturity

src/Playwright/Program.cs
pavelfeldman: assembly?

PR #1520, opened 2021-06-10
fix: driver discovery and folder consolidation

### LINKED ISSUE #1515: can no longer use PLAYWRIGHT_DRIVER_PATH as a workaround to use from LINQPad - can we add that support back, please?
Earlier (1.11.1-alpha-2) I was able to set the PLAYWRIGHT_DRIVER_PATH environment variable to point to the exact dir that contained playwright.cmd as a workaround for the fact that in LINQPad the existing logic didn't come up with the "right" path. This seems to have broken in 1.12.1

The actual path of the playwright.cmd that needs to get used is under $env:UserProfile\\.nuget\packages\microsoft.playwright\\[version]\Drivers\win-x64\native\playwright.cmd

This was the code I was using with 1.11.1-alpha-2 to set the env var for use in linqpad:

```csharp
var playwrightAssembly = typeof(Playwright).Assembly;
var driversPath = new FileInfo(playwrightAssembly.Location).Directory!.FullName;
while (Directory.Exists(Path.Combine(driversPath, "Drivers")) == false && Path.GetPathRoot(driversPath) != driversPath)
{
    driversPath = Path.GetDirectoryName(driversPath);
}
// since the default path it checks includes a .playwright dir that doesn't exist for us, need to point to the exact dir with the .cmd file
driversPath = Path.Combine(driversPath, "Drivers", ((RuntimeInformation.OSArchitecture == Architecture.X64) ? "win-x64" : "win-x86"), "native");
Environment.SetEnvironmentVariable("PLAYWRIGHT_DRIVER_PATH", driversPath);
```

AFAICT this support for PLAYWRIGHT_DRIVER_PATH got removed with https://github.com/microsoft/playwright-dotnet/commit/55867b054cf1df8969fd935d98d536e37a8669e1#diff-a387ff4bbdf5b27cc00330eae6e3c41df555197a80d35f8cc21dec46152e8d0cL289-L292

Now there's no workaround for pointing to the driver location, so trying to use from linqpad doesn't have a simple workaround and we get a PlaywrightException trying to Playwright.CreateAsync as it's looking for the driver under $env:UserProfile\.nuget\packages\Microsoft.Playwright\1.12.1\lib\net5.0\\.playwright\node\win32_x64\playwright.cmd

Message: 
`Driver not found: [userprofile dir here]\.nuget\packages\Microsoft.Playwright\1.12.1\lib\net5.0\.playwright\node\win32_x64\playwright.cmd
`
```
   at Microsoft.Playwright.Helpers.Paths.GetExecutablePath()
   at Microsoft.Playwright.Transport.Connection.GetProcess()
   at Microsoft.Playwright.Transport.Connection..ctor()
   at Microsoft.Playwright.Playwright.CreateAsync()
   at UserQuery.Main(), line 1
```

If we can add that back (not sure why the env var support needed to be killed off-hand?), users can then have a workaround for when the existing logic doesn't Just Work(tm)

avodovnik:
Closes #1515, #1527

### DIFF
--- a/src/Playwright/Helpers/Paths.cs
+++ b/src/Playwright/Helpers/Paths.cs
@@ -33,7 +33,13 @@ namespace Microsoft.Playwright.Helpers
     {
         internal static string GetExecutablePath()
         {
-            var assemblyDirectory = new DirectoryInfo(AppContext.BaseDirectory);
+            DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
+            if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, "Microsoft.Playwright.dll")))
+            {
+                var assemblyLocation = typeof(Playwright).Assembly.Location;
+                assemblyDirectory = new FileInfo(assemblyLocation).Directory;
+            }
+
             string executableFile = GetPath(assemblyDirectory.FullName);
             if (File.Exists(executableFile))
             {
--- a/src/Playwright/Helpers/Paths.cs
+++ b/src/Playwright/Helpers/Paths.cs
@@ -33,19 +33,24 @@ namespace Microsoft.Playwright.Helpers
     {
         internal static string GetExecutablePath()
         {
-            var assembly = typeof(Playwright).Assembly;
-            string driversPath = new FileInfo(assembly.Location).Directory.FullName;
-
-            string executableFile = GetPath(driversPath);
+            var assemblyDirectory = new FileInfo(typeof(Playwright).Assembly.Location).Directory;
+            string executableFile = GetPath(assemblyDirectory.FullName);
             if (File.Exists(executableFile))
             {
                 return executableFile;
             }
 
-            throw new PlaywrightException($@"Driver not found: {executableFile}");
+            // if the above fails, we can assume we're in the nuget registry
+            executableFile = GetPath(assemblyDirectory.Parent.Parent.FullName, "Drivers");
+            if (File.Exists(executableFile))
+            {
+                return executableFile;
+            }
+
+            throw new PlaywrightException($"Driver not found: {executableFile}");
         }
 
-        private static string GetPath(string driversPath)
+        private static string GetPath(string driversPath, string containingFolder = ".playwright")
         {
             string platformId;
             string runnerName;
@@ -69,7 +74,7 @@ namespace Microsoft.Playwright.Helpers
                 throw new PlaywrightException("Unknown platform");
             }
 
-            return Path.Combine(driversPath, ".playwright", "node", platformId, runnerName);
+            return Path.Combine(driversPath, containingFolder, "node", platformId, runnerName);
         }
     }
 }
--- a/src/Playwright/Helpers/Paths.cs
+++ b/src/Playwright/Helpers/Paths.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Playwright.Helpers
     {
         internal static string GetExecutablePath()
         {
-            var assemblyDirectory = new FileInfo(typeof(Playwright).Assembly.Location).Directory;
+            var assemblyDirectory = new DirectoryInfo(AppContext.BaseDirectory);
             string executableFile = GetPath(assemblyDirectory.FullName);
             if (File.Exists(executableFile))
             {
@@ -41,7 +41,7 @@ namespace Microsoft.Playwright.Helpers
             }
 
             // if the above fails, we can assume we're in the nuget registry
-            executableFile = GetPath(assemblyDirectory.Parent.Parent.FullName, "Drivers");
+            executableFile = GetPath(assemblyDirectory.Parent.Parent.FullName);
             if (File.Exists(executableFile))
             {
                 return executableFile;
@@ -50,7 +50,7 @@ namespace Microsoft.Playwright.Helpers
             throw new PlaywrightException($"Driver not found: {executableFile}");
         }
 
-        private static string GetPath(string driversPath, string containingFolder = ".playwright")
+        private static string GetPath(string driversPath)
         {
             string platformId;
             string runnerName;
@@ -74,7 +74,7 @@ namespace Microsoft.Playwright.Helpers
                 throw new PlaywrightException("Unknown platform");
             }
 
-            return Path.Combine(driversPath, containingFolder, "node", platformId, runnerName);
+            return Path.Combine(driversPath, ".playwright", "node", platformId, runnerName);
         }
     }
 }
--- a/src/Playwright/Playwright.csproj
+++ b/src/Playwright/Playwright.csproj
@@ -40,29 +40,30 @@
   </ItemGroup>
   <Target Name="DedupeDriver" AfterTargets="CopyFilesToOutputDirectory" BeforeTargets="CopyPlaywrightFilesToOutput">
     <ItemGroup>
-      <DriverPackage Include="DriversRaw\linux\package\**" />
-      <DriverNodeLicense Include="DriversRaw\linux\LICENSE" />
-      <DriverNodeLinux Include="DriversRaw\linux\node" />
-      <DriverNodeMac Include="DriversRaw\mac\node" />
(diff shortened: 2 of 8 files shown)


avodovnik:
We still need to figure out *why* #1515 happens - the `Path.Combine` output seems wrong and it's a separate problem.

src/Playwright/Playwright.csproj
```diff
@@ -27,6 +27,7 @@
   <Import Project="../Common/SignAssembly.props" />
   <Import Project="build/Microsoft.Playwright.targets" />
   <ItemGroup>
+    <PackageReference Include="lvk.naturally" Version="1.0.2" />
```
avodovnik, line 30: I'm not sure I want to introduce a new dependency for this, but this let's us target the "latest" nuget package of the current version, due to our semver strategy.

src/Playwright/Helpers/Paths.cs
```diff
@@ -34,18 +35,55 @@ internal static class Paths
         internal static string GetExecutablePath()
         {
             var assembly = typeof(Playwright).Assembly;
-            string driversPath = new FileInfo(assembly.Location).Directory.FullName;
+            string driversPath, executableFile;
 
-            string executableFile = GetPath(driversPath);
+            // in .NET 6, the notion of single-assembly apps means, we won't have a location
```
pavelfeldman, line 40: It is unlikely that we support single-assembly apps, because we require driver and browser, so I would remove this and close single-assembly as Won't fix.

src/Playwright/Helpers/Paths.cs
```diff
@@ -34,18 +35,55 @@ internal static class Paths
...
+                executableFile = GetPath(driversPath);
+                if (File.Exists(executableFile))
+                {
+                    return executableFile;
+                }
+            }
+
+            // fallback to the NuGet repository
+            driversPath = GetNugetPackagePath(assembly.GetName().Version.ToString(3));
+            executableFile = GetPath(driversPath, "Drivers");
```
pavelfeldman, line 55: Can we use .playwright to align? This Drivers vs .playwright thing confuses me all the time!

src/Playwright/Helpers/Paths.cs
```diff
@@ -34,18 +35,55 @@ internal static class Paths
...
+
+                executableFile = GetPath(driversPath);
+                if (File.Exists(executableFile))
+                {
+                    return executableFile;
+                }
+            }
+
+            // fallback to the NuGet repository
+            driversPath = GetNugetPackagePath(assembly.GetName().Version.ToString(3));
```
pavelfeldman, line 54: ToString(3)! Nice.

src/Playwright/Helpers/Paths.cs
```diff
@@ -34,18 +35,55 @@ internal static class Paths
...
+        {
+            string packagesPath = Environment.GetEnvironmentVariable("NUGET_PACKAGES")
+                ?? Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".nuget", "packages");
+
+            string packagePath = Path.Combine(packagesPath, "microsoft.playwright");
+            if (Directory.Exists(packagePath))
+            {
+                var path = Directory
+                    .GetDirectories(packagePath, version + "*")
+                    .OrderByDescending(x => x, new Naturally.NaturalSortOrderStringComparer())
```
pavelfeldman, line 74: I don't think we should be doing this. In the case we are interested in, assembly itself belongs to nuget registry, so you just need to pick a corresponding driver.

pavelfeldman (changes requested):

src/Playwright/Playwright.csproj
pavelfeldman: We no longer need it, but in general, I would definitely not introduce a dependency for something that can be implemented in three lines, it only adds a security risk.

src/Playwright/Helpers/Paths.cs
jamesmanningrapidscale: Completely wishlist, but at least for the linqpad case, typeof(Microsoft.Playwright.Playwright).Assembly.Location is going to give us the location in the extracted package of the assembly, and 2 up from there is the extracted package root.  Going based on NUGET_PACKAGES and ~/.nuget/packages seems more limited (and has to do this version logic) than just "support running from an extract nuget dir" which would work for any location on disk, although I understand the single-assembly apps means needing to fallback to nuget anyway.

I guess my preference would be a 
1) normal build output - .playwright dir at assembly location
2) extracted nuget package, works regardless of version or extracted location
3) nuget package, needed for single-assembly apps

Something like this (likely not exactly this, just to get the idea across)

```csharp
var driversDirectoryPathsToTry = new List<string>();

var assembly = typeof(Microsoft.Playwright.Playwright).Assembly;
// in .NET 6, the notion of single-assembly apps means, we won't have a location
// and this can throw an exception - let's guard against it
if (!string.IsNullOrEmpty(assembly.Location))
{
    var assemblyParentDir = Path.GetDirectoryName(assembly.Location);

    // normal build output
    driversDirectoryPathsToTry.Add(Path.Combine(assemblyParentDir, ".playwright"));
    // running from extracted nuget dir
    driversDirectoryPathsToTry.Add(Path.Combine(assemblyParentDir, "..", "..", "Drivers"));
}
var driversPathUnderNuget = GetDriversPathUnderNugetPackagePath();
if (!string.IsNullOrEmpty(driversPathUnderNuget))
{
    driversDirectoryPathsToTry.Add(driversPathUnderNuget);
}

foreach (var pathToTry in driversDirectoryPathsToTry)
{
    var runnerPath = Path.Combine(pathToTry, "node", platformId, runnerName);
    if (File.Exists(runnerPath))
    {
        return runnerPath;
    }
}
throw new PlaywrightException($"Microsoft.Playwright drivers not found, tried in paths: {String.Join(", ", driversDirectoryPathsToTry)}");
```

Then iterate over the list and return the first path where

avodovnik: Yeah, that's what I've done now. I think this should cover your use case. As per @pavelfeldman's suggestion, we're not going to fix the single-file just yet, so no use in including that fix in here. Let's fix this first. :)

src/Playwright/Helpers/Paths.cs
avodovnik: The latest change should make them supported - I added a separate build target for `Publish` which copies over the `.playwright` folder. That said, I did add an `#ifdef`, but it also makes our discovery easier.

src/Playwright/Helpers/Paths.cs
avodovnik: Yes. Done.

avodovnik:
- Replaces:

`DriversRaw` -> `.drivers`
`Drivers` -> `.playwright`

- Replaces driver discovery with `AppContext.BaseDirectory` which also works with published applications.
- Adds target for after `Publish` that copies the `.playwright` folder

src/Playwright/build/Microsoft.Playwright.targets
```diff
@@ -21,4 +21,13 @@
     <Message Text="[Playwright] Copying shell script from $(MSBuildThisFileDirectory) to $(OutputPath)..." />
     <Copy SourceFiles="@(_CopyItems)" DestinationFiles="@(_CopyItems->'$(OutDir)\%(RecursiveDir)%(Filename)%(Extension)')"/>
   </Target>
+  <Target Name="PlaywrightCopyAfterPublish" AfterTargets="Publish">
```
pavelfeldman, line 24: It feels like we are giving a bit of false hope that this will work for users. It also breaks their existing builds. Should we let them do it themselves?

pavelfeldman (approved):

src/Playwright/build/Microsoft.Playwright.targets
avodovnik: I've tested this and it will work in the Single-file example, because the `.playwright` folder gets copied next to the file, and the `BaseDirectory` path change means the correct root folder is picked up.

PR #1526, opened 2021-06-11
chore: update CLI targeting and add requirements to readme

### LINKED ISSUE #1525: What versions of .NET does this work with?
I got asked if this works with 3.1 or 5 or both. It doesn't say anywhere in the docs.

avodovnik:
Fixes #1525

### DIFF
--- a/src/Playwright.CLI/Playwright.CLI.csproj
+++ b/src/Playwright.CLI/Playwright.CLI.csproj
@@ -5,7 +5,7 @@
     <PackageId>Microsoft.Playwright.CLI</PackageId>
     <Summary>The Playwright CLI dotnet tool.</Summary>
     <Description>Playwright enables reliable end-to-end testing for modern web apps. It is built to enable cross-browser web automation that is ever-green, capable, reliable and fast. Learn more at https://playwright.dev/dotnet/.</Description>
-    <TargetFrameworks>net5.0</TargetFrameworks>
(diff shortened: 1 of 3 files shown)


avodovnik:
Note, I've targeted `netcoreapp3.0` here, because it's the minimum where `range` is supported. I don't think we need to go any lower unless we get actual requests for it.

kblok (approved):

src/Playwright.CLI/Playwright.CLI.csproj
```diff
@@ -20,8 +20,8 @@
     <IsPackable>true</IsPackable>
   </PropertyGroup>
   <PropertyGroup>
-    <AssemblyVersion>1.0.0</AssemblyVersion>
-    <PackageVersion>$(AssemblyVersion)-alpha-5</PackageVersion>
+    <AssemblyVersion>1.0.1</AssemblyVersion>
```
pavelfeldman, line 23: We should not land stable package versions in trunk, so this should become -alpha-6. You won't even be able to cherry-pick it like this, and for a good reason.

pavelfeldman (changes requested):
If that is the only reason, I'd replace this one line with a 3-line for loop. I had done it locally.

src/Playwright.CLI/Playwright.CLI.csproj
avodovnik: Changed it to `-next-1`.

PR #1529, opened 2021-06-11
chore: extract internal types to Core namespace, expose BrowserType names

avodovnik:


src/Playwright/BrowserTypes.cs
```diff
@@ -0,0 +1,23 @@
+namespace Microsoft.Playwright
+{
+    /// <summary>
+    /// Contains the browser types names.
+    /// </summary>
+    public static class BrowserTypes
```
pavelfeldman, line 6: Let's do the right thing, move internal classes to the Microsoft.Playwright.Internal namespace and get a nice name for this one.

avodovnik: @pavelfeldman I moved them all to `.Core`. I don't like `.Internal` :-) That said, I failed at finding a better name for `BrowserTypes` - open to suggestions! I like it this way, because it's related to `BrowserType`...

pavelfeldman:
Let's make sure this is only about the Core move and rename the change.

PR #1535, opened 2021-06-15
chore: be a good citizen and perform a cleanup task

avodovnik:
Clean-up our Drivers folder, potentially the DriversRaw folder for devs, and the incorrectly copied node and package folders. It also fixes an issue where we copied `node` and `package` folder over to the root of the output dir, instead of inside the `.playwright` folder.

### DIFF
--- a/src/Playwright/build/Microsoft.Playwright.targets
+++ b/src/Playwright/build/Microsoft.Playwright.targets
@@ -16,10 +16,10 @@
   </Target>
   <Target Name="CopyPlaywrightShellToOutput" AfterTargets="CopyFilesToOutputDirectory">
     <ItemGroup>
-      <_CopyItems Include="$(MSBuildThisFileDirectory)playwright.ps1" />
+      <_CopyItemsShell Include="$(MSBuildThisFileDirectory)playwright.ps1" />
     </ItemGroup>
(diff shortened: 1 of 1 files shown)


src/Playwright/build/Microsoft.Playwright.targets
```diff
@@ -31,11 +31,24 @@
...
+    <Message Text="[Playwright] Removing up old Drivers folder..."/>
     <RemoveDir
-        Directories="$(MSBuildProjectDirectory)\Drivers" />
+        Directories="$(MSBuildProjectDirectory)\Drivers" Condition="Exists('$(MSBuildProjectDirectory)\Drivers')" />
+    <RemoveDir
+        Directories="$(MSBuildProjectDirectory)\DriversRaw" Condition="Exists('$(MSBuildProjectDirectory)\DriversRaw')" />
+    <RemoveDir
+      Directories="$(OutDir)\node" Condition="Exists('$(OutDir)\node')"/>
+    <RemoveDir
+          Directories="$(OutDir)\package" Condition="Exists('$(OutDir)\package')"/>
```
kblok, line 44: super nit, we could align these `Directories`

kblok (approved):

PR #1536, opened 2021-06-15
feat: API proxy to CLI call for browser installation

avodovnik:


avodovnik:
After chatting to @pavelfeldman, I'm closing this PR. I'll wait for some more feedback and then, *if* we need it, I'll revisit this.

PR #1537, opened 2021-06-16
make Waiter IAsyncDisposable

SimonCropp:
avoids discarding the task from SendMessageToServerAsync

### DIFF
--- a/src/Playwright/Core/BrowserContext.cs
+++ b/src/Playwright/Core/BrowserContext.cs
@@ -263,7 +263,7 @@ namespace Microsoft.Playwright.Core
             }
 
             timeout ??= DefaultTimeout;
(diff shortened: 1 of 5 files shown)


pavelfeldman:
Could you elaborate? Those messages are supposed to be fire and forget. Does it cause errors? Can we silent-catch locally?

SimonCropp:
> Those messages are supposed to be fire and forget

so it is ok for them to continue running during appdomain shutdown?

pavelfeldman:
I believe so. Those are I/O against the driver connection. As we exit, driver process exits and the transport as terminated. So all the pending I/Os fail, including this one.

SimonCropp:
ok

PR #1538, opened 2021-06-16
fix: do not ignore screen size on NewContext creation

SimonCropp:
this is a guess on my part, since it seems weird that screen size is not used


avodovnik (approved):

avodovnik:
@SimonCropp I added the fix as I was fixing something else, so I've opened a new PR and made sure you were co-author mentioned. Hopefully that's OK!

SimonCropp:
no worries at all.

PR #1539, opened 2021-06-16
dont ignore handle bool in ExposeBindingAsync

SimonCropp:


avodovnik:
This looks like it's also been addressed with the latest batch of cleanups: https://github.com/microsoft/playwright-dotnet/pull/1544.

PR #1540, opened 2021-06-16
use some top level methods

SimonCropp:


kblok:
I think top-level statements are a nice onboarding/sandboxing/playground tool. I don't think that It will help to have a file with a different layout in this project.

avodovnik:
I'd potentially be OK with having this in the CLI project - because it's really just a simple "pass-through", but I don't think I'd want it in the main library. 

That said, using this doesn't give us any benefits that I'm aware of. It's great for small examples & experimentation but... what's the purpose of doing it here?

SimonCropp:
> I think top-level statements are a nice onboarding/sandboxing/playground tool. I don't think that It will help to have a file with a different layout in this project.

no worries. it was only a suggestion :)

PR #1543, opened 2021-06-16
fix: do not ignore screen size on NewContext creation

avodovnik:


avodovnik:
This needs to be rebased when we merge #1542!

src/Playwright/Core/Browser.cs
```diff
@@ -75,7 +75,8 @@ public async Task<IBrowserContext> NewContextAsync(BrowserNewContextOptions opti
                storageStatePath: options.StorageStatePath,
                timezoneId: options.TimezoneId,
                userAgent: options.UserAgent,
-               viewportSize: options.ViewportSize).ConfigureAwait(false)).Object;
+               viewportSize: options.ViewportSize,
+               screenSize: options.ScreenSize).ConfigureAwait(false)).Object;
```
pavelfeldman, line 79: Looks like another missing test, let's add one.

pavelfeldman (approved):

src/Playwright/Core/Browser.cs
avodovnik: https://github.com/microsoft/playwright/pull/7196 I've added them upstream too...

kblok (approved):

PR #1544, opened 2021-06-16
chore(adapters): cleanup - BrowserContext, BrowserType, Keyboard, Selectors

avodovnik:


src/Playwright/Core/BrowserContext.cs
```diff
@@ -155,8 +155,8 @@ public async Task CloseAsync()
 
         public Task<IReadOnlyList<BrowserContextCookiesResult>> CookiesAsync(IEnumerable<string> urls = null) => Channel.CookiesAsync(urls);
 
-        public Task ExposeBindingAsync(string name, Action callback, bool? handle = null)
-            => ExposeBindingAsync(name, _ => callback());
+        public Task ExposeBindingAsync(string name, Action callback, BrowserContextExposeBindingOptions options = default)
+            => ExposeBindingAsync(name, callback, handle: options?.Handle ?? false);
```
pavelfeldman, line 159: Is this a bug fix? It looks like one! Is there a test missing? Let's add one.

pavelfeldman (approved):

src/Playwright/Core/BrowserContext.cs
avodovnik: Sort of. We already *test* this when using the _correct_ overload ` public Task ExposeBindingAsync<TResult>(string name, Func<BindingSource, IJSHandle, TResult> callback)`, which _expects_ a handle, and we therefore pass it back. For other calls, we _defaulted_ to `false` for this. The reasoning being simple: the result coming back from the driver would be the handle and we'd serialize it anyway, so we'd need the type. 

TL;DR: this might not actually be an option we need to expose to the user.

kblok (approved):

src/Playwright/Core/BrowserContext.cs
avodovnik: So, this is a weird one. It's not actually a bug fix. It's a redundant call. Not sure what's worse. 

The thing is, in this particular method, the handle can never be used anyway, so even when the user requests it, and gets it back from the driver, it's inaccessible inside the `callback`. HOWEVER, because of this, I've prepared a [separate PR](https://github.com/microsoft/playwright-dotnet/pull/1562/files#diff-0c59668d9d913f09e435a88b88e3b3613e5f2021a837b028b63dcb2b710872d9) (that needs rebasing when this one merges). 

That one introduces the `options` as an optional parameter to the one method where you would reasonably expect the handle to be available, and added tests to reflect that. But, for this call in particular, I'm still torn if we should just leave it to not do anything, or even validate and throw an exception, saying it's not supported, if passed.

PR #1547, opened 2021-06-17
remove some test utils that are not used

SimonCropp:


src/Playwright.Tests/BaseTests/PlaywrightAssert.cs
```diff
@@ -48,7 +46,5 @@ internal static class PlaywrightAssert
                 return t;
             }
         }
-
-        internal static void DebugLog(string text) => TestContext.Progress.WriteLine(text);
```
pavelfeldman, line 52: Let's keep this one please, we use it for debugging.

pavelfeldman:
All these utility classes are copied from the main package. We should figure out something to reuse originals rather than further diverge them.

src/Playwright.Tests/BaseTests/PlaywrightAssert.cs
SimonCropp: done

avodovnik:
Some of this has already been removed in the recent changes, so I'm closing this PR as stale. Sorry about that.

PR #1550, opened 2021-06-17
fix some spelling

SimonCropp:


src/Playwright.Tests/DefaultBrowserContext1Tests.cs
```diff
@@ -249,7 +249,7 @@ public async Task ShouldSupportJavascriptEnabledOption()
 
         [PlaywrightTest("defaultbrowsercontext-1.spec.ts", "should support httpCredentials option")]
         [Test, Timeout(TestConstants.DefaultTestTimeout)]
-        public async Task ShouldRupportHttpCredentialsOption()
+        public async Task ShouldRepportHttpCredentialsOption()
```
pavelfeldman, line 252: I don't think Repport is much better than Rupport, I guess you want Support.

pavelfeldman (approved):

src/Playwright.Tests/DefaultBrowserContext1Tests.cs
SimonCropp: lol. true. fixed

PR #1552, opened 2021-06-17
fix null ref in binding call

SimonCropp:
the `_initializer?` on the line above imply `_initializer` can be null. So i assume we should null check before extracting args?

### DIFF
--- a/src/Playwright/Core/BindingCall.cs
+++ b/src/Playwright/Core/BindingCall.cs
@@ -42,15 +42,18 @@ namespace Microsoft.Playwright.Core
                     },
                 };
 
-                if (methodParams.Length == 1 && methodParams[0] == typeof(IJSHandle))
+                if (_initializer != null)
                 {
-                    args.Add(_initializer.Handle);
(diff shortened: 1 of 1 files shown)


src/Playwright/Core/BindingCall.cs
```diff
@@ -42,15 +42,18 @@ internal async Task CallAsync(Delegate binding)
                     },
                 };
 
-                if (methodParams.Length == 1 && methodParams[0] == typeof(IJSHandle))
+                if (_initializer != null)
```
pavelfeldman, line 45: It is never null, the call sites above are wrong.

SimonCropp: want me to fix the call sites in a diff pr?

pavelfeldman: Yes. Although I barely keep the pace approving those.

mxschmitt:
Closing since it seems stale and not needed anymore.

PR #1553, opened 2021-06-17
remove some redundant state machines

SimonCropp:


pavelfeldman:
What's the rationale behind the change? Also looks like bots aren't happy.

SimonCropp:
gimme some time to have a look at the bots

SimonCropp:
the rational is perf

pavelfeldman:
I know nothing about .NET, but I hope implications of this are negligible. Even if there is a microbenchmark that can detect this, I don't think we would see it in the Playwright setting. I'm Ok with accepting it if some code style / best practices suggest returning Task though.

SimonCropp:
i should note this approach is already used in many places in the code base. eg 

```
        private Task ExposeBindingAsync(string name, Delegate callback, bool handle = false)
        {
            if (Bindings.ContainsKey(name))
            {
                throw new PlaywrightException($"Function \"{name}\" has been already registered");
            }

            Bindings.Add(name, callback);

            return _channel.ExposeBindingAsync(name, handle);
        }
```

PR #1559, opened 2021-06-17
target typed new

SimonCropp:
want a PR that moves over to that approach?


SimonCropp:
i should note it is already sued in ~300 places in the code base. it is just not consistent

pavelfeldman:
Yes, we want it at least in the Tests project!

SimonCropp:
done

PR #1562, opened 2021-06-17
feat: add options to typed callback binding call

avodovnik:


src/Playwright.Tests/BrowserContextExposeFunctionTests.cs
```diff
@@ -129,5 +129,27 @@ public async Task ExposeBindingHandleLikeInDocumentation()
...
+                result.TrySetResult(await t.AsElement().TextContentAsync());
+            };
+
+            await Context.ExposeBindingAsync("clicked", act, new() { Handle = true });
+
+            await page.SetContentAsync("<script>\n" +
+             "  document.addEventListener('click', event => window.clicked(event.target));\n" +
+             "</script>\n" +
+             "<div>Click me</div>\n" +
+             "<div>Or click me</div>\n");
```
kblok, line 149: This will look prettier with `@""`

kblok (approved):

src/Playwright.Tests/BrowserContextExposeFunctionTests.cs
avodovnik: But the rest of the file has this syntax, so I'm keeping it. At some point, we need to standardize on using `@""` and fix *all* of these.

avodovnik:
This can only be merged after #1544, and needs rebasing before that, but I've opened it already to give us a discussion point.

dgozman (approved):
I am curios how this works if I pass the callback that does not match the `Handle` option? Specifically, `Action<BindingSource, int>` with `Handle=true` and `Action<BindingSource, IJSHandle>` with `Handle=false`. Should it just not compile instead? Can we make it so?

avodovnik:
> I am curios how this works if I pass the callback that does not match the `Handle` option? Specifically, `Action<BindingSource, int>` with `Handle=true` and `Action<BindingSource, IJSHandle>` with `Handle=false`. Should it just not compile instead? Can we make it so?

It's a weird one, yeah. I don't think a compile-time error is an option, because it would require ignoring the generated options (https://github.com/microsoft/playwright-dotnet/blob/5c7673763a13740ae52d14231b03a5e5077c3c44/src/Playwright/API/Generated/Options/BrowserContextExposeBindingOptions.cs#L60). 

What I would seriously consider doing, though, more and more, is implementing our own analysers that would warn the user.  We've talked about that before, with threading issues, so this might be the final push we need.

dgozman:
> It's a weird one, yeah. I don't think a compile-time error is an option, because it would require ignoring the generated options

Well, if it is more idiomatic to replace this generated option with template type parameter, why not do that? We can leave this `handle` option to non-dotnet ports.

PR #1565, opened 2021-06-18
remove redundant sn settings

SimonCropp:


pavelfeldman:
Looks like https://github.com/microsoft/playwright-dotnet/pull/1572#pullrequestreview-691094533 is fixing it.

PR #1566, opened 2021-06-18
chore(adapters): Route, Mouse, Frame, Page, ElementHandle

avodovnik:


kblok (approved):

src/Playwright/Core/ElementHandle.cs
```diff
@@ -50,96 +50,106 @@ internal ElementHandle(IChannelOwner parent, string guid, ElementHandleInitializ
...
+            options ??= new();
+            if (options.Type == null && !string.IsNullOrEmpty(options.Path))
             {
-                type = DetermineScreenshotType(path);
+                options.Type = DetermineScreenshotType(options.Path);
             }
 
-            byte[] result = Convert.FromBase64String(await _channel.ScreenshotAsync(path, omitBackground, type, quality, timeout).ConfigureAwait(false));
+            byte[] result = Convert.FromBase64String(await _channel.ScreenshotAsync(
+                options.Path,
```
pavelfeldman, line 81: Sometimes you pass them as named, but sometimes as positional args. What should it be? Also, every time I see options passed as is I feel like Channel could just serialize them reflectively.

pavelfeldman (approved):

PR #1572, opened 2021-06-18
fix(strong name): perform actual signing of the assembly

avodovnik:


pavelfeldman:
What does it do exactly?

avodovnik:
> What does it do exactly?

It switches us away from `PublicSign` which only uses the public key side of it, to doing it "properly". The _downside_ of that is, that it requires the *private* key to be available during build time - but for us, this isn't an actual problem. As previously stated, Strong Names aren't security, they're used for **identity** and our private key resides in the repo.

The problem with `PublicSign` itself is that it doesn't **actually** sign the assembly, which means that in some cases, the verification will fail (these cases are [pointed out in the .NET Core repo](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/public-signing.md)), like a partially trusted AppDomain. 

TL;DR: I'm proposing we do full signing. I'm keeping this PR as draft because I'm still researching if there **is** a downside I'm forgetting/not seeing.

KirillOsenkov:
Yes, this PR is the best way forward. Since you have the private key and public key pair (596 bytes, while only public key is 160 bytes), you can fully sign and disable PublicSign. This will be the best.

It's good to fully sign whenever possible, and in your case you're lucky since the private key is already in the repo (this is the recommended approach). When you fully sign, no need to either PublicSign or DelaySign.

One should never DelaySign. If you don't have the private key, prefer PublicSign since that works in most scenarios (except 1) GAC 2) partial trust and 3) appdomains).

It's important to have a strong name because it's viral and if someone is not strong named, they cause a lot of pain for all consumers. To use IVT the consumers have to be strong named (so you can specify the public key of the consumers that you want to expose your internal APIs to). Since Roslyn is strong named and relies on IVT heavily, it kind of "poisons" the entire ecosystem into strong naming.

To verify whether an assembly is fully signed, public signed or delay signed, you can use https://github.com/KirillOsenkov/CodeCleanupTools#lbi-listbinaryinfo

To convert an assembly from delay-signed to public-signed, you can use the FakeSign utility (the difference is a single bit, FakeSign.exe just flips the bit). Here's FakeSign: https://www.nuget.org/packages/fakesign

pavelfeldman (approved):

PR #1576, opened 2021-06-19
Params based AddCoogkies

SimonCropp:


src/Playwright/API/Generated/IBrowserContext.cs
```diff
@@ -163,6 +163,17 @@ public partial interface IBrowserContext
         /// </param>
         Task AddCookiesAsync(IEnumerable<Cookie> cookies);
 
+        /// <summary>
```
pavelfeldman, line 166: The folder Generated should have given you an idea :-). I am not sure what the params implications are for versioning, but it is not options friendly anyways, so we won't be able to add optional options to this overload. It does not seem to be worth it in the general case.

kblok:
That was an explicit decision when the surface was defined. I'm collecting `contrib` ideas. This is one of them https://github.com/hardkoded/playwright-sharp-contrib/issues/16 .

SimonCropp:
ok will close for now

Issue #986, opened 2020-11-07
[Feature]: Add BrowserType.connect()


andliang:
Can I connect to a browser instance by passing a web socket endpoint?

Playwright's Node API supports it as documented [here](https://github.com/microsoft/playwright/blob/master/docs/api.md#class-browserserver) (look for wsEndpoint.

kblok:
@andliang, Playwright bindings don't have the browser server implemented (yet).
Could you share your use case so it can be prioritized?

andliang:
I use browserless.io because Azure AppService doesn't allow chrome running in it. Browserless.io has sample code ([here](https://docs.browserless.io/docs/playwright.html)) on how to use their service using Node

kblok:
cc @arjunattam

kurtisane:
Any news on this ?

kblok:
@kurtisane there is an open issue upstream https://github.com/microsoft/playwright/issues/4687
You can go and give your thumb there.

Xeevis:
PlaywrightSharp doesn't have the `Connect` capability at all? I thought Playwright API does support it. My use-case is to use browserless hosted locally in a docker container to reliably consume multiple headless & isolated browser instances in parallel and with queues from multiple apps.

Guess I'll have to stick to PuppeteerSharp for the time being 😔

```csharp
var browser = await Puppeteer.ConnectAsync(new ConnectOptions() { BrowserWSEndpoint = "ws://localhost:3000" })
```

sourav-kundu:
Is there any way by which I can connect to a **wsEndpoint** using PlaywrightSharp?
Its supported in the node API 
https://playwright.dev/docs/api/class-browsertype#browsertypeconnectparams

kblok:
@sourav-kundu we will! not yet :)

kblok:
cc @pavelfeldman

avodovnik:
Moving this to `v1.14` as it's not looking like it'll get finished in time for 1.13.

bbhxwl:
> Can I connect to a browser instance by passing a web socket endpoint?
> 
> Playwright's Node API supports it as documented [here](https://github.com/microsoft/playwright/blob/master/docs/api.md#class-browserserver) (look for wsEndpoint.

How to deploy a remote browser server using docker?

PR #1602, opened 2021-07-04
feat: Support for connecting remote browser

Archish27:
fixes: #986


src/Common/Version.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <AssemblyVersion>1.13.0</AssemblyVersion>
+    <AssemblyVersion>1.14.0</AssemblyVersion>
```
kblok, line 3: This is managed by the team during the release.

```suggestion
    <AssemblyVersion>1.13.0</AssemblyVersion>
```

src/Playwright/API/Generated/IBrowserType.cs
```diff
@@ -69,6 +70,20 @@ public partial interface IBrowserType
...
+        /// Connect the remote browser with <paramref name="wsEndpoint"/>
+        /// and returns the remote browser.
+        /// </para>
+        /// <code>
+        /// var browser = await playwright.Chromium.ConnectAsync("ws://url");
+        /// </code>
+        /// </summary>
+        /// <param name="wsEndpoint">WebSocket Endpoing</param>
+        /// /// <param name="options">Connect options</param>
+        Task<IBrowser> ConnectAsync(string wsEndpoint, BrowserTypeConnectOptions? options = default);
```
kblok, line 85: Do we have a PR upstream generating this new API?

src/Playwright/Core/PlaywrightImpl.cs
```diff
@@ -76,6 +76,8 @@ internal PlaywrightImpl(IChannelOwner parent, string guid, PlaywrightInitializer
 
         public ISelectors Selectors { get => _initializer.Selectors; }
 
+        public PlaywrightInitializer Intitializer { get => _initializer; }
```
kblok, line 79: Why are we exposing this?

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +122,39 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
...
+            Browser browser = playwright.Intitializer.PreLaunchedBrowser;
+            browser.IsRemote = true;
+            browser.Disconnected += Browser_Disconnected;
+            return browser;
+        }
+
+        private void Browser_Disconnected(object sender, IBrowser e)
+        {
+            _parent.Connection.Close("Browser Disconnected");
+        }
```
kblok, line 158:
```suggestion
        private void Browser_Disconnected(object sender, IBrowser e)
            => _parent.Connection.Close("Browser Disconnected");
```

src/Playwright/Playwright.csproj
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
```
kblok, line 1: I think we could avoid all the changes in this file.

kblok (changes requested):

src/Playwright/Core/PlaywrightImpl.cs
Archish27: Aah yup initializer is not required. PreLaunchedBrowser is required which can be retrieved by `_initializer.PreLaunchedBrowser`

src/Playwright/API/Generated/IBrowserType.cs
Archish27: https://github.com/microsoft/playwright/pull/7469

kblok:
Just a reminder. Let's look for `connect` tests upstream and implement them here.

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +122,39 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
...
+        public async Task<IBrowser> ConnectAsync(string wsEndpoint, BrowserTypeConnectOptions options = null)
+        {
+            options ??= new BrowserTypeConnectOptions();
+
+            var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.SetMinimumLevel(LogLevel.Debug);
+                builder.AddFilter((f, _) => f == "PlaywrightSharp.Playwright");
+            });
+            WebSocketTransport webSocketTransport = new WebSocketTransport(wsEndpoint, options, loggerFactory);
```
kblok, line 135:
```suggestion
            var webSocketTransport = new WebSocketTransport(wsEndpoint, options, loggerFactory);
```

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +122,39 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
...
+            await webSocketTransport.ConnectAsync().ConfigureAwait(false);
+            _parent.Connection.Transport = webSocketTransport;
+
+            Connection connection = new Connection(webSocketTransport, loggerFactory);
```
kblok, line 139:
```suggestion
            var connection = new Connection(webSocketTransport, loggerFactory);
```

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +122,39 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
...
+            options ??= new BrowserTypeConnectOptions();
+
+            var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.SetMinimumLevel(LogLevel.Debug);
+                builder.AddFilter((f, _) => f == "PlaywrightSharp.Playwright");
+            });
+            WebSocketTransport webSocketTransport = new WebSocketTransport(wsEndpoint, options, loggerFactory);
+            await webSocketTransport.ConnectAsync().ConfigureAwait(false);
+            _parent.Connection.Transport = webSocketTransport;
```
kblok, line 137: Question here. Should I be able to `Launch` and `Connect` to different browsers at the same time?

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +122,39 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
...
+            var playwright = await connection.WaitForObjectWithKnownNameAsync<PlaywrightImpl>("Playwright").ConfigureAwait(false);
+            playwright.Connection = connection;
+
+            if (playwright.Intitializer.PreLaunchedBrowser == null)
+            {
+                _parent.Connection.Close("Disconnected");
+                throw new PlaywrightException("Malformed endpoint. Did you use launchServer method?");
+            }
+
+            Browser browser = playwright.Intitializer.PreLaunchedBrowser;
```
kblok, line 149:
```suggestion
            var browser = playwright.Intitializer.PreLaunchedBrowser;
```

kblok (changes requested):
looking good!

src/Playwright/Core/BrowserType.cs
Archish27: It won't work until different playwright objects are created. Since transport is acting as a shared variable the values can be conflicting. Hence, to differentiate different playwright objects need to be created then the above case will work.

kblok: I don't see Java, assigning the connection to the parent https://github.com/microsoft/playwright-java/blob/06b88b3d4b57e25ae021e516ff2036a9306f4624/playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java#L78

Archish27: I tried Java way transport was not getting picked as WebTransport, default transport is set by default as StdIOTransport when `BrowserType` instance is created throughout the execution StdIOTransport was been used. Hence, had to do such a change & set WebSocketTransport when connect being used.

src/Common/Version.props
Archish27: fixed

src/Playwright/Core/PlaywrightImpl.cs
Archish27: fixed

src/Playwright/Core/BrowserType.cs
Archish27: fixed

src/Playwright/Playwright.csproj
Archish27: fixed

src/Playwright/Core/BrowserType.cs
Archish27: fixed

src/Playwright/Core/BrowserType.cs
Archish27: fixed

src/Playwright/Core/BrowserType.cs
Archish27: fixed

mxschmitt:
Thank you for your contribution! 💯  Having the upstream tests mirrored, or at least the important ones like we did in Python would be very useful to ensure this feature works in the future and handles all the edge-cases. See here the code pointers for reference:

- https://github.com/microsoft/playwright/blob/master/tests/chromium/chromium.spec.ts#L96-L333
- https://github.com/microsoft/playwright/blob/master/tests/browsertype-connect.spec.ts
- https://github.com/microsoft/playwright-python/blob/master/tests/async/test_browsertype_connect.py
- https://github.com/microsoft/playwright-python/blob/master/tests/async/test_browsertype_connect_cdp.py

Archish27:
> Thank you for your contribution! 💯 Having the upstream tests mirrored, or at least the important ones like we did in Python would be very useful to ensure this feature works in the future and handles all the edge-cases. See here the code pointers for reference:
> 
> * https://github.com/microsoft/playwright/blob/master/tests/chromium/chromium.spec.ts#L96-L333
> * https://github.com/microsoft/playwright/blob/master/tests/browsertype-connect.spec.ts
> * https://github.com/microsoft/playwright-python/blob/master/tests/async/test_browsertype_connect.py
> * https://github.com/microsoft/playwright-python/blob/master/tests/async/test_browsertype_connect_cdp.py

How do I launch server in playwright dotnet there is no `LaunchServer` equivalent method? Also, what is the expected result? How should I test? I tried spawning test server which was available in `CLI` but unable to do so? Can you please guide here

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -109,9 +103,53 @@ public void ShouldTerminateNetworkWaiters()
...
+
+
+        private class BrowserServer
+        {
+            public Process process { get; set; }
+            public string wsEndpoint { get; set; }
+
+        }
+
+        private BrowserServer launchServer(IBrowserType browserType)
```
kblok, line 118:
```suggestion
        private BrowserServer LaunchServer(IBrowserType browserType)
```

src/Playwright/Core/PlaywrightImpl.cs
```diff
@@ -76,6 +76,8 @@ internal PlaywrightImpl(IChannelOwner parent, string guid, PlaywrightInitializer
 
         public ISelectors Selectors { get => _initializer.Selectors; }
 
+        public Browser PreLaunchedBrowser { get => _initializer.PreLaunchedBrowser; }
```
kblok, line 79:
```suggestion
        public Browser PreLaunchedBrowser => _initializer.PreLaunchedBrowser; 
```

And internal right?

src/Playwright/Helpers/Paths.cs
```diff
@@ -29,9 +29,9 @@
 
 namespace Microsoft.Playwright.Helpers
 {
-    internal static class Paths
+    public static class Paths
     {
-        internal static string GetExecutablePath()
+        public static string GetExecutablePath()
```
kblok, line 34: hmmm I think that I would rather duplicate the code than adding this to the API surface. Let's wait for more feedback here.

src/Playwright.Tests/BrowserTypeConnectTests.cs
Archish27: fixed

src/Playwright/Core/PlaywrightImpl.cs
Archish27: fixed

Archish27:
@pavelfeldman Can you please review?

src/Playwright/Playwright.cs
```diff
@@ -22,26 +22,68 @@
...
+            _playwrightServerProcess.Exited += (_, _) => Close();
+
+            var transport = new StdIOTransport(_playwrightServerProcess, loggerFactory);
+            var connection = new Connection(transport, loggerFactory);
             var playwright = await connection.WaitForObjectWithKnownNameAsync<PlaywrightImpl>("Playwright").ConfigureAwait(false);
             playwright.Connection = connection;
             return playwright;
         }
+
+        private static Process GetProcess()
```
avodovnik, line 63: This isn't the right place for this. The responsibility of the `Playwright` class should be incredibly simple. My first thought is to move the creation (and ownership) of the process into the `transport` class - because they're the only ones actually caring about it. So, basically, we'd end up with

`new Connection(new StdIOTransport())` 

or

`new Connection(new WebSocketTransport())`

and then the rest happens automagically.

src/Playwright/Transport/Connection.cs
```diff
@@ -85,13 +75,35 @@ public Connection()
 
         public bool IsClosed { get; private set; }
 
+        public IConnectionTransport Transport { get; set; }
```
avodovnik, line 78: This can be `internal`.

src/Playwright/Core/BrowserType.cs
avodovnik: Yeah, this "feels" wrong. Let's investigate why this is needed and fix the underlying cause?

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +121,36 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
...
+        {
+            options ??= new BrowserTypeConnectOptions();
+
+            var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.SetMinimumLevel(LogLevel.Debug);
+                builder.AddFilter((f, _) => f == "PlaywrightSharp.Playwright");
+            });
+            var webSocketTransport = new WebSocketTransport(wsEndpoint, options, loggerFactory);
+            await webSocketTransport.ConnectAsync().ConfigureAwait(false);
```
avodovnik, line 135: Since (I assume) we always have to call `ConnectAsync`, something like a `static WebSocketTransport.Connect : WebsocketTransport` would feel more natural.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        public event EventHandler<MessageReceivedEventArgs> MessageReceived;
+
+        public event EventHandler<LogReceivedEventArgs> LogReceived;
+
+        public event EventHandler<TransportClosedEventArgs> TransportClosed;
+
+        public bool IsClosed { get; private set; }
+
+        public async Task ConnectAsync()
```
avodovnik, line 47: See above comment. I wouldn't have any other `public` methods that aren't part of the interface, in this class.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,186 @@
...
+                return;
+            }
+
+            if (_webSocket != null)
+            {
+                _webSocket.Dispose();
+            }
+        }
+
+        private string GetUserAgent()
```
avodovnik, line 163: Nit: 
```suggestion
        private string GenerateUserAgent()
```

avodovnik (commented):
Thank you again for the contribution! I left a few comments - happy to work with you through them!

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -32,26 +37,16 @@ namespace Microsoft.Playwright.Tests
     public class BrowserTypeConnectTests : PlaywrightTestEx
```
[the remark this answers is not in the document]
avodovnik: BTW, this is the case for all files touched. @kblok did you have a particular approach to this before?

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Making it a private method & calling in a constructor itself

src/Playwright/Core/BrowserType.cs
Archish27: yes not required though since when a WebSocket object is created it will create connection & give back instance of websocket

src/Playwright/Core/Browser.cs
```diff
@@ -55,6 +55,8 @@ internal Browser(IChannelOwner parent, string guid, BrowserInitializer initializ
 
         public bool IsConnected { get; private set; }
 
+        public bool IsRemote { get; set; }
```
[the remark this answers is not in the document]
Archish27: removing it not required

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -109,9 +103,53 @@ public void ShouldTerminateNetworkWaiters()
...
+            public Process process { get; set; }
+            public string wsEndpoint { get; set; }
+
+        }
+
+        private BrowserServer LaunchServer(IBrowserType browserType)
+        {
+            try
+            {
+                var browserServer = new BrowserServer();
```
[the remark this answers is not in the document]
Archish27: This should be `BrowserServer browserServer = new();`
Have made change 👍

README.md
```diff
@@ -34,6 +35,25 @@ class Program
 }
 ```
 
+### Remote Browser
```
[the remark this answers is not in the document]
Archish27: So should I keep it or remove?

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -109,9 +103,53 @@ public void ShouldTerminateNetworkWaiters()
...
+
+
+        private class BrowserServer
+        {
+            public Process process { get; set; }
+            public string wsEndpoint { get; set; }
+
+        }
+
+        private BrowserServer LaunchServer(IBrowserType browserType)
```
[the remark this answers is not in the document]
Archish27: I have added [SetUp] & TearDown for it

src/Playwright/Core/BrowserType.cs
Archish27: @kblok I have resolved this issue. It was from the script side I wasn't closing `Browser`. Hence the connection was not getting set properly after adding it. Now we can run both connect & launch together.

README.md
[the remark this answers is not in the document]
avodovnik: Remove it. The docs go in a separate PR upstream.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -91,7 +98,6 @@ public void ShouldRejectNavigationWhenBrowserCloses()
         }
 
         [PlaywrightTest("browsertype-connect.spec.ts", "should reject waitForSelector when browser closes")]
-        [Test, Ignore("SKIP WIRE")]
```
avodovnik, line 94: Test that's empty should not have `Ignore` removed. Either implement the test, or leave the `Ignore`.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -31,27 +35,30 @@ namespace Microsoft.Playwright.Tests
     [Parallelizable(ParallelScope.Self)]
     public class BrowserTypeConnectTests : PlaywrightTestEx
     {
+        private BrowserServer _browserServer;
+
+        [SetUp]
+        public void SetUpAsync()
+        {
+            _browserServer = LaunchServer(BrowserType);
```
avodovnik, line 43: Is there a reason to still have the `LaunchServer` method then? The idea of a `SetUp` method is to have it set everything up. Let's avoid the option of someone launching a separate server, unless we really need it.

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +118,24 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
                 recordHarOmitContent: options.RecordHarOmitContent,
                 recordVideo: Browser.GetVideoArgs(options.RecordVideoDir, options.RecordVideoSize)).ConfigureAwait(false)).Object;
         }
+
+        public async Task<IBrowser> ConnectAsync(string wsEndpoint, BrowserTypeConnectOptions options = null)
+        {
+            options ??= new BrowserTypeConnectOptions();
+#pragma warning disable CA2000
```
avodovnik, line 125:
```suggestion
#pragma warning disable CA2000 // Dispose objects before losing scope
```

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +118,24 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
...
+            var transport = new WebSocketTransport(wsEndpoint, options);
+#pragma warning restore CA2000
+            var connection = new Connection(transport);
```
avodovnik, line 128:
```suggestion
#pragma warning disable CA2000 // Dispose objects before losing scope
            var connection = new Connection(new WebSocketTransport(wsEndpoint, options));
#pragma warning restore CA2000
```

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -109,9 +115,54 @@ public void ShouldTerminateNetworkWaiters()
...
         public void ShouldRespectSelectors()
         {
         }
+
+
+        private class BrowserServer
+        {
+            public Process Process { get; set; }
+            public string WSEndpoint { get; set; }
+
```
avodovnik, line 127:
```suggestion
```

avodovnik (commented):
This is looking really good! A few minor changes still needed, and I think we'll be ready to go!

src/Playwright.Tests/BrowserTypeConnectTests.cs
Archish27: fixed

src/Playwright/Core/BrowserType.cs
Archish27: fixed

src/Playwright/Core/BrowserType.cs
Archish27: fixed

src/Playwright.Tests/BrowserTypeConnectTests.cs
Archish27: Have removed method added in `SetUp`

src/Playwright.Tests/BrowserTypeConnectTests.cs
Archish27: Added back

src/Playwright.Tests/BrowserTypeConnectTests.cs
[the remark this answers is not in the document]
avodovnik: I'm resolving this for now, I'll fix it later for the whole solution.

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -129,6 +149,11 @@ private void Dispose(bool disposing)
                 _readerCancellationSource.Cancel();
                 _readerCancellationSource.Dispose();
             }
+
+            if (_process != null)
+            {
+                _process.Dispose();
+            }
```
avodovnik, line 156:
```suggestion
            _process?.Dispose();
```

avodovnik (approved):
I'm happy with this as it is. I added a minor nit, but nothing horrible left. We'll get it merged ASAP!

src/Playwright/Transport/StdIOTransport.cs
Archish27: Aahh  :(  Have a habit of Java for sure will change in upcoming PRs as I have to additional connect tests

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,194 @@
...
+        public event EventHandler<LogReceivedEventArgs> LogReceived;
+
+        public event EventHandler<TransportClosedEventArgs> TransportClosed;
+
+        public bool IsClosed { get; private set; }
+
+        public async Task SendAsync(string message)
+        {
+            var messageBuffer = Encoding.UTF8.GetBytes(message);
+            var messagesCount = (int)Math.Ceiling((double)messageBuffer.Length / DefaultBufferSize);
```
pavelfeldman, line 47: Style: You would typically implement this as following:
```cs
for (var offset = 0; offset < messageBuffer.length; offset += DefaultBufferSize) {
    var count = Math.Min(messageBuffer.length - offset, messageBuffer.length);
    new ArraySegment<byte>(messageBuffer, offset, count);
}
```

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,194 @@
...
+                TransportClosed?.Invoke(this, new TransportClosedEventArgs { CloseReason = closeReason });
+                _readerCancellationSource.Cancel();
+            }
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
```
pavelfeldman, line 90: Why is this here?

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,194 @@
...
+                            result = await _webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), _readerCancellationSource.Token).ConfigureAwait(false);
+
+                            if (result.MessageType == WebSocketMessageType.Close)
+                            {
+                                await
+                                    _webSocket.CloseAsync(WebSocketCloseStatus.NormalClosure, string.Empty, _readerCancellationSource.Token).ConfigureAwait(false);
+                            }
+                            else
+                            {
+                                var str = Encoding.UTF8.GetString(buffer, 0, result.Count);
```
pavelfeldman, line 143: Bug: You can't decode partial strings.

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -71,6 +75,8 @@ public void Close(string closeReason)
             if (!IsClosed)
             {
                 IsClosed = true;
+                _process?.Kill();
```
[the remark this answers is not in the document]
Archish27: Ah yes, I moved the process from Connection class to StdIOTransport will add tr/catch here.

src/Playwright/Core/BrowserType.cs
```diff
@@ -117,5 +118,24 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
                 recordHarOmitContent: options.RecordHarOmitContent,
                 recordVideo: Browser.GetVideoArgs(options.RecordVideoDir, options.RecordVideoSize)).ConfigureAwait(false)).Object;
         }
+
+        public async Task<IBrowser> ConnectAsync(string wsEndpoint, BrowserTypeConnectOptions options = null)
+        {
+            options ??= new BrowserTypeConnectOptions();
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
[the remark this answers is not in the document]
Archish27: I can remove it but ensuring that WebSocket objects & process objects get cleaned up. Ideally, we should keep it. thoughts? or any suggestion on refactoring?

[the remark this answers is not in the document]
Archish27: Also, websocket contains dispose method that needs to be disposed of. This is after removing `IDisposable`
image

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Freeing the WebSocket transport object once the job is done. Hence, required.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: How would I identify `endOfMessage` here? 
`Count <= 0` ?

src/Playwright/Core/BrowserType.cs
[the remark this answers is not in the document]
avodovnik: I think using `IDisposable` is the lesser of the two evils, because it gives us more good than bad. Yes, it's annoying having to disable the one warning, but, also, it's the recommended practice: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000#when-to-suppress-warnings

src/Playwright/Transport/WebSocketTransport.cs
avodovnik: It's a [common implementation](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-dispose-method) of `Dispose` in .NET. Basically, because we're calling `Dispose(true)` one line above, we call the `Suppress` to ensure we don't unnecessarily call the finalizer twice. The *actual* clean-up already happens in the first call, then we just clean the rest.

Archish27: So, suppress is required right? to clean the rest

src/Playwright/Transport/StdIOTransport.cs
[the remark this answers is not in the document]
Archish27: fixed

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -31,27 +35,59 @@ namespace Microsoft.Playwright.Tests
...
-            await browserServer.CloseAsync();
-            */
+        [PlaywrightTest("browsertype-connect.spec.ts", "should be able to reconnect to a browser")]
+        [Test, Timeout(TestConstants.DefaultTestTimeout)]
+        public async Task ShouldBeAbleToReconnectToBrowserAsync()
+        {
+            var browser = await BrowserType.ConnectAsync(_browserServer.WSEndpoint);
+            var context = await browser.NewContextAsync();
+            var page = await context.NewPageAsync();
+            await page.GotoAsync("https://google.com");
```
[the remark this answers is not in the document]
Archish27: fixed

src/Playwright/Program.cs
```diff
@@ -37,6 +37,8 @@ public static int Main(string[] args)
             return p.Run(args);
         }
 
+        public static string GetExecutablePath() => Paths.GetExecutablePath();
```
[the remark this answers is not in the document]
Archish27: How am I suppose to get the path in a test should I implement test util? 
`playwright.ps1` is not working for me as I am on mac

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -71,6 +75,8 @@ public void Close(string closeReason)
...
+                _process?.Dispose();
```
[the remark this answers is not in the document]
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
avodovnik: To clarify, `Suppress` doesn't clean, it prevents cleaning from happening *again* because the relevant (managed) finalization will happen in the `Dispose(true)` call. That's my interpretation at least.

pavelfeldman: But we are not doing anything in this dispose, why do we even need it?

src/Playwright/Program.cs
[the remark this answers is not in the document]
pavelfeldman: There is a `playwright.sh` next to it for your mac

src/Playwright/Transport/WebSocketTransport.cs
pavelfeldman: There was no endOfMessage in your snippet, I'm just proposing a shorter and a more standard alternative. For loop will end when the message is sent.

src/Playwright/Program.cs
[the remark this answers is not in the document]
avodovnik: I wonder if you can try launching `dotnet Microsoft.Playwright.dll launch-server ...` as the process straight from the test. The DLL is compiled, and it'll be right there, and we can rely on `dotnet` existing because, well, we're a .NET app :-).

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,194 @@
...
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+
+            if (_readerCancellationSource != null)
+            {
+                _readerCancellationSource.Cancel();
+                _readerCancellationSource.Dispose();
+            }
```
avodovnik:
```suggestion
                _readerCancellationSource?.Cancel();
                _readerCancellationSource?.Dispose();
```

src/Playwright/Transport/WebSocketTransport.cs
avodovnik: He is: https://github.com/microsoft/playwright-dotnet/pull/1602/files/d10acf1b24cc4dc4fbdd8f94072b1597d99e64b4#diff-b29da783aeafa3800dd5b5b9d0c05f44cd36d1dcfc662527c209f55593aa9693R92

He's cancelling & disposing the cancellation source.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: What needs to be done? here should I add try/catch?

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Program.cs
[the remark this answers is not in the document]
Archish27: I followed what other languages are doing -> https://github.com/microsoft/playwright-java/blob/a1ef49cd0389c6a32725a74ad01298069552ae0f/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java#L68

& also in actual code, we are spawning from `.playwright/{os}` path. So ideally in test, we should be testing with the same path this ensures we are testing the correct functionality.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: I tried breaking few cases. I will try to refactor it later.

src/Playwright/Transport/WebSocketTransport.cs
pavelfeldman: Ah, this now huge inline comment confused me. Thanks!

src/Playwright/Transport/WebSocketTransport.cs
pavelfeldman: Reading https://en.wikipedia.org/wiki/UTF-8 would be a good start.

Archish27: We do similar for StdIOTransport -> https://github.com/microsoft/playwright-dotnet/blob/87b0ca33ca9849678e256ef3741fa5959f7f0cc8/src/Playwright/Transport/StdIOTransport.cs#L184

src/Playwright/Playwright.cs
```diff
@@ -38,7 +39,9 @@ public static class Playwright
         /// <returns>A <see cref="Task"/> that completes when the playwright driver is ready to be used.</returns>
         public static async Task<IPlaywright> CreateAsync()
         {
-            var connection = new Connection();
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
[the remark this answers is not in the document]
Archish27: Resolving as it same as earlier

src/Playwright/Program.cs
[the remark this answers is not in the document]
avodovnik: Yes, if you run `dotnet Microsoft.Playwright.dll` it will spawn the OS specific version, using all the internal stuff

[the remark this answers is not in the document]
Archish27: Ideally, we shouldn’t be doing it in this way. Since we are not testing the actual functionality. In the source we are spawning on some different path & in the test we are spawning a different path. To test functionality should be the same.

src/Playwright/Transport/WebSocketTransport.cs
pavelfeldman: The message is not partial there. As I said above, you can't decode partial messages.

src/Playwright/Program.cs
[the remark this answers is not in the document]
pavelfeldman: It is going to be the same, because we are going to be running the same code to compute it. And we don't need to expose it in the api that way.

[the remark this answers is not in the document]
Archish27: updated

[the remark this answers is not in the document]
Archish27: resolving this...

src/Playwright/Transport/WebSocketTransport.cs
Archish27: One approach that I can think of accumulating all data in a `byte` list i.e it will have data from remote & then encoding it with string.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,207 @@
...
+                                }
+                                else
+                                {
+#if NETSTANDARD
+#pragma warning disable CA1835 // We can't use ReadOnlyMemory on netstandard
+                                    await memoryStream.WriteAsync(buffer, 0, result.Count, _readerCancellationSource.Token).ConfigureAwait(false);
+#pragma warning restore CA1835
+#else
+                                await memoryStream.WriteAsync(new(buffer, 0, result.Count), _readerCancellationSource.Token).ConfigureAwait(false);
+#endif
```
avodovnik, line 149:
```suggestion
#if NETSTANDARD
                                    await memoryStream.WriteAsync(buffer.Array, 0, result.Count, _readerCancellationSource.Token).ConfigureAwait(false);
#else
                                await memoryStream.WriteAsync(new(buffer.Array, 0, result.Count), _readerCancellationSource.Token).ConfigureAwait(false);
#endif
```

No need for pragmas when you've got conditional compilation + using the new `.Array` byref call, assuming the comment above gets merged too.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,207 @@
...
+        {
+            try
+            {
+                if (_webSocket.State == WebSocketState.Closed)
+                {
+                    Close("Closed");
+                }
+                else
+                {
+                    byte[] buffer = new byte[DefaultBufferSize];
```
avodovnik, line 125:
```suggestion
                    var buffer = WebSocket.CreateClientBuffer(DefaultBufferSize, DefaultBufferSize);
```

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,207 @@
...
+
+                    while (!token.IsCancellationRequested && _webSocket.State == WebSocketState.Open)
+                    {
+                        WebSocketReceiveResult result;
+                        using (MemoryStream memoryStream = new())
+                        {
+                            do
+                            {
+                                result = await _webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), _readerCancellationSource.Token).ConfigureAwait(false);
```
avodovnik, line 134:
```suggestion
                                result = await _webSocket.ReceiveAsync(buffer, _readerCancellationSource.Token).ConfigureAwait(false);
```

avodovnik: Did you intentionally use `_readerCancellationSource.Token` instead of the `token` you're already provided in the method?

avodovnik (approved):
I think the new change addresses Pavel's comments, as it now correctly reads the stream until it gets an `EndOfMessage`.

avodovnik (commented):
Needs formatting (`dotnet format`) and review of the comments made.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Boom💥  This is nice I was reading about this today morning :) Nice catch 🚀

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Aahh naa my bad this needs to be refactored 👍

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Thanks @pavelfeldman for your eyes on this :) I have fixed this.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,205 @@
...
+            if (_options.Headers is IDictionary<string, string> dictionary && dictionary.Keys.Any(f => f != null))
+            {
+                foreach (var kv in dictionary)
+                {
+                    if (kv.Value != null)
+                    {
+                        _webSocket.Options.SetRequestHeader(kv.Key, kv.Value);
+                    }
+                }
+            }
```
avodovnik, line 196: Can we simplify this to:
```suggestion
            foreach (var item in _options?.Headers ?? Array.Empty<KeyValuePair<string, string>>())
            {
                _webSocket.Options.SetRequestHeader(item.Key, item.Value);
            }
```

avodovnik: Or, alternatively, if you really want to filter out the ones who have `null` in the key (is that even possible?):

```suggestion
                foreach (var item in
                (_options?.Headers ?? Array.Empty<KeyValuePair<string, string>>())
                    .Where(x => !string.IsNullOrEmpty(x.Key)))
            {
                _webSocket.Options.SetRequestHeader(item.Key, item.Value);
            }
```

avodovnik:
Looking at the test failures, it's reading like a `null reference` in https://github.com/microsoft/playwright-dotnet/pull/1602/files#diff-4fde6de6200a1a761d3c6fe0ad8118dbb590117fb667507beb26f468fb110da8R80. 

I'm wondering if that's related to parallel execution. It'd be a good idea to identify the root cause here... But, we can guard against the process dying, for example, by doing

```csharp
_browserServer?.Process?.Kill();`
```

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -38,11 +39,14 @@ internal class StdIOTransport : IConnectionTransport, IDisposable
...
-        internal StdIOTransport(Process process)
+        internal StdIOTransport()
         {
-            _process = process;
-            process.ErrorDataReceived += (_, e) => LogReceived?.Invoke(this, new(e.Data));
-            process.BeginErrorReadLine();
+            _process = GetProcess();
+            _process.StartInfo.Arguments = "run-driver";
+            _process.Start();
+            _process.Exited += (_, _) => Close("Process exited");
```
avodovnik, line 47: Hm, you have this, but then you also have the same in `[TearDown]`. That might also be the reason for failures, i.e., if that process dies (legitimately), this gets triggered, and then the tear down happens & re-triggers it. 

You can also just have a `Close` method on the `BrowserServer` and it checks it hasn't been called yet, if you wanted to.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Wow this looks clean 👍

src/Playwright/Transport/StdIOTransport.cs
Archish27: I don't this will affect since we are spawning two different servers one is for connect tests & which is separate from the process object I don't think this can be an issue. Since the above close will close the server for execution & the test will close the process of the test playwright server.

Archish27: Have updated tests it was due to the playwright process not getting started. Have fixed tests

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,199 @@
...
+                        WebSocketReceiveResult result;
+                        using (MemoryStream memoryStream = new())
+                        {
+                            do
+                            {
+                                result = await _webSocket.ReceiveAsync(buffer, token).ConfigureAwait(false);
+
+                                if (result.MessageType == WebSocketMessageType.Close)
+                                {
+                                    await
```
pavelfeldman, line 138: It is weird that you initiate socket closure upon receiving the event that it was closed. In either case, CloseAsync seems to be a wrong thing to call: https://mcguirev10.com/2019/08/17/how-to-close-websocket-correctly.html. I would just call `HandleSocketClosed()` and break here. HandleSocketClosed should be called from both Close() and here.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,199 @@
...
+                                        _webSocket.CloseAsync(WebSocketCloseStatus.NormalClosure, string.Empty, _readerCancellationSource.Token).ConfigureAwait(false);
+                                }
+                                else
+                                {
+#if NETSTANDARD
+                                    await memoryStream.WriteAsync(buffer.Array, 0, result.Count, _readerCancellationSource.Token).ConfigureAwait(false);
```
pavelfeldman, line 144: I don't think memcpy should be async.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,199 @@
...
+        private async Task ConnectAsync()
+        {
+            await _webSocket.ConnectAsync(new Uri(_wsEndpoint), _readerCancellationSource.Token).ConfigureAwait(false);
+            ScheduleTransportTask(ReceiveAsync, _readerCancellationSource.Token);
+        }
+
+        private void ScheduleTransportTask(Func<CancellationToken, Task> func, CancellationToken cancellationToken)
+            => Task.Factory.StartNew(() => func(cancellationToken), cancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Current);
+
+        private async Task ReceiveAsync(CancellationToken token)
```
pavelfeldman, line 115: Since this is dispatching multiple incoming messages, let's call it `DispatchIncomingMessagesAsync`

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,199 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Playwright.Transport
+{
+    internal class WebSocketTransport : IConnectionTransport, IDisposable
+    {
+        private const int DefaultBufferSize = 1024;  // Byte buffer size
```
pavelfeldman, line 17: Bump it to 16K?

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        private void SetOptions()
+        {
+            _webSocket.Options.SetRequestHeader("User-Agent", GenerateUserAgent());
+            foreach (var item in _options?.Headers ?? Array.Empty<KeyValuePair<string, string>>())
+            {
+                _webSocket.Options.SetRequestHeader(item.Key, item.Value);
+            }
+
+            var timeout = _options.Timeout;
```
pavelfeldman, line 192: `connect({timeout})` implies that there is a budget of `timeout` that we have to establish the connection before returning from connect, has nothing to do with KeepAlive.

Archish27:
@pavelfeldman @avodovnik Can please review all at once? Too much back & forth happening.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: This will be async since we are calling from the async method. `Write` synchronously blocks hence have to use async here.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: yup, that also works, I think no need for looping messages & sending in chunks we can send directly. Thus increasing the buffer size 👍

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,199 @@
...
+                        var offset = DefaultBufferSize * i;
+                        var count = DefaultBufferSize;
+                        var lastMessage = (i + 1) == messagesCount;
+
+                        if ((count * (i + 1)) > messageBuffer.Length)
+                        {
+                            count = messageBuffer.Length - offset;
+                        }
+
+                        await _webSocket.SendAsync(new ArraySegment<byte>(messageBuffer, offset, count), WebSocketMessageType.Text, lastMessage, _readerCancellationSource.Token).ConfigureAwait(false);
```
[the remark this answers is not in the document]
Archish27: Nice 👍

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Make sense have configured timeout for `CancellationTokenSource`

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        /// <inheritdoc/>
+        public void Close(string closeReason)
+        {
+            if (!IsClosed)
+            {
+                IsClosed = true;
+                _ = _webSocket.CloseAsync(WebSocketCloseStatus.NormalClosure, closeReason, _readerCancellationSource.Token).ConfigureAwait(false);
+                TransportClosed?.Invoke(this, new TransportClosedEventArgs { CloseReason = closeReason });
```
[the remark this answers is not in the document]
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: have fixed this

src/Playwright/Transport/WebSocketTransport.cs
pavelfeldman: What is wrong with blocking? We are presumably copying a chunk of memory. Maybe the MemoryStream is not the best data structure for it?

pavelfeldman: It looks like MemoryStream is ok, just don't use the async versions of these methods.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        private void Close(Exception ex)
+        {
+            System.Diagnostics.Debug.WriteLine(ex);
+            Close(ex.ToString());
+        }
+
+        private void HandleSocketClose(string closeReason)
+        {
+            _webSocket.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, closeReason, _readerCancellationSource.Token).ConfigureAwait(false);
```
pavelfeldman, line 94: It is still weird that you react to the "socket closed" event via closing the socket. What should happen here instead is handling the socket closure: proper tear down and emitting the transport closed event. And it should happen when the socket closes, either by remote or by local.

pavelfeldman:
> Too much back & forth happening

I agree. Addressing review comments would help minimizing the churn. So far I need to re-read and repeat the same review comments over and over. Typically we would close such PRs for hurting the core team productivity, but each time I think "if this all is addressed, we get a good chance of landing it". So I'll be throttling my reviews a little for this to not consume all of my reviewing capacity.

Unfortunately, there is also entire layer of `isRemote` operation that this PR does not address. And given the back and forth that you are referring to, I would not be adding it into this change.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: updated, it will invoke TransportClosed event

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Updated to synchronous call

Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -70,9 +74,16 @@ public void Close(string closeReason)
         {
             if (!IsClosed)
             {
-                IsClosed = true;
-                TransportClosed?.Invoke(this, new() { CloseReason = closeReason });
-                _readerCancellationSource.Cancel();
+                try
+                {
+                    IsClosed = true;
+                    TransportClosed?.Invoke(this, new() { CloseReason = closeReason });
```
pavelfeldman, line 80: We should not silent-catch code, especially code that invokes callbacks.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,180 @@
...
+        private readonly int _timeout;
+
+        internal WebSocketTransport(
+            string wsEndpoint = default,
+            BrowserTypeConnectOptions options = default)
+        {
+            _webSocket = new ClientWebSocket();
+            _wsEndpoint = wsEndpoint;
+            _options = options;
+            _slowMo = _options?.SlowMo != null ? (int)_options?.SlowMo : 0;
```
pavelfeldman, line 31: `_slowMo = _options?.SlowMo ?? 0;`

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,180 @@
...
+            _timeout = _options?.Timeout != null ? (int)_options?.Timeout : 30000;
```
pavelfeldman, line 32: _timeout = _options?.Timeout ?? 30000;

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,180 @@
...
+            _readerCancellationSource = new CancellationTokenSource(_timeout);
```
pavelfeldman, line 33: Bug: if I read it correctly, after 30 seconds of operation this web socket will die. That's not our intent with timeout.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        /// <inheritdoc/>
+        public void Close(string closeReason)
+        {
+            if (!IsClosed)
+            {
+                IsClosed = true;
+                HandleSocketClose(closeReason);
```
pavelfeldman, line 73: When user closes transport, you should close the web socket. If I understand the current workflow, this will automatically get called.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,180 @@
...
+        private void Close(Exception ex)
+        {
+            System.Diagnostics.Debug.WriteLine(ex);
+            Close(ex.ToString());
+        }
+
+        private void HandleSocketClose(string closeReason)
+        {
+            TransportClosed?.Invoke(this, new TransportClosedEventArgs { CloseReason = closeReason });
+            _readerCancellationSource?.Cancel();
```
pavelfeldman, line 93: We are handling the socket close, so all the socket operations will fail, I don't think you need this Cancel() anymore?

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,180 @@
...
+        private void ScheduleTransportTask(Func<CancellationToken, Task> func, CancellationToken cancellationToken)
+            => Task.Factory.StartNew(() => func(cancellationToken), cancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Current);
+
+        private async Task DispatchIncomingMessagesAsync(CancellationToken token)
+        {
+            try
+            {
+                if (_webSocket.State == WebSocketState.Closed)
+                {
+                    Close("Closed");
```
pavelfeldman, line 111: This is probably the third time I'm making this comment - we should not react to the fact that socket is closed with an attempt to Close the socket. This is where you call HandleSocketClosed()

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,180 @@
...
+                        WebSocketReceiveResult result;
+                        using (MemoryStream memoryStream = new())
+                        {
+                            do
+                            {
+                                result = await _webSocket.ReceiveAsync(buffer, token).ConfigureAwait(false);
+
+                                if (result.MessageType == WebSocketMessageType.Close)
+                                {
+                                    Close("Closed");
```
pavelfeldman, line 128: Same here, this should be HandleSocketClosed()

src/Playwright/Transport/StdIOTransport.cs
Archish27: Since it includes `process.kill` so had to enclose it in try/catch.
This is the current master code which does similar thing -> https://github.com/microsoft/playwright-dotnet/blob/f689cf4ec76b652c93946b183a46ab02222fd7d5/src/Playwright/Transport/Connection.cs#L449

Will only keep process object in try/catch others will remove out of it.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: yup not required

Archish27: It is required, if there are any task queued that needs to be cancelled since socket is closed. It is good to have. We do similar thing for `StdIOTransport` - https://github.com/microsoft/playwright-dotnet/blob/41740926d374220d1becd95756cc62099b425d52/src/Playwright/Transport/StdIOTransport.cs#L75

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/StdIOTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,183 @@
...
+
+namespace Microsoft.Playwright.Transport
+{
+    internal class WebSocketTransport : IConnectionTransport, IDisposable
+    {
+        private const int DefaultBufferSize = 16000;  // Byte buffer size
+        private readonly ClientWebSocket _webSocket;
+        private readonly string _wsEndpoint;
+        private readonly CancellationTokenSource _readerCancellationSource;
+        private readonly CancellationTokenSource _connectCancellationSource;
```
pavelfeldman, line 20: What's up with all these GH warnings?

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -113,5 +151,64 @@ public void ShouldTerminateNetworkWaiters()
...
         }
+
+
+        private class BrowserServer
+        {
+            public Process Process { get; set; }
+            public string WSEndpoint { get; set; }
+        }
+
+        private static string GetExecutablePath()
```
pavelfeldman, line 162: Why copy-paste this code and not run Program?

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,183 @@
...
+        {
+            _webSocket = new ClientWebSocket();
+            _wsEndpoint = wsEndpoint;
+            _options = options;
+            _slowMo = _options?.SlowMo ?? 0;
+            _timeout = _options?.Timeout ?? 30000;
+            _readerCancellationSource = new CancellationTokenSource();
+            _connectCancellationSource = new CancellationTokenSource((int)_timeout);
+            SetRequestHeaders();
+            _ = ConnectAsync();
```
pavelfeldman, line 37: Move .ConnectAsync to the call site instead. You are in async there, so you can await `transport.ConnectAsync()` there and handle timeout in the right stack.

pavelfeldman (commented):
It is looking pretty good. Let's fix the test, a couple of nits and you are good to land. Note that this won't enable remote operation until `isRemote` is also handled, in a separate change.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Ahh yes the object needs to be disposed.

src/Playwright.Tests/BrowserTypeConnectTests.cs
Archish27: Since, serverEnpoint is always changing we need to return that server endpoint & use in tests. 
If I start using Program I won't be able to get server endpoint in tests. Hence, had to do this.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Anyways, this reader token seems to be redundant. Removing all together.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: Make sense, but was in thought process when websocket transport object is created it should start connecting. Cool will follow you approach where connect needs to be explicitly called.

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,183 @@
...
+    internal class WebSocketTransport : IConnectionTransport, IDisposable
+    {
+        private const int DefaultBufferSize = 16000;  // Byte buffer size
+        private readonly ClientWebSocket _webSocket;
+        private readonly string _wsEndpoint;
+        private readonly CancellationTokenSource _readerCancellationSource;
+        private readonly CancellationTokenSource _connectCancellationSource;
+        private readonly BrowserTypeConnectOptions _options;
+        private readonly float _slowMo;
+        private readonly float _timeout;
```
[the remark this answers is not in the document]
Archish27: fiixed

src/Playwright/Transport/WebSocketTransport.cs
Archish27: updated

src/Playwright/Transport/WebSocketTransport.cs
Archish27: fixed

Archish27:
@pavelfeldman Also, can you take a look to this PR: [API Changes](https://github.com/microsoft/playwright/pull/7469) 
I think checks are failing since it not merged in master.

pavelfeldman:
In the last patch set `_readerCancellationSource` was removed. But that was the way you controlled the lifetime of the actual socket. Now I see `_connectCancellationSource` in `WebSocketTransport::Dispose` which does not sound right. Why did that change?

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -0,0 +1,174 @@
...
+            catch (Exception ex)
+            {
+                Close(ex);
+            }
+        }
+
+        /// <inheritdoc/>
+        public void Close(string closeReason)
+        {
+            if (!IsClosed)
```
pavelfeldman, line 63: This method no longer closes the web socket :/

pavelfeldman:
I'll land as is and finish this as a follow up.

Archish27:
@pavelfeldman @avodovnik Thanks for you review & valuable time. Happy to see getting this merged 🚀 😄

PR #1603, opened 2021-07-05
fix: do not reset driver path if assembly location is empty

avodovnik:


src/Playwright/Helpers/Paths.cs
```diff
@@ -37,7 +37,14 @@ internal static string GetExecutablePath()
             if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, "Microsoft.Playwright.dll")))
```
JoelEinbinder, line 37: The second half of this condition appears to be unused? We check assemblyDirectory.Exists unconditionally on line 40.

avodovnik: There's a (weird) condition that happens when the directory exists, but the file doesn't exist. We then _rewrite_ the directory from `Assembly.Location` which can be `null` for some cases (specifically, for single file applications). 

*BUT* for most cases, the that file will exist.

pavelfeldman:
Is there a bug for this?

avodovnik:
> Is there a bug for this?

Yes, referenced it now.

src/Playwright/Helpers/Paths.cs
```diff
@@ -37,7 +37,14 @@ internal static string GetExecutablePath()
...
             {
                 var assemblyLocation = typeof(Playwright).Assembly.Location;
-                assemblyDirectory = new FileInfo(assemblyLocation).Directory;
+                if (!string.IsNullOrEmpty(assemblyLocation))
+                {
+                    // For clarification, we only overwrite the directory if the assembly location is
+                    // not empty, like when running from a single file packaged application. In that
+                    // context, getting the dll is unsupported, but also not needed. The rest of the
+                    // functionalities will work, and we can leave it at that.
+                    assemblyDirectory = new FileInfo(assemblyLocation).Directory;
```
pavelfeldman, line 46: So `Assembly.Location` is an empty string...  Then you are looking at the `cwd`'s directory? But what is the point of having assemblyDirectory point relative to `cwd`?

avodovnik: In cases like LinqPad, the cwd will be different to the assembly location.

pavelfeldman: But will it point anywhere reasonable? Also, this is not the bug you refer to then? What are you fixing?

avodovnik: Yes.

So, the default/correct lookup is `AppContext.BaseDirectory`. But, that can point to the wrong place sometimes (like when loaded in LinqPad, or interactive), so we double-check if our DLL is deployed next to it. If it's not, we revert to the assembly location. Now, for special cases, like the one in the last comment (single file apps) of the linked issue, that is *empty*, so we were setting the location to an empty string. 

This particular edge case is that the `.playwright` folder is there, and the path returned from `BaseDirectory` is *NOT EMPTY*, however, it does not have `Microsoft.Playwright.dll` because it's embedded in a single file. So our check incorrectly went down this route.

pavelfeldman (approved):

src/Playwright/Helpers/Paths.cs
```diff
@@ -33,31 +33,38 @@ internal static class Paths
...
         internal static string GetExecutablePath()
         {
-            DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
-            if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, "Microsoft.Playwright.dll")))
+            /*
+             * Scenarios this has to cover:
+             * 1) The simple one, where the .playwright folder is next to the BaseDirectory, which is almost always
+             * 2) Scenarios where we have to rely on the assembly location of Microsoft.Playwright.dll, but can't be
+             *      sure that's always correct, because it can be empty for some cases (hence 1. is first)
+             * 3) Scenarios where we get loaded from NuGet and **NOT** copied, which is rare, but implies we're loaded
```
pavelfeldman, line 41: Could you include the file tree snippet in the comment to make it clear why drivers are not next to .dll and we need to climb the filesystem tree? I.e. why (2) does not cover (3). Same would be interesting for (1), i.e. when (2) does not cover (1).

PR #1604, opened 2021-07-05
fix: re-introduce ability to output diagnostics log

### LINKED ISSUE #1592: [Bug] driver stderr does not get forwarded (DEBUG=pw:api does not work)
I used to do...

`var playwright = await Playwright.CreateAsync(playwrightLogger, debug: "pw:api");`

...however now that doesn't work anymore, as far as I can tell.

I've tried setting the `DEBUG` env var to `pw:api` both for the user and the machine, but that didn't seem to help. What am I doing wrong?

avodovnik:
Closes #1592

### DIFF
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -55,14 +55,9 @@ namespace Microsoft.Playwright.Transport
         private int _lastId;
         private string _reason = string.Empty;
 
-        public Connection()
+        public Connection(ILoggerFactory loggerFactory)
         {
-            _loggerFactory = LoggerFactory.Create(builder =>
-            {
-                builder.SetMinimumLevel(LogLevel.Debug);
-                builder.AddFilter((f, _) => f == "PlaywrightSharp.Playwright");
-            });
-
+            _loggerFactory = loggerFactory;
             _logger = _loggerFactory.CreateLogger<Connection>();
             var debugLogger = _loggerFactory?.CreateLogger<PlaywrightImpl>();
 
@@ -74,7 +69,7 @@ namespace Microsoft.Playwright.Transport
             _playwrightServerProcess.Exited += (_, _) => Close("Process exited");
             _transport = new StdIOTransport(_playwrightServerProcess, _loggerFactory);
             _transport.MessageReceived += Transport_MessageReceived;
-            _transport.LogReceived += (_, e) => debugLogger?.LogInformation(e.Message);
+            _transport.LogReceived += (_, e) => debugLogger?.LogDebug(e.Message);
             _transport.TransportClosed += (_, e) => Close(e.CloseReason);
         }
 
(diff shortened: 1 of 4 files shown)


src/Playwright.Tests/BaseTests/SetUpFixture.cs
```diff
@@ -0,0 +1,46 @@
...
+    class SetUpFixture
+    {
+        [OneTimeSetUp]
+        public void ConfigurePlaywrightTests()
+        {
+            // Use Default Logging
+            PlaywrightTest.LoggerFactory = Extensions.Logging.LoggerFactory.Create(builder => builder
+                   .ClearProviders()
+                   // To debug Playwright for .NET, we can set this to Trace
+                   .SetMinimumLevel(LogLevel.Debug)
```
avodovnik, line 42: I'm leaving this here, because it'll come in handy in the future. But this is technically not needed.

pavelfeldman:
Maybe we could fix https://github.com/microsoft/playwright-dotnet/issues/1592#issuecomment-868788168 miminally? As in other ports...

src/Playwright/Transport/Connection.cs
```diff
@@ -284,7 +279,7 @@ private void Transport_MessageReceived(object sender, MessageReceivedEventArgs e
                 return;
             }
 
-            _logger?.LogInformation($"pw:channel:event {e.Message}");
+            _logger?.LogTrace($"pw:channel:event {e.Message}");
```
mxschmitt, line 282: in Python we have something like that:

```
if process.env.DEBUGP === "1":
  console.log(message)
```
for messages which we receive and send. I would recommend doing it too in .NET since this should never be interesting for the user.

avodovnik:
> Maybe we could fix [#1592 (comment)](https://github.com/microsoft/playwright-dotnet/issues/1592#issuecomment-868788168) minimally? As in other ports...

Depends on what you consider "minimally". I think my latest change is the least amount of code still reasonable. I removed the NUnit specific logger - so we only use one, and it works for both `Console` and inside Test Explorer.:

image

image

That said, I don't want to remove the ability to pass a custom `ILoggerFactory` because that's the "norm" of logging things in .NET and if a user has their own logging factory, I want to make sure they can pass one. It makes sense to use this. 

If we want less changes, I can skip invoking the event and write directly to `System.Console.Error`. But, like I said, I think the amount of code is minimal but has the maximum possible impact/usability.

src/Playwright/Transport/PlaywrightLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Playwright.Transport
+{
+    internal class PlaywrightLogger : ILogger
+    {
+        private static readonly Dictionary<string, ConsoleColor> _debugLoggerColorMap = new()
+        {
+            { "pw:api", ConsoleColor.Cyan },
```
mxschmitt, line 13: can we drop that color map? I think not needed for .NET. and this would simplify it and reduce the maintenance overhead.

PR #1606, opened 2021-07-08
fix(stderr): remove ILogger and ILoggerFactory and re-route stderr

### LINKED ISSUE #1592: [Bug] driver stderr does not get forwarded (DEBUG=pw:api does not work)

avodovnik:
Closes #1592

### DIFF
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -33,7 +33,6 @@ using System.Linq;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Logging;
 using Microsoft.Playwright.Core;
 using Microsoft.Playwright.Helpers;
 using Microsoft.Playwright.Transport.Channels;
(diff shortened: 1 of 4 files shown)


src/Playwright/Transport/Connection.cs
```diff
@@ -461,5 +453,14 @@ private void Dispose(bool disposing)
             {
             }
         }
+
+        [Conditional("DEBUG")]
+        private void TraceMessage(string message)
+        {
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DEBUGPWD")))
```
dgozman, line 460: Is `DEBUGPWD` some kind of standard debug env in dotnet? If not, why don't we use the same `DEBUG=pw*` as we do everywhere?

dgozman (commented):
Where is the "re-route stderr" part? I guess I am missing some context...

avodovnik:
> Where is the "re-route stderr" part? I guess I am missing some context...

https://github.com/microsoft/playwright-dotnet/pull/1606/files/b35e427918022dfb5b11f81cabd47bc70f869548#diff-a387ff4bbdf5b27cc00330eae6e3c41df555197a80d35f8cc21dec46152e8d0cR65

src/Playwright/Transport/Connection.cs
avodovnik: No, it's purely internal, and it's meant purely for developing the dotnet port. It's the same approach as in [Python](https://github.com/microsoft/playwright-python/blob/21b7d1ca59dce3b22c17d348bce54b7ae0e0780d/playwright/_impl/_transport.py#L72).

The PR encompasses two major changes.

1) The `stderr` redirect now happens in a single line (see link above)
2) The internal _tracing_ that's potentially useful for the debugging of the port itself, happens with the `TraceMessage` method and is not compiled into the production builds.

avodovnik: To add some context: we *used* to add a bunch of our (=dotnet specific) internal logging to the same output produces by the driver, when using `DEBUG=pw*`. This no longer happens. So, whatever you set in your `DEBUG`, you get the output into `stderr`, displayed either in NUnit's Test Window, or in the Console. 

For our internal stuff, because I still want it there, I moved it to the right place, which is to use the `System.Diagnostics` mechanism.

avodovnik:
Upstream docs: https://github.com/microsoft/playwright/pull/7468/

dgozman (approved):

PR #1607, opened 2021-07-08
chore(roll): rolling driver to 1.13.0-next-1625812834000

avodovnik:


mxschmitt (commented):
note that baseURL needs also implementation on the client side. Upstream its used in the urlMatches function which gets reflect in:
- waitForRequest
- waitForResponse
- waitForURL
- route

src/Playwright.Tests/BrowserContextBaseUrlTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        [PlaywrightTest("browsercontext-base-url.spec.ts", "should construct the URLs correctly when a baseURL without a trailing slash in browser.newPage is passed to page.goto")]
+        [Test, Timeout(TestConstants.DefaultTestTimeout)]
+        public async Task ShouldConstructTheURLsCorrectlyWihoutTrailingSlash()
+        {
+            var serverUri = new Uri(Server.Prefix);
+            var page = await Browser.NewPageAsync(new() { BaseURL = new Uri(serverUri, "/url-construction").ToString() });
+
+            Assert.AreEqual(new Uri(serverUri, "/mypage.html"), (await page.GotoAsync("mypage.html")).Url);
```
mxschmitt, line 64: Does page.Url return a string or a URL object? If it's a URL object, then we need to fix it in the docs. Otherwise I would change the assertion as suggested (and below):

```suggestion
            Assert.AreEqual(Server.Prefix + "/mypage.html", (await page.GotoAsync("mypage.html")).Url);
```

src/Playwright.Tests/PageNetworkResponseTests.cs
```diff
@@ -174,5 +174,27 @@ public async Task ShouldWaitUntilResponseCompletes()
             serverResponseCompletion.SetResult(true);
             Assert.AreEqual("hello world!", await responseText);
         }
+
+        [PlaywrightTest("har.spec.ts", "should return security details directly from response")]
+        [Test, SkipBrowserAndPlatform(skipLinux: true, skipFirefox: true)]
```
mxschmitt, line 179: upstream its webkit+linux (and below). Whats the reason that you are skipping it for FF?

src/Playwright/Core/Browser.cs
```diff
@@ -140,12 +141,14 @@ public async Task<IPage> NewPageAsync(BrowserNewPageOptions options = default)
...
                 StorageState = options.StorageState,
                 StorageStatePath = options.StorageStatePath,
+                BaseURL = options.BaseURL,
             };
 
             var context = (BrowserContext)await NewContextAsync(contextOptions).ConfigureAwait(false);
 
             var page = (Page)await context.NewPageAsync().ConfigureAwait(false);
             page.OwnedContext = context;
+            context.BaseUrl = options.BaseURL;
```
mxschmitt, line 151: don't you store the context options already on the context when creating a new context? Thats how its done upstream and in Python.

src/Playwright/Core/Frame.cs
```diff
@@ -535,6 +538,9 @@ private Waiter SetupNavigationWaiter(string apiName, float? timeout)
 
         private bool UrlMatches(string url, string matchUrl, Regex regex, Func<string, bool> match)
         {
+            var baseUrl = (Page.Context as BrowserContext)?.BaseUrl;
```
mxschmitt, line 541: we should only combine if its not starting with * and a string. The second condition is for match always true but I miss a * check there.

src/Playwright/Core/Page.cs
```diff
@@ -337,7 +337,7 @@ public Task<IResponse> RunAndWaitForNavigationAsync(Func<Task> action, PageRunAn
             });
 
         public Task<IRequest> WaitForRequestAsync(string urlOrPredicate, PageWaitForRequestOptions options = default)
-            => InnerWaitForEventAsync(PageEvent.Request, null, e => e.Url.UrlMatches(urlOrPredicate), options?.Timeout);
+            => InnerWaitForEventAsync(PageEvent.Request, null, e => e.Url.UrlMatches(urlOrPredicate.CombineUrlWithBase(Context.BaseUrl)), options?.Timeout);
```
mxschmitt, line 340: Instead of merging it everywhere I would pass the baseURL inside URLMatches like we do in Python and if baseURL is given + does not start with * I would join.

(and below)

src/Playwright.Tests/PageNetworkResponseTests.cs
avodovnik: Typo. Good spot.

src/Playwright/Core/Browser.cs
avodovnik: No, we don't store the `options` classes AFAICS.

src/Playwright/Core/Frame.cs
avodovnik: Is there a test case that I missed for this? Also, do you mean `baseUrl` or the `url` start with a `*`? If the latter, then we *do* check:

```
 if (string.IsNullOrEmpty(baseUrl) || !Uri.IsWellFormedUriString(url, UriKind.RelativeOrAbsolute))
```

Because that won't be a well-formed URI string and we'll just return the unmerged URL.

src/Playwright/Core/Page.cs
avodovnik: It's just a preference of using an extension method. I don't think it makes a difference.

src/Playwright.Tests/BrowserContextBaseUrlTests.cs
mxschmitt: Would be great if you could align this with upstream, then its easier to follow instead of doing URI magic.

src/Playwright/Core/Frame.cs
mxschmitt: yes this is a test-case which was not there, let me add one.

src/Playwright/API/Generated/IFrame.cs
```diff
@@ -636,8 +650,7 @@ public partial interface IFrame
         /// selector, the first will be used. See <a href="./selectors.md">working with selectors</a>
         /// for more details.
         /// </param>
-        /// <param name="options">Call options</param>
-        Task<bool> IsHiddenAsync(string selector, FrameIsHiddenOptions? options = default);
+        Task<bool> IsHiddenAsync(string selector);
```
pavelfeldman, line 653: This is an API breaking change, we should probably revert it upstream for simplicity.

src/Playwright/API/Generated/IFrame.cs
```diff
@@ -650,8 +663,7 @@ public partial interface IFrame
         /// selector, the first will be used. See <a href="./selectors.md">working with selectors</a>
         /// for more details.
         /// </param>
-        /// <param name="options">Call options</param>
-        Task<bool> IsVisibleAsync(string selector, FrameIsVisibleOptions? options = default);
+        Task<bool> IsVisibleAsync(string selector);
```
pavelfeldman, line 666: ditto

src/Playwright/API/Generated/Options/BrowserNewContextOptions.cs
```diff
@@ -47,31 +47,32 @@ public BrowserNewContextOptions(BrowserNewContextOptions clone)
         {
             if (clone == null) return;
             AcceptDownloads = clone.AcceptDownloads;
-            IgnoreHTTPSErrors = clone.IgnoreHTTPSErrors;
```
pavelfeldman, line 50: It would be easier to review if we landed the sorting roll separately.

src/Playwright/Core/Browser.cs
pavelfeldman: This is a nit, but I'm with Max - we could store the entire thing for simplicity.

src/Playwright/Core/Page.cs
pavelfeldman, line 340: I it easy to miss one of the call sites. I'd suggest that you expose `Context.UrlMatches(url,  urlOrPredicate)` that would look at its private BaseUrl and do the right thing. If we didn't do that upstream, we should have.

pavelfeldman (commented):
This looks good, only a couple of comments. Could you split it into patches, for example:
1. Sorted generator (options)
2. BaseURL infrastructure
3. Resource timing tests
4. Everything else

src/Playwright/API/Generated/IFrame.cs
pavelfeldman: https://github.com/microsoft/playwright/pull/7600

src/Playwright/Core/Frame.cs
avodovnik: Added that test. And separate explicit check.

avodovnik:
@pavelfeldman https://github.com/microsoft/playwright-dotnet/pull/1616 is the sorting PR... Hopefully the rebase won't be too horrible after that merges.

src/Playwright/Core/BrowserType.cs
```diff
@@ -77,23 +77,21 @@ public async Task<IBrowserContext> LaunchPersistentContextAsync(string userDataD
             options ??= new BrowserTypeLaunchPersistentContextOptions();
             return (await _channel.LaunchPersistentContextAsync(
                 userDataDir,
-                args: options.Args,
```
pavelfeldman, line 80: Curious why this shuffled

pavelfeldman (approved):

src/Common/Version.props
```diff
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
     <AssemblyVersion>1.13.0</AssemblyVersion>
-    <PackageVersion>$(AssemblyVersion)-next-1</PackageVersion>
-    <DriverVersion>1.12.0-1623185968000</DriverVersion>
+    <PackageVersion>$(AssemblyVersion)-next-2</PackageVersion>
```
mxschmitt, line 4: does this mean we publish a new beta version?

mxschmitt (approved):

src/Common/Version.props
avodovnik: Yeah, it'd be ready for that. But that's a manually triggered process.

src/Playwright/Core/BrowserType.cs
avodovnik: IIRC this was me, applying a refactoring, re-ordering the parameters to be in the same order as in the interface.

PR #1618, opened 2021-07-15
chore(eng): improve rolling

avodovnik:
This PR introduces the submodule for upstream dependency. The intent here is to help us ensure that we have a "stable" state of the docs from which our API was generated from, etc.

Additionally, it introduces a build task helper that encapsulates a bunch of common commands (like downloading the driver, generating the API, etc.) which are common during the dev process.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,0 +1,69 @@
+[CmdletBinding(PositionalBinding=$false)]
+Param(
+  [switch][Alias('a')]$api,
+  [switch]$certs,
+  [switch][Alias('d')]$driver,
+  [switch][Alias('r')]$restore,
+  [switch][Alias('b')]$build,
+  [switch][Alias('e')]$all,
+  [switch][Alias('t')]$test,
+  [switch][Alias('h')]$help,
+  [switch]$prereqs,
+  [ValidateSet("Debug", "Release")][Alias('c')][string]$configuration = "Debug",
+  [ValidateSet("net5.0", "netcoreapp3.1")][Alias('f')][string]$framework = "net5.0",
+  [int]$workers = 6,
+  [Parameter(ValueFromRemainingArguments=$true)][String[]]$properties
+)
+
+function Get-Help() 
+{
+  Write-Host "Actions (defaults to -help):"
+  Write-Host "  -api                    Performs the API generation using the currently checked out submodule."
+  Write-Host "  -certs                  Installs the https certificates for localhost tests."
+  Write-Host "  -driver                 Downloads the drivers."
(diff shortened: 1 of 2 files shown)


mxschmitt:
I'm worried about the submodule, because in theory we already have a relation between the driver (commitinfo file) and the upstream repository. I would maybe add a bot like this to ensure the generated code is up2date which should fit for your needs:

```yml
  verify_generated_files:
    name: Verify Generated Files
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2
      - name: Setup .NET Core
        uses: actions/setup-dotnet@v1
        with:
          dotnet-version: 5.0.100
      - name: Install dotnet-format
        run: dotnet tool install --global --version 5.0.211103 dotnet-format
      - name: Download Drivers
        run: dotnet run -p ./src/tools/Playwright.Tooling/Playwright.Tooling.csproj -- download-drivers --basepath .
      - name: Read commit SHA
        id: read-upstream-commit-sha
        run: echo "::set-output name=hash::$(cat src/Playwright/.drivers/linux/package/commitinfo)"
      - uses: actions/checkout@v2
        with:
          repository: 'microsoft/playwright'
          ref: ${{ steps.read-upstream-commit-sha.outputs.hash }}
          path: 'playwright'
      - name: Generate
        run: |
          node playwright/utils/doclint/generateDotnetApi.js src/Playwright
          rm -rf playwright
      - name: Verify generated API is up to date
        run: git diff --exit-code
```

Then instead of maintaining an upstream revision number (git submodule) and the driver version you only need to take care about a driver version like its done in Java and Python and they are always in sync since the bot needs to be green.

avodovnik:
I've added the integrity check into #1620 but it uses the submodule.

build.ps1
```diff
@@ -0,0 +1,97 @@
...
+function Get-SubmoduleStatus() {
+  $smStatus = git submodule status
+  return $smStatus
+}
+
+function Invoke-InitializeSubmodule($enableReset = $true) {
+  # Check the status of the submodule, if not initialized, let's do that
+  if ((Get-SubmoduleStatus).StartsWith("-") -or ($enableReset -and $reset)) {
+    Write-Host "🔨 Initializing git submodule..." -NoNewline
+    git submodule update --init >$null 2>&1
```
aslushnikov, line 36: FWIW, replacing this line with `git clone https://github.com/microsoft/playwright && git checkout $UPSTREAM_SHA` would be equal! However, both are fine.

aslushnikov (commented):
lgtm

PR #1620, opened 2021-07-15
chore(eng): introduce API integrity checking

avodovnik:
Needs #1607

### DIFF
--- a/.github/workflows/integrity-checks.yml
+++ b/.github/workflows/integrity-checks.yml
@@ -1,0 +1,33 @@
+name: Integrity Checks
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
+  pull_request:
+    branches:
+      - main
+      - release-*
+
+jobs:
+ verify_generated_files:
+    name: Perform API Integrity Check
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 0
+      - name: Setup .NET Core
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 5.0.x
+      - name: Install prerequisites and download drivers
+        shell: pwsh
+        run: .\build.ps1 driver -prereqs
+      - name: Generate API
+        run: .\build.ps1 roll
+        shell: pwsh
+      - name: Verify generated API is up to date
+        run: git diff --exit-code


.github/workflows/integrity-checks.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: .\build.ps1 -prereqs -driver
+      - name: Read commit SHA
+        id: read-upstream-commit-sha
+        run: echo "::set-output name=hash::$(cat src/Playwright/.drivers/linux/package/commitinfo)"
+      - name: Checkout submodule
+        run: git checkout ${{ steps.read-upstream-commit-sha.outputs.hash }}
+        working-directory: ./playwright
+      - name: Generate API
+        run: .\build.ps1 -api
+        shell: pwsh
```
mxschmitt, line 37: what is pwsh?

.github/workflows/integrity-checks.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          dotnet-version: 5.0.x
+      - name: Install prerequisites and download drivers
+        shell: pwsh
+        run: .\build.ps1 -prereqs -driver
+      - name: Read commit SHA
+        id: read-upstream-commit-sha
+        run: echo "::set-output name=hash::$(cat src/Playwright/.drivers/linux/package/commitinfo)"
+      - name: Checkout submodule
+        run: git checkout ${{ steps.read-upstream-commit-sha.outputs.hash }}
```
mxschmitt, line 33: if this means we need still maintain the revision manually by running "git checkout" I think we don't need the submodule stuff. I thought the main point of it was that the submodule is pinned to the upstream commit.

.github/workflows/integrity-checks.yml
avodovnik: https://docs.github.com/en/actions/guides/building-and-testing-powershell

.github/workflows/integrity-checks.yml
avodovnik: Sort of. 

I could have just checked the commit id of the submodule, and if that matches the driver, be done with it. But that's not really good enough IMHO.

So, I explicitly checkout the commit **from the driver** and verify there are no API changes in that. The submodule doesn't change it drastically, aside from the fact it's there.

avodovnik: To reiterate, submodule mainly helps with development and rolling, but it's still a tool that needs to be used; meaning, it won't automagically solve all our rolling problems. But I'm betting it will reduce some - hence why I want to try them in one of our ports.

aslushnikov (commented):
LGTM

PR #1626, opened 2021-07-21
chore(tests): fix local nuget test, remove unused actions

avodovnik:


src/Playwright.LocalNugetTest/UnitTest1.cs
```diff
@@ -22,26 +22,21 @@
...
-            using var playwright = await Playwright.CreateAsync();
-            await using var browser = await playwright.Chromium.LaunchAsync();
-
-            var page = await browser.NewPageAsync();
-            Console.WriteLine("Navigating google");
-            await page.GotoAsync("http://www.google.com");
-
-            Assert.Contains("Google", await page.TitleAsync());
+            TestContext.WriteLine("Navigating to Microsoft.com...");
+            await Page.GotoAsync("https://microsoft.com");
```
mxschmitt, line 38: hmm, external sites are not ideal there, maybe Just a Eval 1+1 == 2?

.github/workflows/nuget-package-tests.yml
```diff
@@ -15,35 +15,36 @@ jobs:
...
-        run: rm -r ~/.cache/ms-playwright
-      - name: Delete browsers on windows
-        if: matrix.os == 'windows-latest'
-        run: Remove-Item -path $env:USERPROFILE\AppData\Local\ms-playwright -recurse
+          dotnet-version: 5.0.x
+      - name: Install prerequisites and download drivers
+        shell: pwsh
+        run: .\build.ps1 driver -prereqs
+      - name: Install Playwright CLI
+        run: dotnet tool install --global Microsoft.Playwright.CLI --version 1.0.1-next-1
```
mxschmitt, line 32: maybe not hardcode the version there? but unsure.

mxschmitt (approved):

.github/workflows/nuget-package-tests.yml
avodovnik: So, this ~is~ was there because CLI wasn't pushed until yesterday, and the tool will default to the release version which failed to correctly find the DLLs needed in this case.

avodovnik, line 32:
```suggestion
        run: dotnet tool install --global Microsoft.Playwright.CLI
```

.github/workflows/nuget-package-tests.yml
```diff
@@ -15,35 +15,36 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-20.04, macos-latest]
```
mxschmitt, line 18: nit: be consistent with version there (everywhere pinning or everywhere latest - I personally prefer the first one)

avodovnik:
```suggestion
        os: [windows-latest, ubuntu-latest, macos-latest]
```

PR #1631, opened 2021-07-22
chore(roll): update README with relevant versions

avodovnik:


mxschmitt (approved):
can we add this to the powershell script that this gets done automatically after a roll and get verified on CI that this is always up2date?

(This what we do in Java + Python)

avodovnik:
It's _technically_ already there, but the readme source didn't exist for the version yet. I'll replace most of that logic when I update to the `sha` based download (hopefully).

PR #1647, opened 2021-07-29
chore(tests): fix stalling SSL tests on Mac OS

avodovnik:


.github/workflows/tests.yml
```diff
@@ -21,129 +21,77 @@ jobs:
...
-      - name: Download Drivers
-        run: dotnet run -p ./src/tools/Playwright.Tooling/Playwright.Tooling.csproj -- download-drivers --basepath .
-      - name: Build CLI
-        run: dotnet build ./src/Playwright.CLI/
-      - name: Build NUnit
-        run: dotnet build ./src/Playwright.NUnit/
+          BROWSER: CHROMIUM
+        working-directory: src
+        run: |
+          xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- bash -c "ulimit -c unlimited && dotnet test ./Playwright.Tests/Playwright.Tests.csproj -c Debug -f net5.0 --logger \"trx\" -- NUnit.NumberOfTestWorkers=1"
```
mxschmitt, line 97:
```suggestion
          xvfb-run dotnet test ./Playwright.Tests/Playwright.Tests.csproj -c Debug -f net5.0 --logger "trx" -- NUnit.NumberOfTestWorkers=1
```
lets do what our users do, for Python we also have it like that.

.github/workflows/tests.yml
```diff
@@ -21,129 +21,77 @@ jobs:
...
-          dotnet dev-certs https --clean
-          dotnet dev-certs https -ep src/Playwright.Tests.TestServer/testCert.cer
-      - name: Clean
-        run: dotnet clean ./src/Playwright.sln --configuration Debug && dotnet nuget locals all --clear
-      - name: Download Drivers
-        run: dotnet run -p ./src/tools/Playwright.Tooling/Playwright.Tooling.csproj -- download-drivers --basepath .
-      - name: Download Browsers
-        run: dotnet run -p ./src/Playwright -f net5.0 install
-      - name: Run tests (!Linux)
+          dotnet run -p .\src\Playwright\Playwright.csproj -f net5.0 -- install
```
mxschmitt, line 42:
```suggestion
          playwright install
```
why does this not work yet?

avodovnik: The CLI isn't installed. Nor do I want to install it - to avoid running into weird issues with versioning/NuGet packages.

.github/workflows/tests.yml
```diff
@@ -21,129 +21,77 @@ jobs:
...
-        if: ${{ matrix.os == 'macos-latest' }}
-        uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 3.1.100
-        env:
-          DOTNET_INSTALL_DIR: /users/runner/alternate/dotnet
-      - name: Setup .NET Core (Not on mac)
-        if: ${{ matrix.os != 'macos-latest' }}
+          submodules: true
+          fetch-depth: 0
```
mxschmitt, line 68: why fetch-depth: 0?

avodovnik: IIRC this was needed to ensure I get the submodule at the right commit

avodovnik: Also, that was true for the integrity check PR. It's possible I could get away with `1` here. I'll give it a go. It might speed it up ever so slightly.

.github/workflows/tests.yml
avodovnik: Sure, this is a legacy thing that was in here since the beginning. If we can simplify it, I'm all for it.

mxschmitt (approved):

PR #1650, opened 2021-08-02
feat(test): introduce Playwright.MSTest

avodovnik:


src/Playwright.Testing.Core/IBrowserTest.cs
```diff
@@ -0,0 +1,35 @@
...
+ * SOFTWARE.
+ */
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Playwright.Testing.Core
+{
+    public interface IBrowserTest
+    {
+        public IBrowser Browser { get; }
```
Meir017, line 31:
```suggestion
        IBrowser Browser { get; }
```
same for rest of properties

avodovnik:
Based on a conversation with @pavelfeldman, I'll be changing this a little bit - moving back to draft.

PR #1655, opened 2021-08-04
fix(cli): fix .NET's escaping of quotes in cli args

### LINKED ISSUE #1653: [Bug]: playwright CLI run emulate devices script won't work
### Playwright version

Version 1.13.0-1626733671000

### Operating system

Windows

### What browsers are you seeing the problem on?

Chromium

### Other information

.net 6 preview6

### What happened? / Describe the bug

playwright CLI run Emulate devices script won't work

### Code snippet to reproduce your bug

```shell
cmd$> playwright open --device="iPhone 11" wikipedia.org
```

### Relevant log output

```shell
Device descriptor not found: 'iPhone', available devices are:
...
...
```

avodovnik:
Closes #1653

### DIFF
--- a/src/Playwright/Program.cs
+++ b/src/Playwright/Program.cs
@@ -39,11 +39,7 @@ namespace Microsoft.Playwright
 
         public int Run(string[] args)
         {
-            // we need to use this logic, because .NET by design does some weird magic with the quotes
-            // so, we'll use a bit of our own magic... We'll start by calling the same base mathod, however
-            // we know it'll return the executable & path in the first argument. We'll use this to replace
-            // the entire string, which we can access from the "original" command line that we were called with
-            // which doesn't have the quotes stripped, but does include the full path
+            // we need to use the original command line to avoid getting quotes stripped by the runtime
             // see https://github.com/microsoft/playwright-dotnet/issues/1653
             args = Environment.GetCommandLineArgs();
             var lineArguments = Environment.CommandLine.Replace(args[0], string.Empty);
--- a/src/Playwright/Program.cs
+++ b/src/Playwright/Program.cs
@@ -39,23 +39,14 @@ namespace Microsoft.Playwright
 
         public int Run(string[] args)
         {
-            args ??= Array.Empty<string>();
-            for (int i = 0; i < args.Length; i++)
-            {
-                var a = args[i];
-                if (a.Contains(" "))
-                {
-                    var left = a.IndexOf('=');
-                    if (left >= 0)
-                    {
-                        args[i] = a.Substring(0, left + 1) + "\"" + a.Substring(left + 1, a.Length - left - 1) + "\"";
-                    }
-                    else
-                    {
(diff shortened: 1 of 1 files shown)


src/Playwright/Program.cs
```diff
@@ -39,6 +39,24 @@ public static int Main(string[] args)
...
+            args ??= Array.Empty<string>();
+            for (int i = 0; i < args.Length; i++)
+            {
+                var a = args[i];
+                if (a.Contains(" "))
+                {
+                    var left = a.IndexOf('=');
+                    if (left >= 0)
+                    {
+                        args[i] = a.Substring(0, left + 1) + "\"" + a.Substring(left + 1, a.Length - left - 1) + "\"";
```
yury-s, line 51: what if value is already quoted, won't it end up double quoting it?

yury-s (approved):
This place looks rather brittle, can you add a test for this with a couple of ways the args could be passed?

src/Playwright/Program.cs
avodovnik: .NET will end up stripping those quotes.

avodovnik:
Just for future reference:

```powershell
# The values returned from Environment.GetCommandLineArgs();
C:\Users\anvod\.dotnet\tools\.store\microsoft.playwright.cli\1.0.0\microsoft.playwright.cli\1.0.0\tools\net5.0\any\Microsoft.Playwright.CLI.dll open --device iPhone 11 wikipedia.org
# Our parsed one
 open --device "iPhone 11" wikipedia.org
```

src/Playwright/Program.cs
```diff
@@ -39,6 +39,15 @@ public static int Main(string[] args)
 
         public int Run(string[] args)
         {
+            // we need to use this logic, because .NET by design does some weird magic with the quotes
+            // so, we'll use a bit of our own magic... We'll start by calling the same base mathod, however
+            // we know it'll return the executable & path in the first argument. We'll use this to replace
+            // the entire string, which we can access from the "original" command line that we were called with
+            // which doesn't have the quotes stripped, but does include the full path
```
yury-s, line 46: I'd abbreviate the comment to something like 'Use original command line which doesn't have the quotes stripped.'

yury-s (approved):

src/Playwright/Program.cs
avodovnik, line 46:
```suggestion
            // we need to use the original command line to avoid getting quotes stripped by the runtime
```

avodovnik:
Including the "original" comment here, to keep track of the reasoning:

> We need to use this logic, because .NET by design does some weird magic with the quotes, so, we'll use a bit of our own magic... We'll start by calling the same base mathod, however we know it'll return the executable & path in the first argument. We'll use this to replace the entire string, which we can access from the "original" command line that we were called with which doesn't have the quotes stripped, but does include the full path.

PR #1657, opened 2021-08-04
chore: fix response test

avodovnik:


src/Playwright.Tests/PageNetworkResponseTests.cs
```diff
@@ -171,8 +172,25 @@ public async Task ShouldReturnSecurityDetails()
         {
             var response = await Page.GotoAsync(HttpsServer.EmptyPage);
             var details = await response.SecurityDetailsAsync();
-            Assert.IsNotEmpty(details.Issuer);
-            Assert.IsNotEmpty(details.Protocol);
+
+            var name = "puppeteer-tests";
```
mxschmitt, line 176: pretty hard to read IMO, is it possible to assert the whole details object like we do in other languages?
- https://github.com/microsoft/playwright/blob/master/tests/har.spec.ts#L447
- https://github.com/microsoft/playwright-python/blob/39a9cb13d00486d5b368cf7d5f71abcd1bcd9939/tests/async/test_network.py

avodovnik: No, it's not. Not without doing a ton of extra leg work that adds no value (implementing custom `IComparable`) which I don't want to  do for something small like this test.

And yes, the TLS 1.3 is annoying, but again, not a deal-breaker. Also, we're not testing the test server's TLS capabilities, so I'm not particularly bothered by it.

mxschmitt (approved):

PR #1667, opened 2021-08-11
chore(cli): target netcoreapp3.1

martincostello:
Add `netcoreapp3.1` as a Target Framework so there is a supported LTS release targeted by the CLI tool once .NET Core 2.1 goes out of support [later this month](https://devblogs.microsoft.com/dotnet/net-august-2021/#net-core-2-1-end-of-life).


avodovnik (commented):
I'm unconvinced at the moment - shouldn't we just replace the targeting of `2.1` with `3.1` in this case?

martincostello:
I could also do that - but removing a TFM (though it is soon unsupported) is I believe a breaking change from a SemVer perspective, so that's why for now I just _added_ one.

If you want the old one removed, then I can do that instead.

avodovnik:
@martincostello We've had a conversation in the team about this, but I'm still waiting for one more piece of feedback from someone internally, for us to form a proper plan on how to address this.

Right now, it's leaning towards the same approach we follow upstream, which is that we **do not** consider removing a target framework for an unsupported LTS version, a breaking change. I'll keep you posted and I'll merge the PR ASAP.

martincostello:
Cool, thanks 👍

avodovnik (approved):

PR #1673, opened 2021-08-12
fix(timeout): default timeout was being set to 0 instead of null

### LINKED ISSUE #1672: [Bug]: RunAndWaitForNavigationAsync default timeout
### Playwright version

1.13.0

### Operating system

MacOS

### What browsers are you seeing the problem on?

Chromium

### Other information

_No response_

### What happened? / Describe the bug

When I use `RunAndWaitForNavigationAsync` on an `IPage`, it never times out. The documentation says the default is 30 seconds, so I did not expect my test to wait for the heat death of the universe.

I narrowed this down to `(Page as Page)?.DefaultNavigationTimeout` being 0 in [Frame.cs](https://github.com/microsoft/playwright-dotnet/blob/9440246c6606f34a190829bd772e61e0b84c7ff9/src/Playwright/Core/Frame.cs#L538).

### Code snippet to reproduce your bug

```shell
var playwright = await Playwright.CreateAsync();
var browser = await playwright.Chromium.LaunchAsync(new BrowserTypeLaunchOptions { Headless = false });

// It works as expected if I set it manually like this:
// page.SetDefaultNavigationTimeout(2_000);

await page.RunAndWaitForNavigationAsync(async () =>
{
    await page.GotoAsync("https://kernel.org");
    var element = await page.QuerySelectorAsync("a[href=\"./category/about.html\"]");
    await element.ClickAsync();
}, new PageRunAndWaitForNavigationOptions
{
    UrlRegex = new Regex(@"not-about.html$"),

    // It also works if I set it manually like this:
    // Timeout = 2_000,
});
```

### Relevant log output

_No response_

avodovnik:
Closes #1672

### DIFF
--- a/src/Playwright.Tests/Attributes/PlaywrightTestAttribute.cs
+++ b/src/Playwright.Tests/Attributes/PlaywrightTestAttribute.cs
@@ -33,6 +33,10 @@ namespace Microsoft.Playwright.Tests
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
     public class PlaywrightTestAttribute : TestAttribute
     {
+        public PlaywrightTestAttribute()
+        {
+        }
+
         /// <summary>
         /// Creates a new instance of the attribute.
         /// </summary>
--- a/src/Playwright.Tests/PageWaitForNavigationTests.cs
+++ b/src/Playwright.Tests/PageWaitForNavigationTests.cs
@@ -324,5 +324,22 @@ namespace Microsoft.Playwright.Tests
             Assert.AreEqual(frame, response.Frame);
             StringAssert.Contains("/frames/one-frame.html", Page.Url);
         }
+
+        [PlaywrightTest]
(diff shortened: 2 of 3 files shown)


src/Playwright/Transport/Channels/PageChannel.cs
```diff
@@ -148,7 +148,7 @@ internal Task SetDefaultTimeoutNoReplyAsync(float timeout)
                     ["timeout"] = timeout,
```
yury-s, line 148: According to protocol.yml timeout is no optional number, won't it throw here if the timeout is null?

avodovnik: Yeah, that's a good point. Thanks for pointing that out. 

I've changed the implementation so that only the backing field can be null in which case the Property itself defaults to `PlaywrightImpl.DefaultTimeout`.

yury-s (approved):

PR #1684, opened 2021-08-16
chore: introduce optional string helper

avodovnik:
When generating, I'll make use of this helper, as it saves a bunch of boilerplate code in several calls.

### DIFF
--- a/src/Playwright/Helpers/JsonExtensions.cs
+++ b/src/Playwright/Helpers/JsonExtensions.cs
@@ -97,5 +97,25 @@ namespace Microsoft.Playwright.Helpers
                     new JsonStringEnumMemberConverter(JsonNamingPolicy.CamelCase),
                 },
             };
+
+#nullable enable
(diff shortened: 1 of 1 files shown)


src/Playwright/Helpers/JsonExtensions.cs
```diff
@@ -97,5 +97,25 @@ internal static JsonSerializerOptions GetNewDefaultSerializerOptions()
                     new JsonStringEnumMemberConverter(JsonNamingPolicy.CamelCase),
                 },
             };
+
+#nullable enable
+        internal static string? ToOptionalString(this JsonElement? element, string name)
```
pavelfeldman, line 102: `GetOptionalString` ?

pavelfeldman (approved):

src/Playwright/Helpers/JsonExtensions.cs
avodovnik: `To` is more consistent with the rest of the helpers, like `ToString`, `ToObject`, etc.

PR #1696, opened 2021-08-19
chore: auto-generate channels, part 1

avodovnik:


src/Playwright/Helpers/ChannelHelpers.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Playwright.Transport.Channels;
+
+#nullable enable
+#pragma warning disable IDE0018 // Inline variable declaration will cause issues due to Roslyn design: https://github.com/dotnet/roslyn/issues/54711
+
+namespace Microsoft.Playwright.Helpers
+{
+    internal static class ChannelHelpers
+    {
+        internal static string? GetString(this JsonElement? element, string name, bool optional)
```
avodovnik, line 14: @pavelfeldman your suggestion fit better now that I've refactored a bunch of the other calls... it feels more natural to say `Get`...

src/Playwright/Transport/Channels/BindingCallChannel.cs
```diff
@@ -22,43 +22,14 @@
...
 
-using System;
-using System.Collections.Generic;
-using System.Threading.Tasks;
 using Microsoft.Playwright.Core;
 
 namespace Microsoft.Playwright.Transport.Channels
 {
-    internal class BindingCallChannel : Channel<BindingCall>
+    internal class BindingCallChannel : BindingCallChannelImpl
```
yury-s, line 29: ditto

src/Playwright/Transport/Channels/SelectorsChannel.cs
```diff
@@ -27,13 +27,10 @@
 
 namespace Microsoft.Playwright.Transport.Channels
 {
-    internal class SelectorsChannel : Channel<Selectors>
+    internal class SelectorsChannel : SelectorsChannelImpl
```
yury-s, line 30: ditto

src/Playwright/Transport/Channels/WorkerChannel.cs
```diff
@@ -23,14 +23,12 @@
...
 using System;
-using System.Collections.Generic;
 using System.Text.Json;
-using System.Threading.Tasks;
 using Microsoft.Playwright.Core;
 
 namespace Microsoft.Playwright.Transport.Channels
 {
-    internal class WorkerChannel : Channel<Worker>
+    internal class WorkerChannel : WorkerChannelImpl
```
yury-s, line 31: ditto

yury-s (approved):
Can we suppress this 'Unchanged files with check annotations' messages? they take so much of screen estate with pure noise

avodovnik:
> Can we suppress this 'Unchanged files with check annotations' messages? they take so much of screen estate with pure noise

Yup: https://github.com/microsoft/playwright-dotnet/pull/1713

src/Playwright/Core/Stream.cs
```diff
@@ -29,15 +29,15 @@ namespace Microsoft.Playwright.Core
...
-        private readonly StreamChannel _channel;
-
         internal Stream(IChannelOwner parent, string guid) : base(parent, guid)
         {
-            _channel = new(guid, parent.Connection, this);
+            Channel = new(guid, parent.Connection, this);
         }
 
-        ChannelBase IChannelOwner.Channel => _channel;
+        ChannelBase IChannelOwner.Channel => Channel;
```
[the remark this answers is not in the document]
avodovnik: Yeah, it's a weird consequence of the current architecture. I'll see if I can get rid of this later. But the three properties with the same name return different (sub)types, i.e. `ChannelBase`, `IChannel<Stream>` and `StreamChannel`. This is something I want to get rid of, but I need to replace all of the channels first.

src/Playwright/Transport/Channels/ArtifactChannel.cs
```diff
@@ -1,61 +1,36 @@
...
+*
+*/
 
-using System.Threading.Tasks;
 using Microsoft.Playwright.Core;
 
 namespace Microsoft.Playwright.Transport.Channels
 {
-    internal class ArtifactChannel : Channel<Artifact>
+    internal class ArtifactChannel : ArtifactChannelImpl
```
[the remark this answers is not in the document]
avodovnik: No, because I need to be able to override some actions and I can't have them be overriden when generated. But, I can generate the original class and add it next to the `impl`, marking it partial. That way, at least I'll only need a separate file *if* I need to override behaviour. Does that sound better?

PR #1701, opened 2021-08-20
chore: roll to latest 1.14

avodovnik:


src/Playwright/API/Generated/IBrowserType.cs
```diff
@@ -66,11 +66,6 @@ namespace Microsoft.Playwright
     /// </summary>
     public partial interface IBrowserType
     {
-        /// <summary><para>This methods attaches Playwright to an existing browser instance.</para></summary>
```
mxschmitt, line 69: Is this removal expected? I think it's better to not roll the release branch against the master branch in this case.

avodovnik: I think this should be a draft for now. I was working on this then ended up fixing some other stuff before getting back to it.

avodovnik:
Killing this one, as I'll roll directly to 1.15

PR #1709, opened 2021-08-23
fix: improve flakiness due to TLS versions

avodovnik:
TLS version mismatch on platforms is causing the test to be flaky yet the assertion provides no extra value


mxschmitt (approved):
I would use "starts with TLS " instead.

PR #1713, opened 2021-08-24
chore(mstest): fix CI warnings and introduce basic smoke test for mstest

### LINKED ISSUE #1711: [Internal] Disable / Fix warnings which gets reflected in GitHub Actions
There is an enormous amount of linting warnings in Pull Requests caused by the CI runs. It seems like they are not deduplicated because linting runs on every CI job. So maybe run linting only a single time, fix the linting warnings or disable them might make sense.

image

avodovnik:
Also fixes #1711

### DIFF
--- a/src/Playwright.TestingHarnessTest/SimpleSmokeTests.NUnit.cs
+++ b/src/Playwright.TestingHarnessTest/SimpleSmokeTests.NUnit.cs
@@ -1,0 +1,29 @@
+using System.Threading.Tasks;
+using Microsoft.Playwright.NUnit;
+using NUnit.Framework;
+
+namespace Playwright.Tests.MSTest
+{
+        
+    public class SimpleSmokeTestsNUnit : PageTest
+    {
+        [Test]
+        public async Task ShouldOpenPlaywright()
+        {
+            Assert.IsNotNull(Page);
+            await Page.GotoAsync("https://www.playwright.dev");
+            var h1 = await Page.TextContentAsync("h1");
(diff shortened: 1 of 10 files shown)


src/Playwright.TestingHarnessTest/SimpleSmokeTests.NUnit.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Playwright.Tests.MSTest
+{
+
+    public class SimpleSmokeTestsNUnit : PageTest
+    {
+        [Test]
+        public async Task ShouldOpenPlaywright()
+        {
+            Assert.IsNotNull(Page);
+            await Page.GotoAsync("https://www.playwright.dev");
```
mxschmitt, line 14: Could this be instead either be setContent or evaluate 1+1==2? Unit tests should work without internet.

avodovnik: So, a couple arguments against that:

1) This isn't really a critical part of our test suite.
2) I want these particular "smoke" tests to accurately represent a common customer scenario, which I feel is usually opening a site and checking for some parameters.  I want this suite to grow a little bit throughout time, but again, not testing *our* functionality as much as giving a few real-world examples of where Playwright is useful.
3) In this particular example, considering where they'll be run, I don't mind that much :)

mxschmitt (approved):
(You might need to use https://docs.microsoft.com/en-us/dotnet/api/system.uri.localpath?view=net-5.0)

PR #1715, opened 2021-08-25
chore: roll to 1.15

avodovnik:


src/Playwright/Transport/Channels/Root.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        internal async Task<PlaywrightImpl> InitializeAsync()
+        {
+            var args = new
+            {
+                sdkLanguage = "csharp",
+            };
+
+            var jsonElement = await _connection.SendMessageToServerAsync(string.Empty, "initialize", args).ConfigureAwait(false);
```
mxschmitt, line 47: In theory, if you land your generated channels, you should be able to do `_channel.Initialize(sdkLanguage="csharp")`. And this class inherits then from `IChannelOwner<Root>`.

src/Playwright/Transport/Connection.cs
```diff
@@ -201,6 +209,11 @@ internal JsonSerializerOptions GetDefaultJsonSerializerOptions()
             return options;
         }
 
+        internal Task<PlaywrightImpl> InitializePlaywrightAsync()
+        {
+            return _rootObject.InitializeAsync();
+        }
+
         internal async Task<T> WaitForObjectWithKnownNameAsync<T>(string guid)
```
mxschmitt, line 217: Keep in mind to do the same for BrowserType.Connect and then you can remove the `WaitForObjectWithKnownNameAsync` method.

avodovnik: Yes, I'm working on that one now.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -50,15 +50,15 @@ public void SetUpAsync()
...
                         UseShellExecute = false,
                         RedirectStandardOutput = true,
                         RedirectStandardInput = true,
-                        RedirectStandardError = true,
+                        RedirectStandardError = false,
                         CreateNoWindow = true,
                     },
                 };
-                browserServer.Process.Start();
+                browserServer.Process.Start();
```
mxschmitt, line 61:
```suggestion
                browserServer.Process.Start();
```

src/Playwright.Tests/BrowserContextStrictSelectorsTests.cs
```diff
@@ -0,0 +1,71 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Darío Kondratiuk
```
mxschmitt, line 4: in theory for all new files we can keep MS only copyright?

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -50,11 +50,11 @@ public void SetUpAsync()
                     StartInfo =
                     {
                         FileName = GetExecutablePath(),
-                        Arguments = $"launch-server {BrowserType.Name}",
+                        Arguments = $@"launch-server {BrowserType.Name}",
```
mxschmitt, line 53:
```suggestion
                        Arguments = $"launch-server {BrowserType.Name}",
```

src/Playwright/API/Generated/.editorconfig
```diff
@@ -15,4 +15,8 @@ dotnet_diagnostic.SA1606.severity = none
 dotnet_diagnostic.SA1614.severity = none
 dotnet_diagnostic.SA1615.severity = none
 dotnet_diagnostic.SA1618.severity = none
-dotnet_diagnostic.SA1629.severity = none
\ No newline at end of file
+dotnet_diagnostic.SA1629.severity = none
+dotnet_diagnostic.CA1041.severity = none # [Obsolete] can exist without message
+dotnet_diagnostic.CS0612.severity = none # We can use [Obsolete] in generated code
+dotnet_diagnostic.SA1300.severity = none # Sometimes, we want to use _ in the name
```
mxschmitt, line 21: when do we want to use underscore?

src/Playwright/Core/BrowserContext.cs
```diff
@@ -325,11 +327,12 @@ internal void OnRoute(Route route, IRequest request)
             foreach (var item in _routes)
             {
                 if (
-                    (item.Url != null && UrlMatches(request.Url, item.Url)) ||
+                    ((item.Times ?? 0) == 0 || item.HandledCount >= item.Times) &&
```
mxschmitt, line 330: I would move this (line 330-333) into an own class like we did in JavaScript. Because it's the same logic as in the Page. Then it exposes a Match function and a Handle function.

src/Playwright/Core/Page.cs
```diff
@@ -1005,11 +1011,12 @@ private void Channel_Route(object sender, RouteEventArgs e)
             foreach (var route in _routes)
             {
                 if (
-                    (route.Url != null && Context.UrlMatches(e.Request.Url, route.Url)) ||
+                    ((route.Times ?? 0) == 0 || route.HandledCount >= route.Times) &&
```
mxschmitt, line 1014: see the comment from the context class.

src/Playwright/Playwright.cs
```diff
@@ -1,50 +1,49 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+using System.Diagnostics.CodeAnalysis;
```
mxschmitt, line 25: did you change line endings in this file?

src/Playwright/Transport/Channels/Root.cs
mxschmitt: discussed offline, will follow up on that once we have generated channels.

mxschmitt (approved):

src/Playwright.Tests/BrowserContextStrictSelectorsTests.cs
avodovnik: yeah, copy-pasta error, good spot

src/Playwright/API/Generated/.editorconfig
avodovnik: An API suggestion from upstream would be prefixed `_` and it's actually _acceptable_ IMO. It's a useless rule for generated files.

src/Playwright/Core/BrowserContext.cs
avodovnik: I will likely be doing that once I clean up the channels. For now, keeping as is.

src/Playwright/Playwright.cs
avodovnik: I think they did get changed, yeah. I was sure I fixed this internally at some point -.-

src/Playwright/API/Generated/.editorconfig
mxschmitt: We genuinely don't want underscore APIs in language bindings. If they are generated on your side, we need to fix it in the upstream API (we had it a few times and removed them then, e.g. response interception or fetch). So I'd remove that message pragma.

PR #1719, opened 2021-08-29
Remove unnecessary blocking over async

slang25:


src/Playwright/Core/Locator.cs
```diff
@@ -47,17 +48,25 @@ public Locator(Frame parent, string selector)
...
-            .Select(x => x.Result)
-            .ToArray();
+        {
+            var handles = await _frame.QuerySelectorAllAsync(_selector).ConfigureAwait(false);
+            return (await handles
+                    .Select(async x => await x.InnerTextAsync().ConfigureAwait(false))
+                    .WaitForResultsAsync()
+                    .ConfigureAwait(false))
+                .ToArray();
+        }
```
avodovnik, line 58:
```suggestion
        => (await (await _frame.QuerySelectorAllAsync(_selector).ConfigureAwait(false))
            .SelectAsync(async x => await x.InnerTextAsync().ConfigureAwait(false)).ConfigureAwait(false))
            .Where(x => !string.IsNullOrEmpty(x))
            .ToArray();
```

src/Playwright/Core/Locator.cs
```diff
@@ -47,17 +48,25 @@ public Locator(Frame parent, string selector)
...
-            .ToArray();
+        {
+            var handles = await _frame.QuerySelectorAllAsync(_selector).ConfigureAwait(false);
+            return (await handles
+                    .Select(async x => await x.TextContentAsync().ConfigureAwait(false))
+                    .WaitForResultsAsync()
+                    .ConfigureAwait(false))
+                .Select(x => x ?? string.Empty) // we don't filter nulls, as per https://github.com/microsoft/playwright/blob/master/src/client/locator.ts#L205
+                .ToArray();
+        }
```
avodovnik, line 69:
```suggestion
        => (await (await _frame.QuerySelectorAllAsync(_selector).ConfigureAwait(false))
            .SelectAsync(async x => await x.TextContentAsync().ConfigureAwait(false) ?? string.Empty).ConfigureAwait(false)) // we don't filter nulls https://github.com/microsoft/playwright/blob/master/src/client/locator.ts#L205
            .ToArray();
```

src/Playwright/Helpers/EnumerableExtensions.cs
```diff
@@ -39,5 +40,10 @@ public static IEnumerable<NameValueEntry> Remap(this IEnumerable<KeyValuePair<st
 
             return input.Select(x => new NameValueEntry(x.Key, x.Value)).ToArray();
         }
+
+        public static async Task<IEnumerable<T>> WaitForResultsAsync<T>(this IEnumerable<Task<T>> listOfTasks)
+        {
+            return await Task.WhenAll(listOfTasks).ConfigureAwait(false);
+        }
```
avodovnik, line 47:
```suggestion
        public static async Task<IEnumerable<TResult>> SelectAsync<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, Task<TResult>> method)
        {
            return await Task.WhenAll(source.Select(x => method(x))).ConfigureAwait(false);
        }
```

avodovnik (commented):
Thanks for the suggestion! I've taken the liberty to propose some changes to simplify the code a bit more.

src/Playwright/Helpers/EnumerableExtensions.cs
```diff
@@ -24,6 +24,7 @@
 
 using System.Collections.Generic;
```
avodovnik, line 25:
```suggestion
using System;
using System.Collections.Generic;
```

avodovnik: This will fix the errors.

src/Playwright/Core/Locator.cs
slang25: Thanks @avodovnik, I've update this PR with your `SelectAsync` suggested change.
I deliberately refactored to avoid `(await (await ` as I found it a bit disorientating personally. Let me know if you disagree and I'll apply the suggestion here 🙂

avodovnik (approved):
Thanks for the contribution, and good spot!

Issue #1730, opened 2021-09-06
[Question]: Can I provide a Chromium channel when using PageTest in unit?


jevgenijusmarinuskinas:
### Your question

Hello,

I can specify BROWSER and HEADED through the means of environment variables, to be used with a PageTest base class in my code. I've looked through the playwright sources, and was unable to find an environment variable for a Chromium channel. Is it possible to somehow control it in this scenario (for example, by setting some property, etc.)?

avodovnik:
No, unfortunately this isn't possible at the moment. I'll convert this into a feature request though, and we can explore adding it.

PavanMudigondaTR:
I have the same question and request. Playwright supports browser_channel environment variable for Python, Java, JS/TS. I am not seeing it working in dotnet. I am struggling to figure out how to run Github Actions for Chrome and MSEdge.

PR #1737, opened 2021-09-13
Enabled channel and slowmo environment variables for testing

mlankamp:
Enabled 2 additional environment variables for running tests with mstest or nunit. Allows running on a chromium channel and allows slowmo to see what's going on.

**Linux**
```
HEADED=1 CHANNEL=chrome-dev SLOWMO=1000 dotnet test
```

**Windows** 
```
set HEADED=1
set CHANNEL=chrome-dev
set SLOWMO=1000
dotnet test
```


mlankamp:
Solution for #1730

avodovnik:
Thanks for the contribution!

I've actually started work on a feature that would also enable this, I have a draft PR up at #1743. I want to make some changes to the way how configuration can be done, outside of "just" using env variables.

mlankamp:
The focus now is on environment variables, but how could we also use runsettings files for configuration? 

The settings are for browserlaunch options, but maybe we want to enable some default context options also, for example the locale or override the user agent.

vertonghenb:
Is there a reason why this is not being approved?

PR #1748, opened 2021-09-17
[WIP] feat: roll to 1.15 - part I, interface changes, does not build

avodovnik:


avodovnik:
When #1756 lands, we need to drop `bdbcc34923ec427913cf11cf08b9c55a17a9ab04` and rebase.

src/Playwright/Core/Response.cs
```diff
@@ -89,10 +81,21 @@ internal Response(IChannelOwner parent, string guid, ResponseInitializer initial
 
         IChannel<Response> IChannelOwner<Response>.Channel => _channel;
 
+        public async Task<Dictionary<string, string>> AllHeadersAsync()
```
mxschmitt, line 84: >  values.join(name.toLowerCase() === 'set-cookie' ? '\n' : ', ');

this seems missing e.g.

src/Playwright.Tests/PageNetworkResponseTests.cs
```diff
@@ -62,7 +62,7 @@ public async Task ShouldWork()
             });
 
             var response = await Page.GotoAsync(Server.EmptyPage);
-            StringAssert.Contains("bar", response.Headers["foo"]);
+            StringAssert.Contains("bar", await response.HeaderValueAsync("foo"));
```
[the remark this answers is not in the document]
avodovnik: Yup, makes sense. It was an interim fix.

src/Playwright.Tests/PageNetworkRequestTest.cs
```diff
@@ -301,11 +300,46 @@ public async Task ShouldSetBodySizeTo0WhenNoResponseBody()
...
+        ['header-a', 'value-a'],
+        ['header-b', 'value-b'],
+        ['header-a', 'value-a-1'],
+        ['header-a', 'value-a-2'],
+      ]
+    })
+");
+            await Task.WhenAll(requestTask, evalTask);
+            var req = requestTask.Result;
+            Assert.AreEqual("value-a, value-a-1, value-a-2", await req.HeaderValueAsync("header-a"));
```
pavelfeldman, line 341: The actual comparison seems to be missing

```
expect(headers.sort((a, b) => a.name.localeCompare(b.name))).toEqual(expectedHeaders.sort((a, b) => a.name.localeCompare(b.name)));
```

src/Playwright/API/Generated/IRequest.cs
```diff
@@ -91,9 +99,27 @@ public partial interface IRequest
...
         Dictionary<string, string> Headers { get; }
 
+        /// <summary>
+        /// <para>
+        /// An array with all the request HTTP headers associated with this request. Unlike
+        /// <see cref="IRequest.AllHeadersAsync"/>, header names are NOT lower-cased. Headers
+        /// with multiple entries, such as <c>Set-Cookie</c>, appear in the array multiple times.
+        /// </para>
+        /// </summary>
+        Task<IReadOnlyList<RequestHeadersArrayResult>> HeadersArrayAsync();
```
pavelfeldman, line 117: RequestHeadersArrayResult - this name is confusing - it is not an array. We already have a NameValueEntry, should we get a public type for it?

src/Playwright/API/Generated/Types/ResponseHeadersArrayResult.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Threading.Tasks;
+
+#nullable enable
+
+namespace Microsoft.Playwright
+{
+    /// <summary>
+    /// Result of calling <see cref="IResponse.HeadersArrayAsync"/>.
+    /// </summary>
+    public partial class ResponseHeadersArrayResult
```
pavelfeldman, line 45: We should make NameValue a special class instead. Like dictionary for objects...

src/Playwright/Core/Page.cs
```diff
@@ -286,8 +286,9 @@ public Task EmulateMediaAsync(PageEmulateMediaOptions options = default)
             var args = new Dictionary<string, object>
             {
                 ["media"] = options?.Media,
-                ["colorScheme"] = options?.ColorScheme,
-                ["reducedMotion"] = options?.ReducedMotion,
+                ["colorScheme"] = options?.ColorScheme ?? ColorScheme.Null,
```
pavelfeldman, line 289: I think there was a semantics difference between the Null and no value that is regressing.

src/Playwright/Core/RawHeaders.cs
```diff
@@ -0,0 +1,46 @@
...
+            foreach (var entry in headers)
+            {
+                var name = entry.Name.ToLower();
+                if (!_headersMap.TryGetValue(name, out List<string> values))
+                {
+                    values = new List<string>();
+                }
+
+                values.Add(entry.Value);
+                _headersMap[name] = values;
```
pavelfeldman, line 23: Move this to line 19.5

src/Playwright/Core/Request.cs
```diff
@@ -53,17 +56,7 @@ internal Request(IChannelOwner parent, string guid, RequestInitializer initializ
...
-            {
-                var name = kv.Name.ToLower();
-
-                // There are case-sensitive dupes :/
-                if (!Headers.ContainsKey(name))
-                {
-                    Headers.Add(kv.Name.ToLower(), kv.Value);
-                }
-            }
+            _headers = new RawHeaders(initializer.Headers.ConvertAll(x => new NameValueEntry(x.Name, x.Value)).ToArray());
```
pavelfeldman, line 59: nit: Feels like we could build RawHeaders off initializer.Headers

src/Playwright/Core/Request.cs
```diff
@@ -130,5 +122,38 @@ internal Request(IChannelOwner parent, string guid, RequestInitializer initializ
...
+
+        public async Task<IReadOnlyList<RequestHeadersArrayResult>> HeadersArrayAsync()
+            => (await GetRawHeadersAsync().ConfigureAwait(false)).HeadersArray.ConvertAll(x => new RequestHeadersArrayResult() { Name = x.Name, Value = x.Value });
+
+        public async Task<string> HeaderValueAsync(string name)
+            => (await GetRawHeadersAsync().ConfigureAwait(false)).Get(name);
+
+        private async Task<RawHeaders> GetRawHeadersAsync()
+        {
+            if (_rawHeaders != null) return _rawHeaders;
```
pavelfeldman, line 147: You need to cache the Task, not the value. Otherwise you are sending request multiple times.

src/Playwright/Core/Response.cs
```diff
@@ -113,5 +107,14 @@ internal void ReportFinished(string erroMessage = null)
         {
             _finishedTask.SetResult(erroMessage);
         }
+
+        private async Task<RawHeaders> GetRawHeadersAsync()
+        {
+            if (_rawHeaders != null) return _rawHeaders;
```
pavelfeldman, line 113: ditto

src/Playwright/Transport/Channels/BrowserContextChannel.cs
```diff
@@ -59,7 +59,7 @@ public BrowserContextChannel(string guid, Connection connection, BrowserContext
 
         internal event EventHandler<BrowserContextChannelResponseEventArgs> Response;
 
-        internal override void OnMessage(string method, JsonElement? serverParams)
+        internal override async Task OnMessageAsync(string method, JsonElement? serverParams)
```
pavelfeldman, line 62: This looks like an infra change that needs to land separately, before this PR?

src/Common/Version.props
```diff
@@ -16,4 +16,4 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
   </PropertyGroup>
-</Project>
+</Project>
```
Archish27, line 19: EOL fixes

src/Playwright/Transport/Channels/PageChannel.cs
```diff
@@ -117,7 +117,29 @@ internal override void OnMessage(string method, JsonElement? serverParams)
...
+                    {
+                        if ("beforeunload".Equals(dialog.Type, StringComparison.Ordinal))
+                        {
+                            try
+                            {
+                                await dialog.AcceptAsync(null).ConfigureAwait(false);
+                            }
+                            catch (PlaywrightException)
+                            {
+                                // Noop
```
Archish27, line 135: Do we need to log anything. Here if something goes wrong, it will be silent here.

src/Common/Version.props
avodovnik: This gets overwritten by the roll tool that we have, not sure this change is relevant.

src/Playwright/API/Generated/IRequest.cs
avodovnik: IMHO, the _right_ thing in that case, if we don't want to use the auto-generated type, is to use `KeyValuePair<string, string>`. I had that initially, but wasn't sure if I should keep it.

src/Playwright/Core/Page.cs
avodovnik: Hm, I fixed the tests to be aligned with upstream. Unless I'm missing something else?

PR #1756, opened 2021-09-23
feat: when rolling, also copy the wwwroot over

avodovnik:


mxschmitt (approved):
Lets maybe get completely rid of the www and do it via gitignore + sync from upstream instead?

mxschmitt (approved):

PR #1740, opened 2021-09-13
fix(CLI): use drivers instead of loading dll

avodovnik:
Closes #1694 

This patch stops the CLI from loading the DLL, which was causing problems with projects targeting .NET Framework 4.8 and instead relies on the `.playwright` folder existing somewhere underneath the root, or specified project path.

PR #1761, opened 2021-09-27
Revert #1738 #1740

pavelfeldman:
This reverts commit 167080c6e04228909cc6b9d156bb8752791c031c.


dgozman (approved):

srigumm:
@pavelfeldman 

Can you help me to understand why the below commit is reverted? I am still facing this issue, #1740 would have fixed this issue.

Revert "fix(CLI): use drivers instead of loading dll (microsoft#1740)"

Thank you.

pavelfeldman:
This commit was reverted provisionally. I was pushing the release and reverted everything that did not look solid. Now that I looked at it, I can see two problems:
- If invoked in C: it was traversing entire filesystem
- If was introducing the Tool -> Playwright dependency, explicitly against the tool design.

pavelfeldman:
@srigumm: I'm following up with the fix to your issue in 1.15.3. In fact I think it wasn't broken before 1.15:

- Try installing Microsoft.Playwright [1.14.1](https://www.nuget.org/packages/Microsoft.Playwright/1.14.1)
- instead of `playwright install` do `./bin/Debug/net48/playwright.ps1 install`

I'll revert more changes on 1.15 to bring it into that former working state.

PR #1813, opened 2021-11-01
Update StdIOTransport.cs

muscatrichard:
Adds `_process.Dispose()` to release other resources such as Node.exe

### DIFF
--- a/src/Playwright/Transport/StdIOTransport.cs
+++ b/src/Playwright/Transport/StdIOTransport.cs
@@ -79,6 +79,7 @@ namespace Microsoft.Playwright.Transport
                 _readerCancellationSource?.Cancel();
                 try
                 {
+                    _process?.Dispose();
                     _process?.Kill();
                 }
                 catch


src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -79,6 +79,7 @@ public void Close(string closeReason)
                 _readerCancellationSource?.Cancel();
                 try
                 {
+                    _process?.Dispose();
```
fr4gles, line 82: Shouldn't `Kill` be called before `Dispose`?

jasomusc: Doesn't make a difference. It still does what it needs to do and close the node.exe instead of leaving it orphan in the processes causing memory leaks.

fr4gles: IMO it makes a huge difference, calling methods on disposed object is not a good practice.

Please refer to https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-6.0#remarks

mxschmitt:
Thank you for your contribution! Closing since a proper fix has been implemented, see https://github.com/microsoft/playwright-dotnet/pull/1864 for more information.

PR #1823, opened 2021-11-08
chore: roll to ToT

JoelEinbinder:
Rolls to 1.17.0-next-1636303803000
cc72299cbc79187144b89d13ac117d702818c575﻿

### DIFF
--- a/src/Playwright/Core/Tracing.cs
+++ b/src/Playwright/Core/Tracing.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Playwright.Core
             await StartChunkAsync().ConfigureAwait(false);
         }
 
-        public Task StartChunkAsync() => _channel.StartChunkAsync();
+        public Task StartChunkAsync(TracingStartChunkOptions options = default) => _channel.StartChunkAsync(options?.Title);
 
         public async Task StopChunkAsync(TracingStopChunkOptions options = null)
         {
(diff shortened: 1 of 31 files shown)


src/Playwright/Transport/Channels/BrowserContextChannel.cs
```diff
@@ -259,8 +259,11 @@ internal Task TracingStopAsync()
                 Guid,
                 "tracingStop");
 
-        internal Task StartChunkAsync()
-            => Connection.SendMessageToServerAsync(Guid, "tracingStartChunk", null);
+        internal Task StartChunkAsync(string title = null)
```
pavelfeldman, line 262: This should be an option, not a parameter. It also breaks binary compatibility.

src/Playwright/Core/Tracing.cs
```diff
@@ -46,7 +46,7 @@ public async Task StartAsync(TracingStartOptions options = default)
             await StartChunkAsync().ConfigureAwait(false);
         }
 
-        public Task StartChunkAsync() => _channel.StartChunkAsync();
+        public Task StartChunkAsync(TracingStartChunkOptions options = default) => _channel.StartChunkAsync(options?.Title);
```
pavelfeldman, line 49: Commented in wrong place first. This breaks binary compatibility.

PR #1855, opened 2021-12-06
chore: add tests for browser connect and fix for browser close for co…

asambstack:
Added tests for browsertype connect.
Fixes https://github.com/microsoft/playwright-dotnet/issues/1784

Relates to https://github.com/microsoft/playwright-dotnet/pull/1602


src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,21 +80,54 @@ public void TearDown()
         }
 
         [PlaywrightTest("browsertype-connect.spec.ts", "should be able to reconnect to a browser")]
-        [Ignore("SKIP WIRE")]
-        public void ShouldBeAbleToConnectToBrowserAsync()
+        public async Task ShouldBeAbleToConnectToBrowserAsync()
```
pavelfeldman, line 83: For all these tests, you should take the original ("should be able to reconnect to a browser" in browsertype-connect.spec.ts for this one), and port them here. Your tests don't seem to be checking the same scenarios.

asambstack: Fixed the tests.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,21 +80,70 @@ public void TearDown()
...
+            Assert.AreEqual(result.ToString(), "121");
+            await browser2.CloseAsync();
         }
 
         [PlaywrightTest("browsertype-connect.spec.ts", "disconnected event should be emitted when browser is closed or server is closed")]
-        [Ignore("SKIP WIRE")]
-        public void DisconnectedEventShouldBeEmittedWhenBrowserIsClosedOrServerIsClosed()
+        public async Task DisconnectedEventShouldBeEmittedWhenBrowserIsClosedOrServerIsClosed()
+        {
+            var browserOne = await BrowserType.ConnectAsync(_browserServer.WSEndpoint);
```
pavelfeldman, line 129: This test does not seem to be on par with the one in https://github.com/microsoft/playwright/blob/main/tests/browsertype-connect.spec.ts.

We should port all the tests as is to ensure connect is flawless.

pavelfeldman:
In order to expose browserType.connect, we should have all the tests from https://github.com/microsoft/playwright/blob/main/tests/browsertype-connect.spec.ts ported and passing. This is substantial amount of code and reviewing it is time-consuming. You are saying "Fixed the tests.", but I can see that some of the tests are incomplete, while other tests are missing. I'm happy to review it again once the tests are in place.

asambstack:
Hey @pavelfeldman , thanks a lot for your feedback!
I've rewritten the tests and ported most of the specs from https://github.com/microsoft/playwright/blob/main/tests/browsertype-connect.spec.ts. The selector feature isn't available for `ConnectAsync` currently, hence haven't included the test for that. Will add that feature in a future PR for the selector support. Have made sure that the all the tests in dotnet are testing the same flow as that of NodeJS.
Looking forward to your review!

asambstack:
Hey @pavelfeldman 
Bumping this thread. Did you get a chance to review the tests?

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+        }
+
+        [PlaywrightTest("browsertype-connect.spec.ts", "should support slowmo option")]
+        public async Task ShouldSupportSlowMo()
         {
+            var browser = await BrowserType.ConnectAsync(_browserServer.WSEndpoint, new BrowserTypeConnectOptions { SlowMo = 200 });
+            var context = await browser.NewContextAsync();
+            var time1 = DateTime.Now;
+            var page = await context.NewPageAsync();
+            await page.GotoAsync(Server.EmptyPage);
```
pavelfeldman, line 143: I don't see these lines in the original test.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
-        public void ShouldHandleExceptionsDuringConnect()
+        [PlaywrightTest("browsertype-connect.spec.ts", "disconnected event should have browser as argument")]
+        public async Task DisconnectedEventShouldHaveBrowserAsArguments()
         {
+            var browser = await BrowserType.ConnectAsync(_browserServer.WSEndpoint);
+            IBrowser disconneced = null;
+            browser.Disconnected += (_, e) => disconneced = e;
+            await browser.CloseAsync();
+            Assert.AreEqual(browser, disconneced);
         }
```
pavelfeldman, line 181: "should handle exceptions during connect" test is missing

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
-        [Ignore("SKIP WIRE")]
-        public void ShouldRejectNavigationWhenBrowserCloses()
+        public async Task ShouldRejectNavigationWhenBrowserCloses()
         {
+            var browser = await BrowserType.ConnectAsync(_browserServer.WSEndpoint);
+            var page = await browser.NewPageAsync();
+            await browser.CloseAsync();
+
+            Assert.AreEqual(browser.IsConnected, false);
+            var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(async () => await page.GotoAsync(Server.EmptyPage));
```
pavelfeldman, line 222: This is not what the original test does.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+            var task = TaskUtils.WhenAll(page.WaitForRequestAsync(Server.EmptyPage), page.WaitForResponseAsync(Server.EmptyPage), browser.CloseAsync());
+            try
+            {
+                await task;
+            }
+            catch (Exception)
+            {
+                foreach (var exception in task.Exception.InnerExceptions)
+                {
+                    Console.WriteLine(exception.Message);
```
pavelfeldman, line 271: Remove debug logging

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+            var context = await browser.NewContextAsync();
+            var page = await context.NewPageAsync();
+            var task = TaskUtils.WhenAll(page.WaitForRequestAsync(Server.EmptyPage), page.WaitForResponseAsync(Server.EmptyPage), browser.CloseAsync());
+            try
+            {
+                await task;
+            }
+            catch (Exception)
+            {
+                foreach (var exception in task.Exception.InnerExceptions)
```
pavelfeldman, line 269: You test will pass (false positive) if there are no exceptions.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+                {
+                    Console.WriteLine(exception.Message);
+                    StringAssert.Contains("Page closed", exception.Message);
+                }
+            }
+        }
+
```
pavelfeldman, line 276: "should respect selectors" test is missing.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+        [PlaywrightTest("browsertype-connect.spec.ts", "should not throw on close after disconnect")]
+        public async Task ShouldNotThrowOnCloseAfterDisconnect()
+        {
+            var browser = await BrowserType.ConnectAsync(_browserServer.WSEndpoint);
+            var page = await browser.NewPageAsync();
+
+            _browserServer.Process.Kill();
+            await browser.CloseAsync();
+            while (browser.IsConnected)
```
pavelfeldman, line 285: style: here and below, can this be put in a form of a task as in the original test?

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+                page.ClickAsync("a"));
+
+            using var tmpDir = new TempDirectory();
+            string userPath = Path.Combine(tmpDir.Path, "download.txt");
+            var download = downloadTask.Result;
+            await download.DeleteAsync();
+            var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => download.SaveAsAsync(userPath));
+            StringAssert.Contains("Target page, context or browser has been closed", exception.Message);
+        }
+
```
pavelfeldman, line 442: "should properly disconnect when connection closes from the client side" test is missing
"should properly disconnect when connection closes from the server side" test is missing

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+            var page = await context.NewPageAsync();
+            await page.GotoAsync(Server.EmptyPage);
+            await context.CloseAsync();
+            await browser.CloseAsync();
+
+
+            Assert.That(harPath, Does.Exist);
+            var logString = System.IO.File.ReadAllText(harPath);
+            StringAssert.Contains(Server.EmptyPage, logString);
+        }
```
pavelfeldman, line 475: "should record trace with sources" test is missing

src/Playwright/Core/BrowserContext.cs
```diff
@@ -166,6 +168,11 @@ public async Task CloseAsync()
         {
             try
             {
+                if (IsClosed)
```
pavelfeldman, line 171: Rename this to IsClosedOrClosing, set it from Channel_Closed handler.

src/Playwright/Core/BrowserType.cs
```diff
@@ -149,8 +149,23 @@ public async Task<IBrowser> ConnectAsync(string wsEndpoint, BrowserTypeConnectOp
...
             }
+            var browser = playwright.PreLaunchedBrowser;
+            browser.IsRemote = true;
+            browser.IsConnectedOverWebSocket = true;
+            browser.Disconnected += (_, e) => CloseBrowser(transport, browser);
+            return browser;
+        }
 
-            return playwright.PreLaunchedBrowser;
+        private void CloseBrowser(WebSocketTransport transport, Browser browser)
```
pavelfeldman, line 159: Rename to be specific to the remote mode.

src/Playwright/Core/Frame.cs
```diff
@@ -195,6 +195,11 @@ public async Task WaitForLoadStateAsync(LoadState? state = default, FrameWaitFor
 
         public async Task<IResponse> WaitForNavigationAsync(FrameWaitForNavigationOptions options = default)
         {
+            if (Page.IsClosed)
```
pavelfeldman, line 198: If you go this path, you need to add 300 checks like this in all API methods. The point is that underlying transport error is bubbled every time the call is made after connection is closed.

src/Playwright/Core/Page.cs
```diff
@@ -449,6 +449,10 @@ public async Task CloseAsync(PageCloseOptions options = default)
         {
             try
             {
+                if (IsClosed)
```
pavelfeldman, line 452: Why is this check here? You can still call CloseAsync twice, even with this check.

pavelfeldman (commented):
The tests look much better now! The check in Page for isClosed is not good though.

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: The selectors functionality isn't available in `ConnectAsync` currently. Will add it in a follow up PR.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -80,71 +82,396 @@ public void TearDown()
...
+        [Skip(SkipAttribute.Targets.Windows)]
+        public async Task ShouldTimeoutInConnectWhileConnecting()
+        {
+            var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(async () => await BrowserType.ConnectAsync(_browserServer.WSEndpoint, new BrowserTypeConnectOptions { Timeout = 1 }));
+            StringAssert.Contains("Timeout 1ms exceeded", exception.Message);
+        }
+
+        [PlaywrightTest("browsertype-connect.spec.ts", "should support slowmo option")]
+        public async Task ShouldSupportSlowMo()
         {
```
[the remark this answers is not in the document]
asambstack: These tests require us to use a server to connect and record the headers. I wasn't able to use the existing "SimpleServer" for the tests. Will need to figure this out and update the server. Doing it in the follow up PR.

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: Have updated it.

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: Have fixed this test

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: Removed the logging.

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: Created a private function to close the client and server.

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: Added this test.

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: As discussed above, need to update the SimpleServer for these tests. Will create a separate PR for modifying "SimpleServer.cs" and writing the tests.

src/Playwright/Core/BrowserContext.cs
asambstack: Done

src/Playwright/Core/Page.cs
asambstack: If this check isn't put there, it'll throw when calling CloseAsync. Doing something similar to Playwright Java - https://github.com/microsoft/playwright-java/blob/main/playwright/src/main/java/com/microsoft/playwright/impl/PageImpl.java#L542

src/Playwright/Core/BrowserType.cs
asambstack: Renamed it.

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: Have changed it accordingly

src/Playwright.Tests/BrowserTypeConnectTests.cs
asambstack: This seems to be an implementational detail for NodeJS. Test isn't present on playwright java as well - https://github.com/microsoft/playwright-java/blob/main/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java#L211

asambstack:
Hey @pavelfeldman 
Did you get a chance to review these tests?

asambstack:
Hey @pavelfeldman 
Gentle bump

mxschmitt:
Merged and I'll follow up with #1908.

asambstack:
Thanks @mxschmitt !
So are we releasing support for connect in the next release?

PR #1857, opened 2021-12-06
chore: migrate build.ps1 to build.sh

mxschmitt:


build.sh
```diff
@@ -0,0 +1,88 @@
...
+}
+
+function roll_driver() {
+  new_driver_version="$1"
+  upstream_package_version=$(node -e "console.log(require('${upstream_repo_path}/package.json').version)")
+  echo "Rolling .NET driver to driver ${new_driver_version} and upstream version ${upstream_package_version}..."
+
+  xml_file_path="./src/Common/Version.props"
+  xml_file_contents=$(cat "${xml_file_path}")
+  xml_file_contents=$(echo "${xml_file_contents}" | sed "s/<DriverVersion>.*<\/DriverVersion>/<DriverVersion>${new_driver_version}<\/DriverVersion>/")
```
aslushnikov, line 58: nit: you can use any symbol for sed scripts to separate parts of expression.

For example:
```bash
sed "s/foo/bar" # default
sed "s|foo|bar" # variations
sed "s#foo#bar"
sed "s foo bar" # or even space would do!
```

aslushnikov (approved):

maw136:
IMHO this was a wrong decision. This made the script platform dependent.

mxschmitt:
> IMHO this was a wrong decision. This made the script platform dependent.

We expect to have mingw/cygwin (aka. Git bash) available on win systems, which you can use to get a bash and run the script from.

PR #1859, opened 2021-12-08
chore: merge helpers and test actual helper classes

mxschmitt:
Before that the tests were not testing the actual Helper class implementation which the users were using. They were testing the helpers located in `src/Playwright.Tests` instead of `src/Playwright`. This PR deletes unused helpers and makes the helpers public so we can test them, not sure if we can also test them without making them public.


dgozman (commented):
I don't like public helpers, let's explore possible solutions.

pavelfeldman:
The copy here was intentional, why is this changing?

pavelfeldman (approved):

PR #1861, opened 2021-12-08
test: unflake Firefox tests

mxschmitt:


src/Playwright.Tests/CapabilitiesTests.cs
```diff
@@ -81,7 +81,11 @@ public async Task ShouldRespectCSP()
         }
 
         [PlaywrightTest("capabilities.spec.ts", "should play video")]
-        [Skip(SkipAttribute.Targets.Webkit)]
+        [Skip(
```
mxschmitt, line 84: note: this is like we have upstream

dgozman (approved):

PR #1862, opened 2021-12-08
chore: rename wwwroot to assets

mxschmitt:
So its in sync with the other languages.


build.sh
```diff
@@ -43,7 +43,7 @@ function download_driver() {
 
 function update_assets() {
```
pavelfeldman, line 44: Let's remove calls to this from the csproj files to keep them simple

pavelfeldman: (same goes for other build commands, basically let's switch this to manual)

pavelfeldman (approved):

PR #1839, opened 2021-11-25
NET 6.0 Support

programatt:
- Updated TFM in projects from `.net5.0` to `.net6.0`
- Fixed CA warnings/Errors 

Builds on OSX-arm64 (M1 MacBookPro) but tons of the tests fail to run because the playwright builds of the browsers haven't been installed by `playwright install` yet. I can't run the dotnet global tool for playwright because it is expecting .NET 5 to be installed but I cannot seem to get it to recognize the recommended workaround for that on the M1 macs https://github.com/dotnet/sdk/issues/20584.

Update: After a hint from @martincostello I was able to run `playwright install` after setting `DOTNET_ROLL_FORWARD=Major` in my shell. I now get one test failure.

Update 2: Now the failing tests are passing, one job appears to have timed out due to taking too long. But the API Integrity check job is failing. I'm not sure if this is due to something I did or if a maintainer needs to step in. 

 I also had to manually run `cp -R ./playwright/tests/assets/ ./src/Playwright.Tests.TestServer/wwwroot/` as the powershell script didn't error, but it also didn't copy all the files over like I expected. Maybe a difference with powershell on OSX vs windows?

Will try and run all the tests on windows-x64 and linux as well.

PR #1884, opened 2021-12-14
chore migrate project over to .NET 6

mxschmitt:
Relates https://github.com/microsoft/playwright-dotnet/pull/1839

### DIFF
--- a/src/Playwright.Tests/NetworkPostDataTests.cs
+++ b/src/Playwright.Tests/NetworkPostDataTests.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Playwright.Tests
 
             await Task.WhenAll(task, actualTask);
 
-            string expectedJsonValue = JsonSerializer.Serialize(value, new()
+            string expectedJsonValue = JsonSerializer.Serialize(value, new JsonSerializerOptions()
             {
                 Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
                 WriteIndented = true
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -136,9 +136,7 @@ namespace Microsoft.Playwright.Transport
                         object obj = propertyDescriptor.GetValue(args);
                         if (obj != null)
                         {
-#pragma warning disable CA1845 // Use span-based 'string.Concat' and 'AsSpan' instead
                             string name = propertyDescriptor.Name.Substring(0, 1).ToLower() + propertyDescriptor.Name.Substring(1);
-#pragma warning restore CA1845 // Use span-based 'string.Concat' and 'AsSpan' instead
                             sanitizedArgs.Add(name, obj);
                         }
                     }
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -136,7 +136,9 @@ namespace Microsoft.Playwright.Transport
                         object obj = propertyDescriptor.GetValue(args);
                         if (obj != null)
                         {
+#pragma warning disable CA1845 // Use span-based 'string.Concat' and 'AsSpan' instead
                             string name = propertyDescriptor.Name.Substring(0, 1).ToLower() + propertyDescriptor.Name.Substring(1);
+#pragma warning restore CA1845 // Use span-based 'string.Concat' and 'AsSpan' instead
                             sanitizedArgs.Add(name, obj);
                         }
                     }
--- a/src/Playwright/Transport/WebSocketTransport.cs
+++ b/src/Playwright/Transport/WebSocketTransport.cs
@@ -104,13 +104,17 @@ namespace Microsoft.Playwright.Transport
                     if (_webSocket.State != WebSocketState.Open)
                         break;
 
-                    memoryStream.Write(buffer.Array, 0, result.Count);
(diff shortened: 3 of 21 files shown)


src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -104,13 +104,17 @@ private async Task DispatchIncomingMessagesAsync()
...
 
-                    memoryStream.Write(buffer.Array, 0, result.Count);
+#pragma warning disable CA1835 // Change the 'WriteAsync' method call to use the 'Stream.WriteAsync(ReadOnlyMemory<byte>, CancellationToken)' overload
+                    await memoryStream.WriteAsync(buffer.Array, 0, result.Count).ConfigureAwait(false);
+#pragma warning restore CA1835 // Change the 'WriteAsync' method call to use the 'Stream.WriteAsync(ReadOnlyMemory<byte>, CancellationToken)' overload
 
                     if (result.EndOfMessage)
                     {
                         string output = Encoding.UTF8.GetString(memoryStream.ToArray());
                         MessageReceived?.Invoke(this, new MessageReceivedEventArgs(output));
```
mxschmitt, line 114: didn't fix the linter suggestions - since we eventually get rid of the WebSocketTrnasport.cs

src/Playwright.Tests/NetworkPostDataTests.cs
```diff
@@ -52,7 +52,7 @@ public async Task ShouldReturnCorrectPostdataBufferForUtf8Body()
 
             await Task.WhenAll(task, actualTask);
 
-            string expectedJsonValue = JsonSerializer.Serialize(value, new()
+            string expectedJsonValue = JsonSerializer.Serialize(value, new JsonSerializerOptions()
```
kblok, line 55: Why are you bringing this back? Same here and in other places.

src/Playwright/Transport/Connection.cs
```diff
@@ -136,7 +136,9 @@ public void Dispose()
                         object obj = propertyDescriptor.GetValue(args);
                         if (obj != null)
                         {
+#pragma warning disable CA1845 // Use span-based 'string.Concat' and 'AsSpan' instead
```
kblok, line 139: Are we adding this because the warning is .NET 6 only?
If so I would ignore it in the ruleset.

src/Playwright/Transport/WebSocketTransport.cs
```diff
@@ -104,13 +104,17 @@ private async Task DispatchIncomingMessagesAsync()
                     if (_webSocket.State != WebSocketState.Open)
                         break;
 
-                    memoryStream.Write(buffer.Array, 0, result.Count);
+#pragma warning disable CA1835 // Change the 'WriteAsync' method call to use the 'Stream.WriteAsync(ReadOnlyMemory<byte>, CancellationToken)' overload
```
kblok, line 107: Same, if it's due to many versions, let's add it to the ruleset.

src/Playwright.Tests/NetworkPostDataTests.cs
mxschmitt: the new() syntax fails because of a new overload to Serialize with 2 arguments.

src/Playwright/Transport/WebSocketTransport.cs
mxschmitt: (we eventually get rid of this file soon, so I'd keep it for now)

src/Playwright/Transport/Connection.cs
mxschmitt: added yeah.

kblok:
My two cents here. .NET 6 is not only the new LTS version (.NET 5 is not LTS)
.NET 6 is also the only version supported for MacOS Arm64 users. This will unlock many developers enjoying the pain of this new chip.

https://dotnet.microsoft.com/en-us/download/dotnet
https://dotnet.microsoft.com/en-us/download/dotnet/6.0

campersau:
Looks like this needs also be updated to `.NET 6`:
https://github.com/microsoft/playwright-dotnet/blob/83af532735a11041183d9c6028e287f92bf9c584/.github/workflows/tests.yml#L15

campersau:
These methods
https://github.com/microsoft/playwright-dotnet/blob/83af532735a11041183d9c6028e287f92bf9c584/src/Playwright/Helpers/JsonExtensions.cs#L50-L51
https://github.com/microsoft/playwright-dotnet/blob/83af532735a11041183d9c6028e287f92bf9c584/src/Playwright/Helpers/JsonExtensions.cs#L60-L61
should call the new overloads on net6.0 which are accepting a `JsonElement` directly:
https://docs.microsoft.com/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-6.0#System_Text_Json_JsonSerializer_Deserialize__1_System_Text_Json_JsonElement_System_Text_Json_JsonSerializerOptions_
https://docs.microsoft.com/de-de/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-6.0#System_Text_Json_JsonSerializer_Deserialize_System_Text_Json_JsonElement_System_Type_System_Text_Json_Serialization_JsonSerializerContext_

mxschmitt:
Closing for now. We extracted some patches out of it and will do it in the next few months.

Issue #1892, opened 2021-12-15
[Bug]: Microsoft.Playwright.CLI version 1.2.1 fails on install


AutioL25187-pki:
### Playwright version

Microsoft.Playwright.CLI 1.2.1

### Operating system

Windows, Linux

### Other information

windows-latest and ubuntu-latest on GitHub Actions.

### What happened? / Describe the bug

Microsoft.Playwright.CLI version 1.2.1 installation fails on GitHub Actions. Installation of version 1.2.0 succeeds.

This command fails (installs 1.2.1): `dotnet tool install -g Microsoft.Playwright.CLI`
This command works: `dotnet tool install -g Microsoft.Playwright.CLI --version 1.2.0`

### Code snippet to reproduce your bug

```shell
dotnet tool install -g Microsoft.Playwright.CLI
playwright install
```

### Relevant log output

```shell
dotnet tool install -g Microsoft.Playwright.CLI
  playwright install
  playwright --version
  shell: /usr/bin/bash -e {0}
You can invoke the tool using the following command: playwright
Tool 'microsoft.playwright.cli' (version '1.2.1') was successfully installed.
Unhandled exception. Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: Cannot perform runtime binding on a null reference
   at CallSite.Target(Closure , CallSite , Object , String[] )
   at System.Dynamic.UpdateDelegates.UpdateAndExecuteVoid2[T0,T1](CallSite site, T0 arg0, T1 arg1)
   at Microsoft.Playwright.CLI.Program.Main(String[] args) in /_/src/Playwright.CLI/Program.cs:line 83
/home/runner/work/_temp/f17a355f-5cad-464e-8c78-7743772a3bf7.sh: line 2:  1773 Aborted                 (core dumped) playwright install
Error: Process completed with exit code 134.
```

luiscantero:
From what I can see, the browser install changed for .NET 6 to this:
```
pwsh bin\Debug\net6.0\playwright.ps1 install
```

[Reference](https://playwright.dev/dotnet/docs/intro)

mxschmitt:
We published 1.2.2 of `Microsoft.Playwright.CLI` which should fix the issues.

PR #1895, opened 2021-12-15
chore: use Join-Path in PS1 script

martincostello:
Use [`Join-Path`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/join-path?view=powershell-7.2) so that the correct path delimiter is used for the executing platform, rather than concatenating two strings and making assumptions.


mxschmitt:
Out of curiosity, is this fixing a bug or just "code-health"?

martincostello:
Just the latter - I looked at the script after seeing it mentioned in [#1892](https://github.com/microsoft/playwright-dotnet/issues/1892#issuecomment-994771630) and it set off my Spidey-Sense.

pavelfeldman (approved):

mxschmitt (approved):

PR #1914, opened 2022-01-06
chore: enhance websocket test-server capabilities

mxschmitt:


src/Playwright.Tests/BaseTests/PlaywrightAssert.cs
```diff
@@ -39,7 +39,7 @@ internal static class PlaywrightAssert
             try
             {
                 await action();
-                Assert.Fail();
+                Assert.Fail($"Expected exception of type '{typeof(T).Name}' which was not thrown");
```
mxschmitt, line 42: drive-by: better exception if the exception does not get thrown

src/Playwright/Transport/Channels/BrowserContextChannel.cs
```diff
@@ -165,15 +165,6 @@ internal Task SetNetworkInterceptionEnabledAsync(bool enabled)
                     ["enabled"] = enabled,
                 });
 
-        internal Task SetHttpCredentialsAsync(HttpCredentials credentials)
```
mxschmitt, line 168: drive-by: not used

pavelfeldman (approved):

PR #1935, opened 2022-01-14
create Route glob regex immediately

campersau:
Currently route urls are converted to regular expressions on every route match which could result in a regex parse exception if the url is invalid.

With this change the regex is created once when the url route is added which means that potential regex parse exceptions are thrown immediately instead of later when the route is matched.

This also fixes a case where it wasn't possible to remove a url route when the browsing context has a base url set.

This also synchronized the code style for adding routes in `BrowserContext` and `Page`.

### DIFF
--- a/src/Playwright.Tests/BrowserContextRouteTests.cs
+++ b/src/Playwright.Tests/BrowserContextRouteTests.cs
@@ -113,6 +113,38 @@ namespace Microsoft.Playwright.Tests
             Assert.AreEqual(new List<int>() { 1 }, intercepted);
         }
 
+        [PlaywrightTest]
+        public async Task ShouldUnroutePageWithBaseUrl()
+        {
+            var options = new BrowserNewContextOptions();
+            options.BaseURL = Server.Prefix;
+
+            await using var context = await Browser.NewContextAsync(options);
+            var page = await context.NewPageAsync();
+            var intercepted = new List<int>();
+
+            await page.RouteAsync("/empty.html", (route) =>
+            {
+                intercepted.Add(1);
+                route.ContinueAsync();
+            });
+
+            Action<IRoute> handler2 = (route) =>
+            {
+                intercepted.Add(2);
+                route.ContinueAsync();
+            };
+            await page.RouteAsync("/empty.html", handler2);
+
+            await page.GotoAsync(Server.EmptyPage);
+            Assert.AreEqual(new List<int>() { 2 }, intercepted);
+
+            intercepted.Clear();
+            await page.UnrouteAsync("/empty.html", handler2);
+            await page.GotoAsync(Server.EmptyPage);
+            Assert.AreEqual(new List<int>() { 1 }, intercepted);
+        }
+
         [PlaywrightTest("browsercontext-route.spec.ts", "should yield to page.route")]
         public async Task ShouldYieldToPageRoute()
         {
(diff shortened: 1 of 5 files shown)


src/Playwright/Core/Page.cs
```diff
@@ -982,9 +949,18 @@ internal void OnFrameNavigated(Frame frame)
...
+                Regex = urlRegex,
+                Url = urlString,
+                Function = urlFunc,
+                Handler = handler,
+                Times = options?.Times,
+            });
+
         private Task RouteAsync(RouteSetting setting)
         {
-            setting.Url = Context.CombineUrlWithBase(setting.Url);
```
campersau, line 987: > This also fixes a case where it wasn't possible to remove a url route when the browsing context has a base url set.

src/Playwright.Tests/BrowserContextRouteTests.cs
```diff
@@ -150,5 +182,25 @@ public async Task ShouldFallBackToContextRoute()
             var response = await page.GotoAsync(Server.EmptyPage);
             Assert.AreEqual("context", await response.TextAsync());
         }
+
+        [PlaywrightTest("browsercontext-route.spec.ts", "should throw on invalid route url")]
```
pavelfeldman, line 186: nit: as this test is specific to C#, you don't need to annotate it with the link to the upstream test.

src/Playwright.Tests/BrowserContextRouteTests.cs
```diff
@@ -113,6 +113,38 @@ public async Task ShouldUnroute()
             Assert.AreEqual(new List<int>() { 1 }, intercepted);
         }
 
+        [PlaywrightTest("browsercontext-route.spec.ts", "should unroute page with base URL")]
```
pavelfeldman, line 116: nit: as this test is specific to C#, you don't need to annotate it with the link to the upstream test.

src/Playwright.Tests/PageRouteTests.cs
```diff
@@ -691,5 +691,23 @@ public async Task ShouldSupportCorsWithDifferentMethods()
 
             Assert.AreEqual(new[] { "DELETE", "electric", "cars" }, resp);
         }
+
+        [PlaywrightTest("page-route.spec.ts", "should throw on invalid route url")]
```
pavelfeldman, line 695: nit: as this test is specific to C#, you don't need to annotate it with the link to the upstream test.

src/Playwright/Core/BrowserContext.cs
```diff
@@ -245,7 +245,7 @@ public async Task<IPage> NewPageAsync()
         }
 
         public Task RouteAsync(string url, Action<IRoute> handler, BrowserContextRouteOptions options = default)
-            => RouteAsync(url, null, null, handler, options);
+            => RouteAsync(url, new Regex(CombineUrlWithBase(url).GlobToRegex()), null, handler, options);
```
pavelfeldman, line 248: Before the change, `RouteAsync` supported either glob (first argument), regex (second argument) or a callback (third argument). You are folding the glob one into regex, which is great. With that, the first argument (glob) should go away (line :364 new code). Along with the `RouteSetting`'s `Url` field.

campersau: Glob/Url ist still needed so the user can `Unroute` it with the same glob/url.

pavelfeldman: Not a C# expert, but I was hoping there was a way to compare those for equality. If not, we could compare underlying patterns + flags?

campersau: Ah you mean that in `UnrouteAsync` a regex is also created from the glob and checked for equality with the route regex?
Then no the regex objects itself won't be equal but their `ToString` representations are:

```c#
var url = @"^http:\/\/localhost:8909\/empty\.html$";
var regex1 = new Regex(url);
var regex2 = new Regex(url);

Console.WriteLine(regex1 == regex2); // false
Console.WriteLine(regex1.Equals(regex2)); // false
Console.WriteLine(regex1.ToString() == regex2.ToString() && regex1.Options == regex2.Options); // true
```
With that we could get rid of the `Url` property, which I have done in https://github.com/microsoft/playwright-dotnet/pull/1935/commits/300bcc2479ea1e6e40505d25da41ea0d132891d9

PR #1938, opened 2022-01-16
chore: align network route handling with upstream

### LINKED ISSUE #1800: [Feature] port over the web-first assertions
We have web first assertions in JavaScript/Java which should get ported over to .NET. Investigate integrations with NUnit and others.

mxschmitt:
Changes:
- align network handling with upstream
- drive-by: https://github.com/microsoft/playwright/commit/3cc839e01327dbf5b6e2a889da8da373b7efd503

### DIFF
--- a/src/Playwright/Core/BrowserContext.cs
+++ b/src/Playwright/Core/BrowserContext.cs
@@ -44,7 +44,7 @@ namespace Microsoft.Playwright.Core
         private readonly Dictionary<string, Delegate> _bindings = new();
         private readonly BrowserContextInitializer _initializer;
         private readonly ITracing _tracing;
-        private List<RouteSetting> _routes = new();
+        private List<RouteHandler> _routes = new();
 
         private float? _defaultNavigationTimeout;
         private float? _defaultTimeout;
@@ -247,13 +247,13 @@ namespace Microsoft.Playwright.Core
         }
 
         public Task RouteAsync(string url, Action<IRoute> handler, BrowserContextRouteOptions options = default)
-            => RouteAsync(new Regex(CombineUrlWithBase(url).GlobToRegex()), null, handler, options);
+            => RouteAsync(url, null, null, handler, options);
 
-        public Task RouteAsync(Regex url, Action<IRoute> handler, BrowserContextRouteOptions options = default)
-            => RouteAsync(url, null, handler, options);
+        public Task RouteAsync(Regex regex, Action<IRoute> handler, BrowserContextRouteOptions options = default)
+            => RouteAsync(null, regex, null, handler, options);
 
-        public Task RouteAsync(Func<string, bool> url, Action<IRoute> handler, BrowserContextRouteOptions options = default)
-            => RouteAsync(null, url, handler, options);
+        public Task RouteAsync(Func<string, bool> func, Action<IRoute> handler, BrowserContextRouteOptions options = default)
+            => RouteAsync(null, null, func, handler, options);
 
         public Task SetExtraHTTPHeadersAsync(IEnumerable<KeyValuePair<string, string>> headers)
             => Channel.SetExtraHTTPHeadersAsync(headers);
@@ -279,13 +279,13 @@ namespace Microsoft.Playwright.Core
         }
 
         public Task UnrouteAsync(string urlString, Action<IRoute> handler = default)
-            => UnrouteAsync(new Regex(CombineUrlWithBase(urlString).GlobToRegex()), null, handler);
+            => UnrouteAsync(urlString, null, null, handler);
 
         public Task UnrouteAsync(Regex urlRegex, Action<IRoute> handler = default)
-            => UnrouteAsync(urlRegex, null, handler);
+            => UnrouteAsync(null, urlRegex, null, handler);
 
         public Task UnrouteAsync(Func<string, bool> urlFunc, Action<IRoute> handler = default)
-            => UnrouteAsync(null, urlFunc, handler);
+            => UnrouteAsync(null, null, urlFunc, handler);
 
         public async Task<T> InnerWaitForEventAsync<T>(PlaywrightEvent<T> playwrightEvent, Func<Task> action = default, Func<T, bool> predicate = default, float? timeout = default)
         {
@@ -326,55 +326,41 @@ namespace Microsoft.Playwright.Core
 
         internal void OnRoute(Route route, IRequest request)
         {
-            foreach (var item in _routes)
+            foreach (var routeHandler in _routes.ToList())
             {
-                if (
-                    ((item.Times ?? 0) == 0 || item.HandledCount >= item.Times) &&
-                    ((item.Regex?.IsMatch(request.Url) == true) ||
-                    (item.Function?.Invoke(request.Url) == true)))
+                if (routeHandler.Matches(request.Url))
                 {
-                    item.Handle(route);
+                    if (routeHandler.Handle(route))
+                    {
+                        _routes.Remove(routeHandler);
+                        if (_routes.Count == 0)
+                        {
+                            DisableInterceptionAsync().ConfigureAwait(false);
+                        }
+                    }
                     return;
                 }
             }
-
-            _ = route.ContinueAsync(new());
+            route.InternalContinueAsync(new()).ConfigureAwait(false);
         }
 
-        internal bool UrlMatches(string url, string glob)
-           => new Regex(CombineUrlWithBase(glob).GlobToRegex()).Match(url).Success;
-
-        internal string CombineUrlWithBase(string url)
-        {
-            var baseUrl = Options?.BaseURL;
-            if (string.IsNullOrEmpty(baseUrl)
-                || (url?.StartsWith("*", StringComparison.InvariantCultureIgnoreCase) ?? false)
-                || !Uri.IsWellFormedUriString(url, UriKind.RelativeOrAbsolute))
-            {
-                return url;
-            }
-
-            var mUri = new Uri(url, UriKind.RelativeOrAbsolute);
-            if (!mUri.IsAbsoluteUri)
-            {
-                return new Uri(new Uri(baseUrl), mUri).ToString();
-            }
-
-            return url;
-        }
-
-        private Task RouteAsync(Regex urlRegex, Func<string, bool> urlFunc, Action<IRoute> handler, BrowserContextRouteOptions options)
+        private Task RouteAsync(string urlString, Regex urlRegex, Func<string, bool> urlFunc, Action<IRoute> handler, BrowserContextRouteOptions options)
             => RouteAsync(new()
             {
+                Url = urlString,
                 Regex = urlRegex,
                 Function = urlFunc,
                 Handler = handler,
                 Times = options?.Times,
+                BaseURL = this.Options?.BaseURL,
             });
 
-        private Task RouteAsync(RouteSetting setting)
+        private Task RouteAsync(RouteHandler route)
         {
-            _routes.Insert(0, setting);
+            // Make sure that the given route match can be executed (e.g. invalid pattern)
+            route.Matches("http://127.0.0.1");
+
+            _routes.Insert(0, route);
 
             if (_routes.Count == 1)
             {
@@ -384,29 +370,32 @@ namespace Microsoft.Playwright.Core
             return Task.CompletedTask;
         }
 
-        private Task UnrouteAsync(Regex urlRegex, Func<string, bool> urlFunc, Action<IRoute> handler = null)
+        private Task UnrouteAsync(string urlString, Regex urlRegex, Func<string, bool> urlFunc, Action<IRoute> handler = null)
             => UnrouteAsync(new()
             {
+                Url = urlString,
                 Function = urlFunc,
                 Regex = urlRegex,
                 Handler = handler,
             });
 
-        private Task UnrouteAsync(RouteSetting setting)
+        private Task UnrouteAsync(RouteHandler route)
         {
-            var newRoutesList = new List<RouteSetting>();
-            newRoutesList.AddRange(_routes.Where(r =>
-                (setting.Regex != null && !(r.Regex == setting.Regex || (r.Regex.ToString() == setting.Regex.ToString() && r.Regex.Options == setting.Regex.Options))) ||
-                (setting.Function != null && r.Function != setting.Function) ||
-                (setting.Handler != null && r.Handler != setting.Handler)));
-            _routes = newRoutesList;
+            _routes = _routes.Where(r => (route.Url != null && r.Url != route.Url) || (route.Regex != null && r.Regex != route.Regex) ||
+                (route.Function != null && r.Function != route.Function) ||
+                (route.Handler != null && r.Handler != route.Handler)).ToList();
 
             if (_routes.Count == 0)
             {
-                return Channel.SetNetworkInterceptionEnabledAsync(false);
+                return DisableInterceptionAsync();
             }
 
             return Task.CompletedTask;
+        }
+
+        internal async Task DisableInterceptionAsync()
+        {
+            await Channel.SetNetworkInterceptionEnabledAsync(false).ConfigureAwait(false);
         }
 
         internal void OnClose()
--- a/src/Playwright/Core/Page.cs
+++ b/src/Playwright/Core/Page.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Playwright.Core
         private readonly ITouchscreen _touchscreen;
         private readonly PageInitializer _initializer;
 
-        private List<RouteSetting> _routes = new();
+        private List<RouteHandler> _routes = new();
         private EventHandler<IFileChooser> _fileChooserEventHandler;
         private bool _fileChooserIntercepted;
         private Video _video;
@@ -81,7 +81,7 @@ namespace Microsoft.Playwright.Core
             _channel.Popup += (_, e) => Popup?.Invoke(this, e.Page);
             _channel.WebSocket += (_, e) => WebSocket?.Invoke(this, e);
             _channel.BindingCall += Channel_BindingCall;
-            _channel.Route += Channel_Route;
+            _channel.Route += (_, e) => OnRoute(e.Route, e.Request);
             _channel.FrameAttached += Channel_FrameAttached;
             _channel.FrameDetached += Channel_FrameDetached;
             _channel.Dialog += (_, e) =>
@@ -120,6 +120,9 @@ namespace Microsoft.Playwright.Core
             _defaultNavigationTimeout = Context.DefaultNavigationTimeout;
             _defaultTimeout = Context.DefaultTimeout;
             _initializer = initializer;
+
+            Close += (_, _) => ClosedOrCrashedTcs.TrySetResult(true);
+            Crash += (_, _) => ClosedOrCrashedTcs.TrySetResult(true);
         }
 
         public event EventHandler<IConsoleMessage> Console;
@@ -273,10 +276,12 @@ namespace Microsoft.Playwright.Core
             }
         }
 
+        internal TaskCompletionSource<bool> ClosedOrCrashedTcs { get; } = new();
+
         public IFrame Frame(string name)
             => Frames.FirstOrDefault(f => f.Name == name);
 
-        public IFrame FrameByUrl(string urlString) => Frames.FirstOrDefault(f => Context.UrlMatches(urlString, f.Url));
+        public IFrame FrameByUrl(string urlString) => Frames.FirstOrDefault(f => Helpers.UrlHelpers.UrlMatches(Context.Options.BaseURL, urlString, f.Url));
 
         public IFrame FrameByUrl(Regex urlRegex) => Frames.FirstOrDefault(f => urlRegex.IsMatch(f.Url));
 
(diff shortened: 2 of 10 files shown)


src/Playwright/Core/Page.cs
```diff
@@ -982,9 +992,9 @@ internal void OnFrameNavigated(Frame frame)
 
         internal void FireResponse(IResponse response) => Response?.Invoke(this, response);
 
-        private Task RouteAsync(RouteSetting setting)
+        private Task RouteAsync(RouteHandler setting)
         {
-            setting.Url = Context.CombineUrlWithBase(setting.Url);
+            setting.Url = Helpers.UrlHelpers.ConstructURLBasedOnBaseURL(Context.Options.BaseURL, setting.Url);
```
campersau, line 997: I am not sure if this line is correct.

```suggestion
```

1. `BrowserContext` does not do it:
https://github.com/microsoft/playwright-dotnet/blob/ac79b4bb5e26be3e4fd33b7a713f2e42ce0e01b6/src/Playwright/Core/BrowserContext.cs#L357-L359
2. It is already called here but with a check for leading `*`:
https://github.com/microsoft/playwright-dotnet/pull/1938/files#diff-9eead3ec1fcbb4945a0bd0376c53196ce1533db4cb6c438509f9cca7847997e3R37
3. It does not work with `Unroute` in some cases where no `*` is used:
https://github.com/microsoft/playwright-dotnet/pull/1935#discussion_r785074031
https://github.com/microsoft/playwright-dotnet/pull/1935/files#diff-639d841671d8aafa409d45a0d8fab0834d702fb8a2ceaf8625970b8a8950812cR116-R146

src/Playwright/Helpers/UrlHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+    internal static class UrlHelpers
+    {
+        internal static bool UrlMatches(string baseURL, string urlString, string matchString, Regex matchRegex = null, Func<string, bool> matchFunction = null)
+        {
+            if (string.IsNullOrEmpty(matchString) && matchRegex == null && matchFunction == null)
+                return true;
+            if (!string.IsNullOrEmpty(matchString) && !matchString.StartsWith("*", System.StringComparison.Ordinal))
+                matchString = ConstructURLBasedOnBaseURL(baseURL, matchString);
+            if (!string.IsNullOrEmpty(matchString))
+                matchRegex = new Regex(matchString.GlobToRegex());
```
campersau, line 39: It would be nice if this would only be done once when the `RouteHandler` gets created. See https://github.com/microsoft/playwright-dotnet/pull/1935

src/Playwright/Helpers/UrlHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (matchRegex != null)
+                return matchRegex.IsMatch(urlString);
+            if (!string.IsNullOrEmpty(matchString) && matchString == urlString)
+                return true;
+            var url = new Uri(urlString);
+            if (url == null)
+                return false;
```
campersau, line 46: I don't think that this check would ever be `true`.
```suggestion
```

Also it would be nice the `Uri` is only created inside of `if (!string.IsNullOrEmpty(matchString))` (the next line) where it gets used.

src/Playwright/Core/BrowserContext.cs
```diff
@@ -326,42 +326,22 @@ public Task<IPage> RunAndWaitForPageAsync(Func<Task> action, BrowserContextRunAn
...
-                    (item.Function?.Invoke(request.Url) == true)))
+                if (routeHandler.Matches(request.Url))
                 {
-                    item.Handle(route);
+                    if (routeHandler.Handle(route))
+                    {
+                        _routes.Remove(routeHandler);
+                        if (_routes.Count == 0)
+                        {
+                            DisableInterceptionAsync().ConfigureAwait(false);
```
pavelfeldman, line 338:
```suggestion
                            _ = DisableInterceptionAsync().ConfigureAwait(false);
```

src/Playwright/Core/BrowserContext.cs
```diff
@@ -326,42 +326,22 @@ public Task<IPage> RunAndWaitForPageAsync(Func<Task> action, BrowserContextRunAn
...
-            }
-
-            var mUri = new Uri(url, UriKind.RelativeOrAbsolute);
-            if (!mUri.IsAbsoluteUri)
-            {
-                return new Uri(new Uri(baseUrl), mUri).ToString();
-            }
-
-            return url;
+            route.InternalContinueAsync(new()).ConfigureAwait(false);
```
pavelfeldman, line 344:
```suggestion
            _ = route.InternalContinueAsync(new()).ConfigureAwait(false);
```

src/Playwright/Helpers/UrlHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Playwright.Helpers
+{
+    internal static class UrlHelpers
+    {
+        internal static bool UrlMatches(string baseURL, string urlString, string matchString, Regex matchRegex = null, Func<string, bool> matchFunction = null)
```
pavelfeldman, line 32: I'd be in favor of resolving relative to the baseURL on the call site. With the separate glob/regex change, matchString will go away anyways, so you'll have to do it on the call site.

mxschmitt: we can follow-up and then make it the same upstream too, wdyt?

src/Playwright/Helpers/UrlHelpers.cs
mxschmitt: ack

src/Playwright/Helpers/UrlHelpers.cs
mxschmitt: we can do it in a follow-up

src/Playwright/Core/Page.cs
mxschmitt: very good find, this should definitely not be there.

src/Playwright/Helpers/UrlHelpers.cs
campersau: Sure, I can do it in #1935 when this gets merged first.

src/Playwright/Core/Route.cs
```diff
@@ -69,15 +69,36 @@ public Task FulfillAsync(RouteFulfillOptions options = default)
...
+        }
+
+        private async Task RaceWithPageCloseAsync(Task task)
+        {
+            var page = (Page)Request.Frame.Page;
+
+            // When page closes or crashes, we catch any potential rejects from this Route.
+            // Note that page could be missing when routing popup's initial request that
+            // does not have a Page initialized just yet.
+            await Task.WhenAny(task, page.ClosedOrCrashedTcs.Task).ConfigureAwait(false);
```
campersau, line 101: `Task.WhenAny` is dangerous when not all tasks it takes are observed (e.g. accessing the `task.Exception` property when an exception occurs) which is the case here.
For example add the following code at the start of the `ShouldNotThrowWhenContinuingWhilePageIsClosing`, `ShouldNotThrowWhenContinuingAfterPageIsClosed` tests and set breakpoints in there:
```c#
TaskScheduler.UnobservedTaskException += (s, e) =>
{
    // set breakpoint here
};
```
Debug these tests `ShouldNotThrowWhenContinuingAfterPageIsClosed`, `ShouldNotThrowWhenContinuingWhilePageIsClosing` together and the `UnobservedTaskException` will be called because the `task.Exception`  property is never observed.

image


I had this issue in my code as well and solved it like this, which will observe all `PlaywrightException`s:
```suggestion
            _ = task.ContinueWith(t => t.Exception.Handle(e => e is PlaywrightException), System.Threading.CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Current);
            await Task.WhenAny(task, page.ClosedOrCrashedTcs.Task).ConfigureAwait(false);
```

campersau: After looking at the code in more detail all the current callsites are already "racing" with the close event.
1. They might get an exception set here like `Target page, context or browser has been closed`:
https://github.com/microsoft/playwright-dotnet/blob/914c8fa6b3de7f5a62e76e059b10d18604064b01/src/Playwright/Transport/Connection.cs#L224-L227

2. And if not they will always get set an exception here:
https://github.com/microsoft/playwright-dotnet/blob/914c8fa6b3de7f5a62e76e059b10d18604064b01/src/Playwright/Transport/Connection.cs#L360-L363

Or is the intention to not throw when the page gets closed? If so then it needs to be suppress like I have shown above or this would also work:

```suggestion
            if (task != await Task.WhenAny(task, page.ClosedOrCrashedTcs.Task).ConfigureAwait(false))
            {
                try
                {
                    // this will resolve (almost) immediately as the exception is already set by one of the two methods mentioned above.
                    await task.ConfigureAwait(false); 
                }
                catch
                {
                    // ignore
                }
            }
```




On node the `process.on('unhandledRejection', function() {})` should also have been called I guess.

mxschmitt: What happens if we don't do it? Our assumption is that modern targets don't crash on unobserved exceptions.

campersau: Yes, they don't crash anymore since `net45` but `TaskScheduler.UnobservedTaskException` is still raised, which might be surprising for users and likely create issues and they can't do anything about that.
https://docs.microsoft.com/dotnet/api/system.threading.tasks.taskscheduler.unobservedtaskexception?view=net-6.0

---

I ran into this the last week as I use `CancellationToken`s instead of timeouts so my code looked something like this:

```c#
var abortTask = Task.Delay(Timeout.Infinite, cancellationToken);
var page = await browserContext.NewPageAsync();
try
{
    var gotoTask = page.GotoAsync("...", new PageGotoOptions { Timeout = 0 });
    if (abortTask == await Task.WhenAny(gotoTask, abortTask))
    {
        cancellationToken.ThrowIfCancellationRequested();
    }
    await gotoTask;

    // ...
}
finally
{
    await page.CloseAsync();
}
```

Then I got `TaskScheduler.UnobservedTaskException` events because the `gotoTask` was unobserved when the cancellation token was canceled. And as we log all `UnobservedTaskException`s it created a lot of unnecessary noise.
So I added a extension method in which I ignored any `PlaywrightException`s with the message `Target closed`:

```c#
var abortTask = Task.Delay(Timeout.Infinite, cancellationToken);
var page = await browserContext.NewPageAsync();
try
{
    var gotoTask = page.GotoAsync("...", new PageGotoOptions { Timeout = 0 }).IgnorePlaywrightTargetClosedException();
    if (abortTask == await Task.WhenAny(gotoTask, abortTask))
    {
        cancellationToken.ThrowIfCancellationRequested();
    }
    await gotoTask;

    // ...
}
finally
{
    await page.CloseAsync();
}

static Task<T> IgnorePlaywrightTargetClosedException<T>(this Task<T> task)
{
    task.ContinueWith(t => t.Exception.Handle(e => e is PlaywrightException && e.Message.Contains("Target closed")));
    return task;
}
```

mxschmitt: @ campersau it would make sense to ensure in all our nunit tests that no such UnobservedTaskException expected are emitted. What do you think?

campersau: Yes, that might be a good strategy to find `UnobservedTaskException`s in the product code. The problem is that in order to know which `UnobservedTaskException` is thrown in which test, they need to run in sequence instead of parallel, which  takes some time (13 minutes on my machine for one TFM).

Here is the code to check for `UnobservedTaskException` for each test: https://github.com/microsoft/playwright-dotnet/commit/3e29b6a062b6b7de2d1c025b233eb95a2cff6cf0
Here is a list of tests for which a `UnobservedTaskException` is currently thrown:
```
Fehler Microsoft:Playwright:Tests:BrowserContextBasicTests:CloseShouldAbortWaitForEvent [46 ms]
Fehler Microsoft:Playwright:Tests:BrowserContextCSPTests:ShouldBypassCSPHeader [1 s]
Fehler Microsoft:Playwright:Tests:BrowserContextCSPTests:ShouldBypassCSPInIframesAsWell [1 s]
Fehler Microsoft:Playwright:Tests:BrowserContextCSPTests:ShouldBypassCSPMetatag [1 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:DisconnectedEventShouldBeEmittedWhenBrowserIsClosedOrServerIsClosed [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:DisconnectedEventShouldHaveBrowserAsArguments [1 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldBeAbleToConnectTwoBrowsersAtTheSameTime [1 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldBeAbleToReconnectToABrowser [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldEmitCloseEventsOnPagesAndContexts [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldErrorWhenSavingDownloadAfterDeletion [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldNotThrowOnCloseAfterDisconnect [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldNotThrowOnContextCloseAfterDisconnect [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldNotThrowOnPageCloseAfterDisconnect [1 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldRejectNavigationWhenBrowserCloses [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldRejectWaitForSelectorWhenBrowserCloses [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldSaveAsVideosFromRemoteBrowser [3 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldSaveDownload [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldSaveHar [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldSetTheBrowserConnectedState [1 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldSupportSlowMo [2 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldTerminateNetworkWaiters [1 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldThrowWhenUsedAfterIsConnectedReturnsFalse [1 s]
Fehler Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldThrowWhenWhenCallingWaitForNavigationAfterDisconnect [2 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ContextAddCookiesShouldWork [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ContextClearCookiesShouldWork [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ContextCookiesShouldWork [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ShouldNotBlockThirdPartyCookies [3 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ShouldSupportBypassCSPOption [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ShouldSupportHttpCredentialsOption [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ShouldSupportOfflineOption [895 ms]
Fehler Microsoft:Playwright:Tests:DefaultBrowserContext1Tests:ShouldSupportUserAgentOption [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowsercontext2Tests:ShouldSupportExtraHTTPHeadersOption [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowsercontext2Tests:ShouldSupportGeolocationAndPermissionsOptions [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowsercontext2Tests:ShouldSupportHasTouchOption [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowsercontext2Tests:ShouldSupportIgnoreHTTPSErrorsOption [1 s]
Fehler Microsoft:Playwright:Tests:DefaultBrowsercontext2Tests:ShouldWorkInPersistentContext [1 s]
Fehler Microsoft:Playwright:Tests:DownloadTests:ShouldExposeStream [1 s]
Fehler Microsoft:Playwright:Tests:PageBasicTests:ShouldFailWithErrorUponDisconnect [183 ms]
Fehler Microsoft:Playwright:Tests:PageBasicTests:ShouldTerminateNetworkWaiters [286 ms]
Fehler Microsoft:Playwright:Tests:PageClickTimeout2Tests:ShouldTimeoutWaitingForVisibilityHiddenToBeGone [5 s]
Fehler Microsoft:Playwright:Tests:PageGotoTests:ShouldNotCrashWhenNavigatingToBadSSLAfterACrossOriginNavigation [495 ms]
Fehler Microsoft:Playwright:Tests:PageRequestContinueTests:ShouldNotThrowWhenContinuingAfterPageIsClosed [175 ms]
Fehler Microsoft:Playwright:Tests:PageRouteTests:ShouldBeAbortableWithCustomErrorCodes [167 ms]
Fehler Microsoft:Playwright:Tests:PopupTests:ShouldRespectRoutesFromBrowserContext [709 ms]
Fehler Microsoft:Playwright:Tests:PopupTests:ShouldRespectRoutesFromBrowserContextUsingWindowOpen [545 ms]
```

Complete output (german): [output.txt](https://github.com/microsoft/playwright-dotnet/files/7895483/output.txt)

Resources:
- https://github.com/nunit/nunit/issues/2220#issue-233750693
- https://github.com/nunit/nunit/issues/2220#issuecomment-430106577
- https://stackoverflow.com/a/3284286

---

For example with the `Microsoft:Playwright:Tests:BrowserTypeConnectTests:ShouldSaveDownload` test I discovered that this code is missing an `await` as it currently raises a `UnobservedTaskException`:
https://github.com/microsoft/playwright-dotnet/blob/829e0d41804809f85f683063623fb3d437d48cb7/src/Playwright/Core/BrowserType.cs#L150

FYI: just discarding the task with `_ = ...` does not work as the `task.Exception` is still unobserved!

mxschmitt: > FYI: just discarding the task with _ = ... does not work as the task.Exception is still unobserved!

yeah this is known. We can get in theory rid of all `_ = ...` things and just keep the `...` afaik.

I'd like to follow-up on that then to keep the diff small of the PR, I rebased this PR back to its original form. We can then add an additional GitHub Action job which runs the tests only in chromium and verifies that there are no unhandled exceptions thrown. This might be better instead of increasing all the jobs up to 13+ minutes.

I started locally with adding a method onto the Task calling it FireAndForget() which is like `myAction().catch(() => {})` in JS. If you are interested in adding the GitHub Action job, let me know, otherwise I can do it too. Thanks!

campersau: Okay! Yes, I can add the GitHub Action job for that.

src/Playwright/Core/BrowserContext.cs
pavelfeldman: Let's make sure to annotate palaces where we need to handle unobserved exception with _ =.

src/Playwright/Core/BrowserContext.cs
pavelfeldman: Same here

src/Playwright/Core/BrowserContext.cs
```diff
@@ -326,55 +326,41 @@ public Task<IPage> RunAndWaitForPageAsync(Func<Task> action, BrowserContextRunAn
...
-            }
-
-            return url;
-        }
-
-        private Task RouteAsync(Regex urlRegex, Func<string, bool> urlFunc, Action<IRoute> handler, BrowserContextRouteOptions options)
+        private Task RouteAsync(string urlString, Regex urlRegex, Func<string, bool> urlFunc, Action<IRoute> handler, BrowserContextRouteOptions options)
             => RouteAsync(new()
             {
+                Url = urlString,
```
pavelfeldman, line 350: You seem to be reverting a good change.

src/Playwright/Core/BrowserContext.cs
```diff
@@ -326,55 +326,41 @@ public Task<IPage> RunAndWaitForPageAsync(Func<Task> action, BrowserContextRunAn
...
                 Regex = urlRegex,
                 Function = urlFunc,
                 Handler = handler,
                 Times = options?.Times,
+                BaseURL = this.Options?.BaseURL,
```
pavelfeldman, line 355: Ditto

src/Playwright/Core/BrowserContext.cs
```diff
@@ -326,55 +326,41 @@ public Task<IPage> RunAndWaitForPageAsync(Func<Task> action, BrowserContextRunAn
...
             });
 
-        private Task RouteAsync(RouteSetting setting)
+        private Task RouteAsync(RouteHandler route)
         {
-            _routes.Insert(0, setting);
+            // Make sure that the given route match can be executed (e.g. invalid pattern)
+            route.Matches("http://127.0.0.1");
```
pavelfeldman, line 361: This looks like we are deliberately making things worse.

src/Playwright/Core/Route.cs
pavelfeldman: Let's annotate it with _ = for now.

src/Playwright/Helpers/UrlHelpers.cs
pavelfeldman: This is going backwards now, let's fix upstream instead.

src/Playwright/Core/Route.cs
campersau: Just noticed that a new GitHub Action job isn't needed since the tests are already running in sequence in CI. `NUnit.NumberOfTestWorkers=1`

So here are all the changes required to make tests pass without raising `UnobservedTaskException`:
https://github.com/microsoft/playwright-dotnet/compare/main...campersau:UnobservedTaskExceptions
Here is a sample CI run: https://github.com/campersau/playwright-dotnet/actions/runs/1718338242

But I will wait with the PR for your `FireAndForget` and adjust to it accordingly.

PR #1939, opened 2022-01-16
chore: implement Artifact stream

mxschmitt:


src/Playwright/Core/Stream.cs
```diff
@@ -42,8 +46,55 @@ internal Stream(IChannelOwner parent, string guid) : base(parent, guid)
...
         public StreamChannel Channel { get; }
 
-        public async ValueTask DisposeAsync() => await Channel.CloseAsync().ConfigureAwait(false);
+        public bool IsClosed { get; set; }
+
+        public StreamImpl StreamImpl => new(this);
+
+        public Task<byte[]> ReadAsync(int size) => Channel.ReadAsync(size);
+
+        public async ValueTask DisposeAsync() => await CloseAsync().ConfigureAwait(false);
```
pavelfeldman, line 55: public ValueTask DisposeAsync() => CloseAsync();

src/Playwright/Core/Stream.cs
```diff
@@ -42,8 +46,55 @@ internal Stream(IChannelOwner parent, string guid) : base(parent, guid)
...
+    internal class StreamImpl : System.IO.Stream
+    {
+        private readonly Stream _stream;
+
+        internal StreamImpl(Stream stream)
+        {
+            _stream = stream;
+        }
+
+        public override bool CanRead => !_stream.IsClosed;
```
pavelfeldman, line 73: This should be always true - your stream supports reading.

src/Playwright/Core/Stream.cs
```diff
@@ -42,8 +46,55 @@ internal Stream(IChannelOwner parent, string guid) : base(parent, guid)
...
+
+        public override bool CanSeek => false;
+
+        public override bool CanWrite => false;
+
+        public override long Length => throw new NotImplementedException();
```
pavelfeldman, line 79: Here and below, throw NotSupportedException instead.

src/Playwright/Core/Stream.cs
```diff
@@ -42,8 +46,55 @@ internal Stream(IChannelOwner parent, string guid) : base(parent, guid)
 
         public StreamChannel Channel { get; }
 
-        public async ValueTask DisposeAsync() => await Channel.CloseAsync().ConfigureAwait(false);
+        public bool IsClosed { get; set; }
```
pavelfeldman, line 49: I don't think you need this.

src/Playwright/Core/Stream.cs
```diff
@@ -23,10 +23,14 @@
  */
 
 using System;
+using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Playwright.Transport;
 using Microsoft.Playwright.Transport.Channels;
 
+#pragma warning disable SA1402 // A C# code file contains more than one unique type.
```
pavelfeldman, line 32: Don't violate, put it into a separate file.

src/Playwright/Core/Stream.cs
mxschmitt: I thought the same but ValueTask != Task:

image

We do the same in other places too.

campersau:
```suggestion
        public ValueTask DisposeAsync() => new ValueTask(CloseAsync());
```
This would work.

PR #1945, opened 2022-01-18
remove unused _waitingForObject from Connection

campersau:
Nobody was adding anything to `_waitingForObject`. Only trying to remove from it. Might be a left over?

The same is true for js:
https://github.com/microsoft/playwright/blob/976af162b07eb9b056f83f4c0439f10d0d4aba5c/packages/playwright-core/src/client/connection.ts#L59
https://github.com/microsoft/playwright/blob/976af162b07eb9b056f83f4c0439f10d0d4aba5c/packages/playwright-core/src/client/connection.ts#L266-L270

### DIFF
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -43,7 +43,6 @@ namespace Microsoft.Playwright.Transport
 {
     internal class Connection : IDisposable
     {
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):
LGTM, feel free to remove it upstream too.

PR #1948, opened 2022-01-19
chore: use GetBytesFromBase64 instead of manually calling Convert.Fro…

campersau:
…mBase64String

### DIFF
--- a/src/Playwright/Transport/Protocol/Generated/RequestInitializer.cs
+++ b/src/Playwright/Transport/Protocol/Generated/RequestInitializer.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Playwright.Transport.Protocol
 
         public string Method { get; set; }
 
-        public byte[] PostData { get; set; }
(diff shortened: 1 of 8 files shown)


src/Playwright/Transport/Protocol/Generated/RequestInitializer.cs
```diff
@@ -36,7 +36,7 @@ internal class RequestInitializer
 
         public string Method { get; set; }
 
-        public string PostData { get; set; }
+        public byte[] PostData { get; set; }
```
mxschmitt, line 39: These are generated, so we can't change them since over the wire it's transferred as a base64 string. The rest looks good.

campersau: System.Text.Json will automatically convert byte[] properties to base64. https://dotnetfiddle.net/i5Ae28
But if this got generated I will revert it.

mxschmitt (approved):

Issue #1845, opened 2021-11-30
Configuration: Specify platform


LukeTOBrien:
Hello,

When you build a project with the Playwright NuGet installed there is a folder called `.playwright` in the `bin/Debug` or `bin/Release`, inside this folder is two more folders: `node` (197MB) and `package` (11.8MB).  
`node` folder contains:

* `linux` (70.9MB)
* `mac` (72.6MB
* `win32_64` (54.2MB)

Is there a way to specify that I only intend to use `win32_64` platform say? That would mean that the other two are not needed and thus save space.  
I also noticed that `win32_64` folder contains `node.exe`, is that strictly necessary? As I have node installed and set in the PATH environmental variable.

The project I am working perhaps is using Playwright differently from intended, I have a WYSIWYG editor to edit a web site and  I want to export the site as a PDF, so I'm using the [Page.Pdf method](https://playwright.dev/docs/api/class-page/#page-pdf) to do that.  
The problem is 197MB is a bit large to send to my server via FTP, so I'm wondering how to reduce.

Thanks,

Luke

vandre:
I'm also using Playwright for PDF generation. As a workaround what I did is manually delete the linux/mac folders on my deployment server. Saved a little bit of space. It seems to still work.

pavelfeldman:
My understanding is that Nuget does not allow us to download selective binaries, for the selected platform. So the only solution to this is platform-specific Nuget packages, something we'd want to see upvoted / validated before we do it.

campersau:
The `CopyPlaywrightFilesToOutput` target could be adjusted to only copy the node folder for the current OS or check for a custom `<PlaywrightPlatform>` property so the user can set it themselves in their project. E.g.

```xml
<PropertyGroup>
  <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux'))">linux</PlaywrightPlatform>
  <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Windows'))">windows</PlaywrightPlatform>
  <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('OSX'))">macos</PlaywrightPlatform>
  <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">all</PlaywrightPlatform>
</PropertyGroup>
[...]

Issue #1850, opened 2021-12-03
[Feature]: Remove NodeJS runtime from package


Liero:
### Feature request

The Problem:

Including a NodeJS binaries slows down a CI/CD pipeline. 

It has to be restored on each build (Azure Hosted pipeline). 

Given that I have multiple projects referencing playwright, I have a gigabyte of unnecessary binaries that I have to upload as build artifacts and downloads in my release pipeline. My CI produces 300GB of build artifacts per months with only two developers pushing the code.

Now imagine, that I other nuget references would include NodeJS runtime as well!
It has to be removed or made optional.

pavelfeldman:
We consider Node.js runtime to be an implementation detail of Playwright. I'm not sure how deduping Node.js helps addressing the CI/CD challenges? We would still need Node.js on CI to run the tests.

> Including a NodeJS binaries slows down a CI/CD pipeline.

My understanding is that we are talking about tens of milliseconds, not even hundreds. Given that the typical test run is minutes, this should be within the error.

> It has to be restored on each build (Azure Hosted pipeline).

Could you elaborate on what is getting restored?

> Given that I have multiple projects referencing playwright, I have a gigabyte of unnecessary binaries that I have to upload as build artifacts and downloads in my release pipeline. My CI produces 300GB of build artifacts per months with only two developers pushing the code.

Do I understand it right that playwright binaries are a part of the build artifacts?

More information would help us understand and prioritize this request.

Liero:
> We consider Node.js runtime to be an implementation detail of Playwright.

Well, `Microsoft.TypeScript.MSBuild` package is also dependent on node and thanks god it does not include yet another .net runtime 🙄 

> We would still need Node.js on CI to run the tests

There is Node.js installed on all Azure hosted build agents or github actions. Otherwise you would use a docker image with Node, or a separate step  to install node as part of you pipeline. Including NodeJS as part of the package is unnecessary in wast majority of cases.

> Could you elaborate on what is getting restored?

Nuget package references are being restored on each CI build. If the Microsoft.Playwright package is 200MB instead of ten, it is it slower about ten seconds. Azure Hosted agent, each pipeline run does a fresh restore, so it has to be downloaded from nuget.org each time.

> Do I understand it right that playwright binaries are a part of the build artifacts?
[...]

Issue #1803, opened 2021-10-26
[Question]: Is it possible to build a "lightweight" NuGet package with support for just Linux/Chromium?


maxmoore14:
### Your question

I had accidentally asked this question on the wrong project, but was referred to this thread: https://github.com/microsoft/playwright/issues/2404

I would like to use Playwright on AWS Lambda with .NET Core but the total package size is too large. I was thinking perhaps I could fork the repo and build a lightweight version, but would appreciate any guidance from the community on how best to do that.

brnbs:
On the NuGet page (https://www.nuget.org/packages/Microsoft.Playwright) it says the package is 96.09 MB. After inspecting the nupkg file, I see the same as @maxmoore14. Can someone please clarify the purpose of this large ".playwright" folder and the included node runtime for all OSes?

mxschmitt:
This should be possible after #1955. (after things change only the driver of the current architecture gets added to the bin directory. Installing single browsers is already possible as of today.)

PR #1955, opened 2022-01-21
copy node folder based on platform

### LINKED ISSUE #1915: [Question]: Driver not found after package a WPF in msix.
### Your question

**Environment:**
**OS:** Win 10 20H2 19042.1415
**IDE:** Visual Studio Pro 2019
**SDK:** .Net Core 3.1
**Playwright:** Microsoft.Playwright 1.16.1

**Issues:**
I'm trying to build a WPF application with Playwright, everything works perfectly until I tried to package the WPF.

I followed the instruction here: https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/modernize-wpf-tutorial-5
to package my WPF app into MSIX, then after deploying it, I got the "Driver not found" error.

Then when checking the App folder, the `.playwright` directory is missing, but in the original WPF app, the `.playwright` folder does exist in the `bin` path.

From my understanding, the packaging process should package all content within the `bin` path, and release them once deployed, but why the `.playwright` folder is missed?

I would like the WPF app could be packaged well so that it could be deployed to other machines easily, so manually installing the playwright on the target machine is not a good choice.

Any idea about this?

### LINKED ISSUE #1952: [BUG] Error when produce single file
When I publish the application, check the produce single file, it will run error.
- Playwright Version: 1.17.3
- Operating System: Windows
- .NET version: .Net6
- Browser: All
- Extra: Console application

System.ArgumentException: The path is empty. (Parameter 'path')
   at System.IO.Path.GetFullPath(String path)
   at System.IO.FileInfo..ctor(String originalPath, String fullPath, String fileName, Boolean isNormalized)
   at System.IO.FileInfo..ctor(String fileName)
   at Microsoft.Playwright.Helpers.Paths.GetExecutablePath()
   at Microsoft.Playwright.Transport.StdIOTransport.GetProcess()
   at Microsoft.Playwright.Transport.StdIOTransport..ctor()
   at Microsoft.Playwright.Playwright.CreateAsync()
   at BrowserKeep.BrowserKeepService.Keep()

Sorry for my poor English.

campersau:
This will by default only copy the node platform folder for the current platform.
The `RuntimeIdentifier` will override it, so you could do:
```
dotnet publish -r linux-x64
dotnet publish -r win-x64
dotnet publish -r osx-x64
```
and it will copy the correct platform folder even when the command is run on another platform. Even
```
dotnet publish -r ubuntu-x64
```
works because `RuntimePack.RuntimeIdentifier` will resolve to `linux-x64`.

To go back to the old behavior of copying all node platforms add the following to your project file:
```xml
<PropertyGroup>
  <PlaywrightPlatform>all</PlaywrightPlatform>
</PropertyGroup>
```` 
If the platform is unknown (which includes `all`) it will also copy everything.

There is also the option of not copying the `.playwright` folder at all by using:
```xml
<PropertyGroup>
  <PlaywrightPlatform>none</PlaywrightPlatform>
</PropertyGroup>
```` 

Fixes https://github.com/microsoft/playwright-dotnet/issues/1845
Fixes #1952
Fixes #1915

Related:
https://github.com/microsoft/playwright-dotnet/issues/1850#issuecomment-989621343
https://github.com/microsoft/playwright-dotnet/issues/1803#issuecomment-953359568

### DIFF
--- a/src/Playwright/build/Microsoft.Playwright.targets
+++ b/src/Playwright/build/Microsoft.Playwright.targets
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Target Name="_GetPlaywrightCopyItems">
+  <Target Name="CopyPlaywrightFilesToOutput" AfterTargets="ProcessFrameworkReferences">
     <PropertyGroup>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">%(RuntimePack.RuntimeIdentifier)</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux'))">linux</PlaywrightPlatform>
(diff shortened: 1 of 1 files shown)


campersau:
Instead of manually copying the `.playwright` folder the files can be added as `<Content>` items which makes it possible to a `PublishSingleFile` with `<IncludeAllContentForSelfExtract>true</IncludeAllContentForSelfExtract>`.

Fixes #1952

campersau:
Added support for setting `<PlaywrightPlatform>none</PlaywrightPlatform>` which won't copy the `.playwright` folder at all.
See https://github.com/microsoft/playwright-dotnet/issues/1850#issuecomment-1019389596 (even though this could be fixed individually with some custom msbuild logic or CI copy exclusions)

mxschmitt (approved):
This patch looks great! We need to follow-up with docs in the intro-csharp.md in `https://github.com/microsoft/playwright` and then we are good I think!

Thank you!

PR #1965, opened 2022-01-25
fix: save trace without path

### LINKED ISSUE #1964: [BUG] In version 1.18 it is not possible to stop tracing without saving the trace
**Context:**
- Playwright Version: 1.18
- Operating System: Linux
- .NET version: .NET 6
- Browser: All

**Describe the bug**

In version 1.18 one cannot stop the tracing without saving it in a file and not getting an excption. The check if the path is empty moved before the check if artifacts are generated and if the path is empty an exception is thrown. In version 1.17 it was different and one could send null into Tracing.StopAsync() to stop the tracing without saving traces to the file.
We used this in our tests to only save the traces if the test fails. 
Is this change made deliberately? If so, how can one stop tracing without saving a file and without getting an exception?

image

mxschmitt:
Fixes #1964

### DIFF
--- a/src/Playwright.Tests/TracingTests.cs
+++ b/src/Playwright.Tests/TracingTests.cs
@@ -143,6 +143,22 @@ namespace Microsoft.Playwright.Tests
             Assert.AreEqual(sourceTraceFileContent, currentFileContent);
         }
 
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

ThisFunctionalTom:
Thx. That was very fast. ❤️

PR #1968, opened 2022-01-25
chore: ensure no unhandled exceptions are thrown

### LINKED ISSUE #1963: [BUG] Ensure no unhandled exceptions are thrown
e.g. implementation: https://github.com/microsoft/playwright-dotnet/compare/main...campersau:UnobservedTaskExceptions

I think there are no blockers cc @campersau.

campersau:
- Adds an `TaskScheduler.UnobservedTaskException` event handler for each test to check if there are any unobserved exceptions
- Sets `[LevelOfParallelism(1)]` to ensure that each test is run in sequence. This was already the case for CI where `NUnit.NumberOfTestWorkers=1` is used, which now could be removed. And now it also applies when run locally.
https://github.com/microsoft/playwright-dotnet/blob/c12b4fcd8061b396ef4d4587dafd9b876202f009/.github/workflows/tests.yml#L42
- Add `IgnoreException` utility which marks all exceptions as handled if a task faulted.
- Fixes dispose order for browser / browsercontext and tempdirectory to avoid unnecessary exceptions. Deleting the tempdirectory before disposing the browser throws a `UnauthorizedAccessException`.

Fixes #1963

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,12 +39,12 @@ jobs:
         if: matrix.os != 'ubuntu-latest'
         env:
           BROWSER: ${{ matrix.browser }}
-        run: dotnet test ./src/Playwright.Tests/Playwright.Tests.csproj -c Debug -f net5.0 --logger "trx;LogFileName=TestResults.xml" -- NUnit.NumberOfTestWorkers=1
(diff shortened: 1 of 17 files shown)


mxschmitt (approved):
amazing work! Yeah we can remove `Unit.NumberOfTestWorkers=1` everywhere now I guess.

pavelfeldman (approved):
This looks awesome! We have more places where we'd like to IgnoreExceptions, but we can follow up on those.

PR #1973, opened 2022-01-26
chore: follow up on ignoring task exceptions

mxschmitt:
#1963

### DIFF
--- a/src/Playwright/Core/BrowserContext.cs
+++ b/src/Playwright/Core/BrowserContext.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Playwright.Core
             set
             {
                 _defaultNavigationTimeout = value;
-                _ = Channel.SetDefaultNavigationTimeoutNoReplyAsync(value);
+                Channel.SetDefaultNavigationTimeoutNoReplyAsync(value).IgnoreException();
             }
         }
 
@@ -132,7 +132,7 @@ namespace Microsoft.Playwright.Core
             set
             {
                 _defaultTimeout = value;
-                _ = Channel.SetDefaultTimeoutNoReplyAsync(value);
+                Channel.SetDefaultTimeoutNoReplyAsync(value).IgnoreException();
             }
         }
 
@@ -351,7 +351,7 @@ namespace Microsoft.Playwright.Core
                 }
             }
 
-            _ = route.ContinueAsync().ConfigureAwait(false);
+            route.ContinueAsync().IgnoreException();
         }
 
         internal async Task DisableInterceptionAsync()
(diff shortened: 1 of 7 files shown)


src/Playwright/Core/BrowserContext.cs
```diff
@@ -453,9 +453,9 @@ private void Channel_OnPage(object sender, BrowserContextPageEventArgs e)
 
         private void Channel_BindingCall(object sender, BindingCallEventArgs e)
         {
-            if (_bindings.TryGetValue(e.BidingCall.Name, out var binding))
+            if (_bindings.TryGetValue(e.BindingCall.Name, out var binding))
             {
-                _ = e.BidingCall.CallAsync(binding);
+                e.BindingCall.CallAsync(binding).ConfigureAwait(false);
```
mxschmitt, line 458: There I'd say we should intentionally not ignore the exception and it should lead to an unhandled exception.

src/Playwright/Core/Page.cs
```diff
@@ -1020,9 +1020,9 @@ private void Channel_Crashed(object sender, EventArgs e)
 
         private void Channel_BindingCall(object sender, BindingCallEventArgs e)
         {
-            if (Bindings.TryGetValue(e.BidingCall.Name, out var binding))
+            if (Bindings.TryGetValue(e.BindingCall.Name, out var binding))
             {
-                _ = e.BidingCall.CallAsync(binding);
+                e.BindingCall.CallAsync(binding).ConfigureAwait(false);
```
mxschmitt, line 1025: ditto

src/Playwright/Core/BrowserContext.cs
pavelfeldman: I don't think you need .ConfigureAwait(false) if you don't await.

src/Playwright/Core/Page.cs
pavelfeldman: ditto

pavelfeldman (approved):

src/Playwright/Core/Route.cs
```diff
@@ -83,12 +83,19 @@ public Task ContinueAsync(RouteContinueOptions options = default)
...
+            }
 
             // When page closes or crashes, we catch any potential rejects from this Route.
             // Note that page could be missing when routing popup's initial request that
             // does not have a Page initialized just yet.
-            // TODO: ignore exceptions
-            await Task.WhenAny(task, page.ClosedOrCrashedTcs.Task).ConfigureAwait(false);
+            if (task != await Task.WhenAny(task.ContinueWith(t => t.Exception.Handle(_ => true), System.Threading.CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default), page.ClosedOrCrashedTcs.Task).ConfigureAwait(false))
+            {
+                task.IgnoreException();
```
campersau, line 97: This will ignore the exceptions twice as they are already ignore above with
```c#
task.ContinueWith(t => t.Exception.Handle(_ => true), System.Threading.CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default)
```
Also `ContinueWith` will return a new task so the if condition is always true.

mxschmitt: feel free to follow-up! Thanks.

PR #1976, opened 2022-01-27
don't suppress route exceptions when page is still open

campersau:
follow up to https://github.com/microsoft/playwright-dotnet/pull/1973#discussion_r793594152

also added test from https://github.com/microsoft/playwright/blob/5ba7903ba098586a13745e0d7ac894f1d55d47aa/tests/page/page-request-continue.spec.ts#L65-L78

### DIFF
--- a/src/Playwright.Tests/PageRequestContinueTests.cs
+++ b/src/Playwright.Tests/PageRequestContinueTests.cs
@@ -127,5 +127,26 @@ namespace Microsoft.Playwright.Tests
             await tsc.Task;
             await done;
         }
+
+        [PlaywrightTest("page-request-continue.spec.ts", "should not allow changing protocol when overriding url")]
+        public async Task ShouldNotAllowChangingProtocolWhenOverridingUrl()
+        {
+            PlaywrightException exception = null;
+            await Page.RouteAsync("**/empty.html", async (route) =>
+            {
+                try
+                {
+                    await route.ContinueAsync(new RouteContinueOptions { Url = "file:///tmp/foo" });
+                }
+                catch (PlaywrightException ex)
+                {
+                    exception = ex;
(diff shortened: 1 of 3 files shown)


src/Playwright/Helpers/TaskHelper.cs
```diff
@@ -193,7 +193,7 @@ private static async Task<bool> TimeoutTask(Task task, TimeSpan timeout)
             using var cancellationToken = new CancellationTokenSource(timeout);
             using (cancellationToken.Token.Register(s => ((TaskCompletionSource<bool>)s).TrySetResult(true), tcs))
             {
-                if (tcs.Task == await Task.WhenAny(task, tcs.Task).ConfigureAwait(false))
+                if (task != await Task.WhenAny(task, tcs.Task).ConfigureAwait(false))
```
campersau, line 196: Just changed that for consistentcy.

src/Playwright.Tests/PageRequestContinueTests.cs
```diff
@@ -127,5 +127,26 @@ public async Task ShouldNotThrowWhenContinuingAfterPageIsClosed()
             await tsc.Task;
             await done;
         }
+
+        [PlaywrightTest("page-request-continue.spec.ts", "should not allow changing protocol when overriding url")]
+        public async Task ShouldNotAllowChangingProtocolWhenOverridingUrl()
+        {
+            PlaywrightException exception = null;
```
campersau, line 134: Adjust to https://github.com/microsoft/playwright/pull/11701/files which is similar to what I had before I have seen the test in JavaScript.

mxschmitt (approved):
I tried it before like you had but missed the else, this was probably the reason why it was not working for me.

PR #1978, opened 2022-01-27
update Locator AllInnerText / AllTextContent

campersau:
`Locator.AllInnerTextAsync` / `Locator.AllTextContentAsync` didn't dispose the intermediate `IElementHandle` objects which is what the first commit fixes.

The second commit changes these to use `Locator.EvaluateAllAsync` instead so there are no intermediate objects created in the first place which results in better performance and memory usage.
This is also what the JS version is using:
https://github.com/microsoft/playwright/blob/0606afb2e66737a352cf8627fa0260ea2fceb267/packages/playwright-core/src/client/locator.ts#L241
https://github.com/microsoft/playwright/blob/0606afb2e66737a352cf8627fa0260ea2fceb267/packages/playwright-core/src/client/locator.ts#L245

Please let me know which version is preferred.

### DIFF
--- a/src/Playwright/Core/Locator.cs
+++ b/src/Playwright/Core/Locator.cs
@@ -61,24 +61,11 @@ namespace Microsoft.Playwright.Core
 
         public ILocator Last => new Locator(_frame, $"{_selector} >> nth=-1");
 
-        public async Task<IReadOnlyList<string>> AllInnerTextsAsync()
-        {
-            var handles = await _frame.QuerySelectorAllAsync(_selector).ConfigureAwait(false);
-            return (await handles
-                    .SelectAsync(async x => await x.InnerTextAsync().ConfigureAwait(false))
-                    .ConfigureAwait(false))
-                .ToArray();
-        }
+        public Task<IReadOnlyList<string>> AllInnerTextsAsync()
+            => WithElementsAsync(h => h.InnerTextAsync());
 
-        public async Task<IReadOnlyList<string>> AllTextContentsAsync()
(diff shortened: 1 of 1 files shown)


src/Playwright/Core/Locator.cs
```diff
@@ -62,23 +62,10 @@ public Locator(Frame parent, string selector, LocatorLocatorOptions options = nu
...
         public async Task<IReadOnlyList<string>> AllTextContentsAsync()
-        {
-            var handles = await _frame.QuerySelectorAllAsync(_selector).ConfigureAwait(false);
-            return (await handles
-                    .SelectAsync(async x => await x.TextContentAsync().ConfigureAwait(false))
-                    .ConfigureAwait(false))
-                .Select(x => x ?? string.Empty) // we don't filter nulls, as per https://github.com/microsoft/playwright/blob/master/src/client/locator.ts#L205
-                .ToArray();
-        }
+            => await EvaluateAllAsync<string[]>("ee => ee.map(e => e.textContent || '')").ConfigureAwait(false); // we don't filter nulls, as per https://github.com/microsoft/playwright/blob/master/src/client/locator.ts#L205
```
mxschmitt, line 68:
```suggestion
            => await EvaluateAllAsync<string[]>("ee => ee.map(e => e.textContent || '')").ConfigureAwait(false);
```

mxschmitt (approved):

mxschmitt:
The second version is preferred since its aligned with upstream, thanks!

Issue #1974, opened 2022-01-27
[internal] upgrade project tooling to .NET 6


mxschmitt:
Started with it a while ago in https://github.com/microsoft/playwright-dotnet/pull/1884 and extracted some patches. It should be possible to upgrade now with minimal code changes (.cs) and no TFM changes since .NET 5 goes EOL in April.

PR #1986, opened 2022-02-05
chore: migrate tooling over to .NET 6

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/1974
Relates https://github.com/microsoft/playwright-dotnet/pull/1884

Motivation: .NET 5 reaches EOL in May 8, 2022 since it was no LTS, LTS is .NET 6. There are no changes in the TFM so no user facing changes and only tooling.

### DIFF
--- a/.build/azure-pipeline-nuget.yml
+++ b/.build/azure-pipeline-nuget.yml
@@ -28,10 +28,17 @@ stages:
     steps:
 
     - task: UseDotNet@2
-      displayName: 'Use .NET Core sdk'
+      displayName: 'Use .NET 5 SDK'
       inputs:
         packageType: sdk
         version: 5.x
+        installationPath: $(Agent.ToolsDirectory)/dotnet
+
+    - task: UseDotNet@2
+      displayName: 'Use .NET 6 SDK'
+      inputs:
+        packageType: sdk
+        version: 6.x
         installationPath: $(Agent.ToolsDirectory)/dotnet
 
     - task: MicroBuildSigningPlugin@3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   test-net5:
-    name: ${{ matrix.browser }}/${{ matrix.os }}/.NET 5
+    name: ${{ matrix.browser }}/${{ matrix.os }}/.NET 6
     runs-on: ${{ matrix.os }}
     timeout-minutes: 45
     strategy:
@@ -25,12 +25,14 @@ jobs:
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 5.0.x
+          dotnet-version: |
+            5.0.x
+            6.0.x
       - name: Install prerequisites and download drivers
         shell: bash
         run: ./build.sh --download-driver
       - name: Building
-        run: dotnet build ./src
+        run: dotnet build -f net5.0 ./src
       - name: Installing Browsers and dependencies
         run: dotnet run --project ./src/Playwright/Playwright.csproj -f net5.0 -- install --with-deps ${{ matrix.browser }}
       - name: Running tests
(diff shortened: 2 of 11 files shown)


src/.editorconfig
```diff
@@ -185,6 +185,7 @@ dotnet_diagnostic.SA1202.severity = none
 dotnet_diagnostic.SA1503.severity = none
 dotnet_diagnostic.RCS1075.severity = none
 dotnet_diagnostic.SA1402.severity = none
+dotnet_diagnostic.CA1844.severity = none
```
mxschmitt, line 188: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1844

Was ocurring here: https://github.com/microsoft/playwright-dotnet/blob/4f52f655bb45143d494848508979067c4654b023/src/Playwright/Core/Stream.cs#L79-L84

.build/azure-pipeline-nuget.yml
```diff
@@ -28,10 +28,10 @@ stages:
     steps:
 
     - task: UseDotNet@2
-      displayName: 'Use .NET Core sdk'
+      displayName: 'Use .NET 6 SDK'
       inputs:
         packageType: sdk
-        version: 5.x
+        version: 6.x
```
mxschmitt, line 34: I think we need to install on our deployment pipeline 5 + 6.

src/Playwright/Core/Artifact.cs
```diff
@@ -69,11 +69,13 @@ public async Task SaveAsAsync(string path)
                 return;
             }
             System.IO.Directory.CreateDirectory(Path.GetDirectoryName(path));
-            await using var stream = await _channel.SaveAsStreamAsync().ConfigureAwait(false);
-
-            using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
+            var stream = await _channel.SaveAsStreamAsync().ConfigureAwait(false);
+            await using (stream)
```
campersau, line 73:
```suggestion
            await using (stream.ConfigureAwait(false))
```

.github/workflows/tests.yml
```diff
@@ -49,14 +51,13 @@ jobs:
...
+      - name: Setup .NET
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 3.1.x
-      - name: Setup .NET 5.0 # needed for our build steps
-        uses: actions/setup-dotnet@v1
-        with:
-          dotnet-version: 5.0.x
+          dotnet-version: |
+            3.1.x
```
kblok, line 58: I would drop 3.1

pavelfeldman (approved):

src/Playwright/Core/Artifact.cs
```diff
@@ -69,11 +69,13 @@ public async Task SaveAsAsync(string path)
...
             }
             System.IO.Directory.CreateDirectory(Path.GetDirectoryName(path));
-            await using var stream = await _channel.SaveAsStreamAsync().ConfigureAwait(false);
-
-            using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
+            var stream = await _channel.SaveAsStreamAsync().ConfigureAwait(false);
+            await using (stream.ConfigureAwait(false))
             {
-                await stream.StreamImpl.CopyToAsync(fileStream).ConfigureAwait(false);
+                using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
```
kblok, line 75: You should be able to put these two usings at the same level

```
await using (stream.ConfigureAwait(false))
using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
{
    ...
}
```

.github/workflows/tests.yml
```diff
@@ -25,12 +25,14 @@ jobs:
...
+            5.0.x
+            6.0.x
       - name: Install prerequisites and download drivers
         shell: bash
         run: ./build.sh --download-driver
       - name: Building
-        run: dotnet build ./src
+        run: dotnet build -f net5.0 ./src
       - name: Installing Browsers and dependencies
         run: dotnet run --project ./src/Playwright/Playwright.csproj -f net5.0 -- install --with-deps ${{ matrix.browser }}
```
kblok, line 37: If we don't drop .NET 5, we should add the version to a matrix to test both versions. I would drop .NET 5.

src/Playwright/Core/Artifact.cs
mxschmitt: linting rules are conflicting there thats why I put it that way: https://github.com/dotnet/roslyn-analyzers/issues/5712

.github/workflows/tests.yml
mxschmitt: we test on 'netcoreapp3.1' thats why its required I guess.

.github/workflows/tests.yml
mxschmitt: lets do it once .NET 5 reaches EOL or we follow-up earlier.

PR #1987, opened 2022-02-05
remove Microsoft.Bcl.AsyncInterfaces reference from net5.0

campersau:
It is only needed for `netstandard2.0`.

### DIFF
--- a/src/Playwright/Playwright.csproj
+++ b/src/Playwright/Playwright.csproj
@@ -36,7 +36,7 @@
     <PackageReference Include="System.Text.Json" Version="6.0.1" />
     <PackageReference Include="Macross.Json.Extensions" Version="2.2.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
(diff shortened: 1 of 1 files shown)


src/Playwright/Playwright.csproj
```diff
@@ -36,7 +36,7 @@
     <PackageReference Include="System.Text.Json" Version="6.0.1" />
     <PackageReference Include="Macross.Json.Extensions" Version="2.2.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
mxschmitt, line 39: looks like we want: `Condition="'$(TargetFramework)' == 'netstandard2.0' or $(TargetFramework.StartsWith('net4'))"` or what do you think?

campersau: No, that isn't needed because this project only targets `netstandard2.0;net5.0`:
https://github.com/microsoft/playwright-dotnet/blob/8c414af5c497e7aa97e47f276d03cb9ee6532c43/src/Playwright/Playwright.csproj#L10

When this project / nuget package is referenced it will automatically use the `netstandard2.0` DLLs for .NET Framework and also its references like `Microsoft.Bcl.AsyncInterfaces`.

mxschmitt (approved):

PR #1989, opened 2022-02-05
test: hide stdout/stderr inside CLI tests

mxschmitt:
Was logging stuff before.


campersau:
The code for `Console.Out` / `StringWriter` could also be moved into `[SetUp]` / `[TearDown]` as it is currently used by all / both of these tests.

aslushnikov (approved):

PR #1995, opened 2022-02-07
chore: simplify transport

mxschmitt:


src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -155,58 +129,35 @@ private void Dispose(bool disposing)
...
+            {
+                byte[] lengthBuffer = new byte[4];
 #if NETSTANDARD
 #pragma warning disable CA1835 // We can't use ReadOnlyMemory on netstandard
-                    int read = await stream.BaseStream.ReadAsync(buffer, 0, DefaultBufferSize, token).ConfigureAwait(false);
+                await stream.BaseStream.ReadAsync(lengthBuffer, 0, 4, token).ConfigureAwait(false);
 #pragma warning restore CA1835
 #else
-                    int read = await stream.BaseStream.ReadAsync(new(buffer, 0, DefaultBufferSize), token).ConfigureAwait(false);
+                await stream.BaseStream.ReadAsync(new(lengthBuffer, 0, 4), token).ConfigureAwait(false);
```
campersau, line 142: This currently assumes that the buffer will be fully read which might not be the case. The code before was correct.

Check if `ReadAsync` actually read all 4 bytes if not keep reading.

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -155,58 +129,35 @@ private void Dispose(bool disposing)
...
-            {
-                var stream = _process.StandardOutput;
-                byte[] buffer = new byte[DefaultBufferSize];
+            var stream = _process.StandardOutput;
 
-                while (!token.IsCancellationRequested && !_process.HasExited)
-                {
+            while (!token.IsCancellationRequested && !_process.HasExited)
+            {
+                byte[] lengthBuffer = new byte[4];
```
campersau, line 136: Would be nice if this could be reused (e.g. put outside of loop) and maybe clear it `lengthBuffer.AsSpan().Clear();` if necessary.

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -153,60 +127,37 @@ private void Dispose(bool disposing)
...
         }
 
-        private async Task GetResponseAsync(CancellationToken token)
+        private async Task OnMessageAsync(CancellationToken token)
         {
-            try
-            {
-                var stream = _process.StandardOutput;
-                byte[] buffer = new byte[DefaultBufferSize];
+            var stream = _process.StandardOutput;
```
campersau, line 132: Maybe get the `BaseStream` here so `.BaseStream` isn't needed below.

```suggestion
            var stream = _process.StandardOutput.BaseStream;
```

src/Playwright/Helpers/DebugHelpers.cs
```diff
@@ -23,41 +23,30 @@
...
-    /// </summary>
-    internal interface IConnectionTransport
+    internal static class DebugHelpers
     {
-        /// <summary>
-        /// Occurs when a message is received.
-        /// </summary>
-        event EventHandler<byte[]> MessageReceived;
+        [Conditional("DEBUG")]
+        internal static void PrintStack(string name)
```
mxschmitt, line 33: added some util which I've created manually over and over for debugging.

kblok (commented):
I know it's a draft. I just went through it quickly to see what's this about :)

PR #2010, opened 2022-02-14
feat: added apiName to connection metadata

mxschmitt:


src/Playwright.Tests/TracingTests.cs
```diff
@@ -65,14 +66,21 @@ await Context.Tracing.StartAsync(new()
...
+            string[] expectedActionApiNames = new string[] { "BrowserContext.NewPageAsync", "Page.GotoAsync", "Page.SetContentAsync", "Page.ClickAsync", "Mouse.MoveAsync", "Mouse.DblClickAsync", "Keyboard.InsertTextAsync", "Page.WaitForTimeoutAsync", "Page.CloseAsync", "Tracing.StopAsync" };
+            Assert.AreEqual(expectedActionApiNames, actualActionApiNames);
+
+            Assert.GreaterOrEqual(events.Where(e => e.Metadata?.ApiName == "Page.GotoAsync").Count(), 1);
+            Assert.GreaterOrEqual(events.Where(e => e.Metadata?.ApiName == "Page.SetContentAsync").Count(), 1);
+            Assert.GreaterOrEqual(events.Where(e => e.Metadata?.ApiName == "Page.ClickAsync").Count(), 1);
+            Assert.GreaterOrEqual(events.Where(e => e.Metadata?.ApiName == "Mouse.MoveAsync").Count(), 1);
+            Assert.GreaterOrEqual(events.Where(e => e.Metadata?.ApiName == "Mouse.DblClickAsync").Count(), 1);
+            Assert.GreaterOrEqual(events.Where(e => e.Metadata?.ApiName == "Keyboard.InsertTextAsync").Count(), 1);
+            Assert.GreaterOrEqual(events.Where(e => e.Metadata?.ApiName == "Page.CloseAsync").Count(), 1);
```
mxschmitt, line 79: add a test for frame.* methods.

src/Playwright/Transport/Connection.cs
```diff
@@ -430,5 +420,89 @@ internal void TraceMessage(string logLevel, object message)
...
+            }
+            if (string.IsNullOrEmpty(apiName))
+            {
+                apiName = lastInternalApiName;
+            }
+            EnsureApiZoneExists();
+            if (ApiZone.Value[0] != null)
+            {
+                return await action().ConfigureAwait(false);
+            }
```
campersau, line 471: Can this be moved to the top? It looks like the expensive `StackTrace(true)` stuff is only needed when there isn't an existing `ApiZone` already.

src/Playwright/Transport/Connection.cs
```diff
@@ -430,5 +420,89 @@ internal void TraceMessage(string logLevel, object message)
...
+            var st = new StackTrace(true);
+            var stack = new List<object>();
+            var lastInternalApiName = string.Empty;
+            var apiName = string.Empty;
+            for (int i = 0; i < st.FrameCount; ++i)
+            {
+                var sf = st.GetFrame(i);
+                string fileName = sf.GetFileName();
+                string cSharpNamespace = sf.GetMethod().ReflectedType.Namespace;
+                bool playwrightInternal = cSharpNamespace == "Microsoft.Playwright" || cSharpNamespace.StartsWith("Microsoft.Playwright.Core", StringComparison.InvariantCultureIgnoreCase) || cSharpNamespace.StartsWith("Microsoft.Playwright.Transport", StringComparison.InvariantCultureIgnoreCase) || cSharpNamespace.StartsWith("Microsoft.Playwright.Helpers", StringComparison.InvariantCultureIgnoreCase);
```
mxschmitt, line 440: I'm not a big fan of it, but we can't go via the fileName, since its not always there, only ReflectedType works then.
We need to ensure that its not our nunit and mstest harness and not our tests which have similar namespaces.

src/Playwright.Tests/BrowserTypeConnectTests.cs
```diff
@@ -41,48 +40,15 @@ public class BrowserTypeConnectTests : PlaywrightTestEx
         private RemoteServer _remoteServer;
 
         [SetUp]
-        public void SetUpAsync()
+        public void SetUpRemoteServer()
         {
-            try
-            {
-                DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
```
mxschmitt, line 48: note: drive-by cleanup

pavelfeldman (approved):

PR #2012, opened 2022-02-15
feat: add web-first assertions

mxschmitt:
#1800

### DIFF
--- a/src/Playwright.Tests/Assertions/LocatorAssertionsTests.cs
+++ b/src/Playwright.Tests/Assertions/LocatorAssertionsTests.cs
@@ -1,0 +1,312 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using static Microsoft.Playwright.Assertions;
+
+namespace Microsoft.Playwright.Tests.Assertions
+{
+    public class LocatorAssertionsTests : PageTestEx
+    {
+        [PlaywrightTest("playwright-test/playwright.expect.spec.ts", "should support toBeChecked")]
+        public async Task ShouldSupportToBeChecked()
+        {
+            await Page.SetContentAsync("<input type=checkbox checked></input>");
+            await Expect(Page.Locator("input")).ToBeCheckedAsync();
+
+            await Expect(Page.Locator("input")).ToBeCheckedAsync(new() { Checked = true });
+            await Expect(Page.Locator("input")).Not.ToBeCheckedAsync(new() { Checked = false });
+
+            var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => Expect(Page.Locator("input")).ToBeCheckedAsync(new() { Checked = false, Timeout = 300 }));
+            StringAssert.Contains("Locator expected not to be checked", exception.Message);
+            StringAssert.Contains("LocatorAssertions.ToBeCheckedAsync with timeout 300ms", exception.Message);
+
+            exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => Expect(Page.Locator("input")).Not.ToBeCheckedAsync(new() { Timeout = 300 }));
+            StringAssert.Contains("Locator expected not to be checked", exception.Message);
+            StringAssert.Contains("LocatorAssertions.ToBeCheckedAsync with timeout 300ms", exception.Message);
+        }
+
+        [PlaywrightTest("playwright-test/playwright.expect.spec.ts", "should support toBeEditable, toBeEnabled, toBeDisabled, toBeEmpty")]
+        public async Task ShouldSupportToBeEditableToBeEnabledToBeDisabledToBeEmpty()
+        {
+            {
+                await Page.SetContentAsync("<input></input>");
+                await Expect(Page.Locator("input")).ToBeEditableAsync();
+            }
+
+            {
+                await Page.SetContentAsync("<button>Text</button>");
+                await Expect(Page.Locator("button")).ToBeEnabledAsync();
+            }
+
+            {
+                await Page.SetContentAsync("<button disabled>Text</button>");
+                var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => Expect(Page.Locator("button")).ToBeEnabledAsync(new() { Timeout = 500 }));
+                StringAssert.Contains("Locator expected to be enabled", exception.Message);
+                StringAssert.Contains("LocatorAssertions.ToBeEnabledAsync with timeout 500ms", exception.Message);
+            }
+
+            {
+                await Page.SetContentAsync("<button disabled>Text</button>");
+                var locator = Page.Locator("button");
+                await locator.EvaluateAsync("e => setTimeout(() => { e.removeAttribute('disabled') }, 500);");
+                await Expect(locator).ToBeEnabledAsync();
+            }
+
+            {
+                await Page.SetContentAsync("<button>Text</button>");
+                var locator = Page.Locator("button");
+                await locator.EvaluateAsync("e => setTimeout(() => { e.setAttribute('disabled', '') }, 500);");
+                await Expect(locator).Not.ToBeEnabledAsync();
+            }
+
+            {
+                await Page.SetContentAsync("<button disabled>Text</button>");
+                var locator = Page.Locator("button");
+                await Expect(locator).ToBeDisabledAsync();
+            }
+
+            {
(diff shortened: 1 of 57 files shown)


Meir017:
Might be out of scope of this but i think adding https://www.fluentassertions.com/ asserts would be really nice here

mxschmitt:
> Might be out of scope of this but i think adding https://www.fluentassertions.com/ asserts would be really nice here

Most likely no, since our assertions are a bit special, they are asynchronous and retry automatically we'll go with an expect like approach for now.

Meir017:
@mxschmitt fair enough, I guess it would just be an Extension library like any other.
something like this:
```cs
// core
await Expect(Page.Locator("foo")).ToBeVisibleAsync();

// fluentassertions version
await Page.Locator("foo").Should().BeVisibleAsync();
```

src/Playwright/Core/AssertionsBase.cs
```diff
@@ -0,0 +1,121 @@
...
+                message = message.Replace("expected to", "expected not to");
+            var result = await ActualLocator.ExpectAsync(expression, expectOptions).ConfigureAwait(false);
+            if (result.Matches == IsNot)
+            {
+                var actual = result.Received;
+                var log = string.Join("\n", result.Log);
+                if (!string.IsNullOrEmpty(log))
+                    log = "\nCall log:\n" + log;
+                if (expected == null)
+                    throw new PlaywrightException($"{message} {log}");
```
kblok, line 74:
```suggestion
                if (expected == null)
                {
                    throw new PlaywrightException($"{message} {log}");
                }
```

src/Playwright/Core/AssertionsBase.cs
```diff
@@ -0,0 +1,121 @@
...
+                expectOptions.Timeout = 5_000;
+            if (expectOptions.IsNot)
+                message = message.Replace("expected to", "expected not to");
+            var result = await ActualLocator.ExpectAsync(expression, expectOptions).ConfigureAwait(false);
+            if (result.Matches == IsNot)
+            {
+                var actual = result.Received;
+                var log = string.Join("\n", result.Log);
+                if (!string.IsNullOrEmpty(log))
+                    log = "\nCall log:\n" + log;
```
kblok, line 72:
```suggestion
                if (!string.IsNullOrEmpty(log))
                {
                    log = "\nCall log:\n" + log;
                }
```

src/Playwright/Core/AssertionsBase.cs
```diff
@@ -0,0 +1,121 @@
...
+                if (expected == null)
+                    throw new PlaywrightException($"{message} {log}");
+                throw new PlaywrightException($"{message} '{FormatValue(expected)}'\nBut was: '{FormatValue(actual)}' {log}");
+            }
+        }
+
+        internal ExpectedTextValue ExpectedRegex(Regex pattern, ExpectedTextValue options = null)
+        {
+            if (pattern == null)
+                throw new ArgumentNullException(nameof(pattern));
```
kblok, line 82:
```suggestion
            if (pattern == null)
            {
                throw new ArgumentNullException(nameof(pattern));
            }
```

kblok: Don't we have a rule for this? 🤓

src/Playwright/Core/Frame.cs
```diff
@@ -561,6 +561,8 @@ public Task WaitForURLAsync(Func<string, bool> url, FrameWaitForURLOptions optio
         public Task DragAndDropAsync(string source, string target, FrameDragAndDropOptions options = null)
             => _channel.DragAndDropAsync(source, target, options?.Force, options?.NoWaitAfter, options?.Timeout, options?.Trial, options?.Strict);
 
+        internal Task<FrameExpectResult> ExpectAsync(string selector, string expression, FrameExpectOptions options = null) => _channel.ExpectAsync(selector, expression, expressionArg: options?.ExpressionArg, expectedText: options?.ExpectedText, expectedNumber: options?.ExpectedNumber, expectedValue: options?.ExpectedValue, useInnerText: options?.UseInnerText, isNot: options?.IsNot, timeout: options?.Timeout);
```
kblok, line 564:
```suggestion
        internal Task<FrameExpectResult> ExpectAsync(string selector, string expression, FrameExpectOptions options = null) 
            => _channel.ExpectAsync(selector, expression, expressionArg: options?.ExpressionArg, expectedText: options?.ExpectedText, expectedNumber: options?.ExpectedNumber, expectedValue: options?.ExpectedValue, useInnerText: options?.UseInnerText, isNot: options?.IsNot, timeout: options?.Timeout);
```

src/Playwright.Tests/Assertions/PageAssertionsTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        [PlaywrightTest("playwright-test/playwright.expect.misc.spec.ts", "should support toHaveURL")]
+        public async Task ShouldSupportToHaveURLAsync()
+        {
+            await Page.GotoAsync("data:text/html,<div>A</div>");
+            await Expect(Page).ToHaveURLAsync("data:text/html,<div>A</div>");
+
+            await Page.GotoAsync("data:text/html,<div>B</div>");
+            var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => Expect(Page).ToHaveURLAsync("wrong", new() { Timeout = 100 }));
```
yury-s, line 62: Do we now have both PlaywrightAssert and Playwright.Assert?  If so let's merge former into the latter.

src/Playwright/Core/LocatorAssertions.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        public Task ToContainTextAsync(Regex expected, LocatorAssertionsToContainTextOptions options = null) =>
+            ExpectImplAsync("to.have.text", ExpectedRegex(expected, new() { MatchSubstring = true, NormalizeWhiteSpace = true }), expected, "Locator expected to contain regex", ConvertToFrameExpectOptions(options));
+
+        public Task ToContainTextAsync(IEnumerable<string> expected, LocatorAssertionsToContainTextOptions options = null) =>
+            ExpectImplAsync("to.contain.text.array", expected.Select(text => new ExpectedTextValue() { String = text, MatchSubstring = true, NormalizeWhiteSpace = true }).ToArray(), expected, "Locator expected to contain text", ConvertToFrameExpectOptions(options));
+
+        public Task ToContainTextAsync(IEnumerable<Regex> expected, LocatorAssertionsToContainTextOptions options = null) =>
+            ExpectImplAsync("to.contain.text.array", expected.Select(regex => ExpectedRegex(regex, new() { MatchSubstring = true, NormalizeWhiteSpace = true })).ToArray(), expected, "Locator expected to contain regex", ConvertToFrameExpectOptions(options));
```
yury-s, line 77: nit: ... to contain text matching regex?

yury-s (approved):

src/Playwright.Tests/Assertions/LocatorAssertionsTests.cs
```diff
@@ -0,0 +1,312 @@
...
+        {
+            {
+                await Page.SetContentAsync("<input></input>");
+                await Expect(Page.Locator("input")).ToBeEditableAsync();
+            }
+
+            {
+                await Page.SetContentAsync("<button>Text</button>");
+                await Expect(Page.Locator("button")).ToBeEnabledAsync();
+            }
```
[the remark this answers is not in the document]
mxschmitt: We use them to easier visualise how they relate to upstream tests.

src/Playwright.Tests/Assertions/PageAssertionsTests.cs
mxschmitt: PlaywrightAssert is only for us internally, its located inside the Playwright.Tests package: https://cs.github.com/microsoft/playwright-dotnet/blob/82710bccaa67eb4a0bd1035ee332c676398e015d/src/Playwright.Tests/BaseTests/PlaywrightAssert.cs#L32

src/Playwright/Transport/Protocol/ExpectedTextValue.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.Playwright.Transport.Protocol
+{
+    internal class ExpectedTextValue
+    {
+        [JsonPropertyName("string")]
+        public string String { get; set; }
+
+        [JsonPropertyName("regexSource")]
+        public string RegexSource { get; set; }
```
nohwnd, line 35: Are default serializer settings used here? Then you are duplicating what your serializer already does for you. 

https://github.com/microsoft/playwright-dotnet/blob/8a4756e820dcb295ba0762a95a2b2a6f197dd415/src/Playwright/Helpers/JsonExtensions.cs#L90-L99


Docs with example:

https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-customize-properties?pivots=dotnet-5-0#use-camel-case-for-all-json-property-names

src/Playwright/Transport/Channels/FrameChannel.cs
```diff
@@ -700,5 +701,29 @@ internal Task DragAndDropAsync(string source, string target, bool? force, bool?
 
             return Connection.SendMessageToServerAsync(Guid, "dragAndDrop", args);
         }
+
+        internal async Task<FrameExpectResult> ExpectAsync(string selector, string expression, object expressionArg, ExpectedTextValue[] expectedText, int? expectedNumber, object expectedValue, bool? useInnerText, bool? isNot, float? timeout)
+        {
+            var args = new Dictionary<string, object>
+            {
+                ["selector"] = selector,
```
nohwnd, line 709: this could be using `[nameof(selector)] = selector` if you plan to keep the parameter names consistent with those keys.

src/Playwright/Core/PageAssertions.cs
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+using System;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Microsoft.Playwright.Transport.Protocol;
+
+namespace Microsoft.Playwright.Core
+{
+    public class PageAssertions : AssertionsBase, IPageAssertions
```
nohwnd, line 32: Not sure about what your other classes do, or if you are concerned about having too many public apis, but this could be kept internal when you have interface for it, because you always use it through that. I see the interfaces are generated, otherwise I wouldn't use them.

Same comment for other assertion classes.

src/Playwright/Core/Locator.cs
```diff
@@ -251,6 +249,14 @@ ILocator ILocator.Locator(string selector, LocatorLocatorOptions options)
         public Task WaitForAsync(LocatorWaitForOptions options = null)
             => _frame.LocatorWaitForAsync(_selector, ConvertOptions<LocatorWaitForOptions>(options));
 
+        internal Task<FrameExpectResult> ExpectAsync(string expression, FrameExpectOptions options = null)
```
nohwnd, line 252: nit: use the same syntax `=>` and formatting as above?

src/Playwright/Transport/Protocol/ExpectedTextValue.cs
mxschmitt: cool, TIL.

src/Playwright/Transport/Channels/FrameChannel.cs
mxschmitt: I'll keep it as it is for now, since it's then more clear. But interesting suggestion!

src/Playwright/Core/AssertionsBase.cs
mxschmitt: https://github.com/microsoft/playwright-dotnet/issues/2024

PR #2016, opened 2022-02-16
devops: continue running tests with .NET 5 for now

mxschmitt:


pavelfeldman:
Hey Max, why the change? We should definitely run tests under .NET 6 as well!

OgreTransporter:
I am in favor of replacing .NET5 with .NET6. EOL of .NET5 is May 8, 2022.

PR #2018, opened 2022-02-18
chore: drop Macross.Json.Extensions from dependencies

mxschmitt:


campersau:
Would you accept a PR which removes this and adds a custom enumconverter like I did here https://github.com/campersau/playwright-dotnet/commit/4d9c4109b9e7b401f7921196d32b2bcb1ecd83e1 ?

PR #2020, opened 2022-02-19
chore: drop Macross.Json.Extensions from dependencies

### LINKED ISSUE #2019: Remove package dependency on Macross.Json.Extensions
package v1.18.1 seems to have an unnecessary dependency on `Macross.Json.Extensions` which seems to be a non-microsoft third party package that is not used.

the reference is suspect and hopefully not a supply chain attack of some sort.

campersau:
Now using a custom `JsonStringEnumMemberConverter` which calls the existing `ToValueString` helper and adds a corresponding `FromValueString`.

Fixes #2019

### DIFF
--- a/src/Playwright/Transport/Converters/JsonStringEnumMemberConverter.cs
+++ b/src/Playwright/Transport/Converters/JsonStringEnumMemberConverter.cs
@@ -1,0 +1,110 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Playwright.Helpers;
+
+namespace Microsoft.Playwright.Transport.Converters
+{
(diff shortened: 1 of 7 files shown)


src/Playwright/Transport/Converters/JsonStringEnumMemberConverter.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            if (reader.TokenType == JsonTokenType.Null)
+            {
+                return null;
+            }
+            else if (reader.TokenType == JsonTokenType.String)
+            {
+                return EnumHelper.FromValueString<TEnum>(reader.GetString());
+            }
+            else if (reader.TokenType == JsonTokenType.Number)
```
campersau, line 58: This code currently is only needed because of this test
https://github.com/microsoft/playwright-dotnet/blob/228903102fb53a218c452aee47d7ce713b2dcd90/src/Playwright.Tests/BrowserContextStorageStateTests.cs#L98
as the returned JSON uses an integer for the `SameSiteAttribute` enum value instead of a string as everywhere else.

kblok (commented):
💯

src/Playwright/Transport/Converters/JsonStringEnumMemberConverter.cs
mxschmitt: Sorry for the late reply. Don't we use the. Enum converter at all in StorageStateAsync?

mxschmitt: Happy to merge but I think the enums are mostly used in user places rather than StorageStateAsync.

campersau: @mxschmitt Yes, the converter isn't used there which is different to what JS does. But that can be fixed separatly by just using the `JsonExtensions.DefaultJsonSerializerOptions`
https://github.com/microsoft/playwright-dotnet/blob/e604ff9ca8d24755c5f137a976cb4b1926f149cd/src/Playwright/Helpers/JsonExtensions.cs#L41

mxschmitt: Ah you mean that in JS it returns it "parsed" and in .NET and Java as a string. Thats expected I think for now and would be a breaking change otherwise.

campersau: No, I mean that in JS you get the enum value as string whereas in dotnet you get it as a number.

dotnet
```c#
var playwright = await Microsoft.Playwright.Playwright.CreateAsync();
await page.GotoAsync("https://google.de");

System.Console.WriteLine(await page.Context.StorageStateAsync());
```
Returns the following (note the `"sameSite":1`):
```json
{"cookies":[{"name":"CONSENT","value":"PENDING\u002B096","url":null,"domain":".google.de","path":"/","expires":1.7092998E+09,"httpOnly":false,"secure":true,"sameSite":1}],"origins":[]}
```

---
JS
```js
import { chromium } from 'playwright';
(async () => {
    const browser = await chromium.launch();
    const page = await browser.newPage();
    await page.goto('https://google.de');

    console.log(JSON.stringify(await page.context().storageState()));

    await browser.close();
})();
```
Returns the following (note the `"sameSite":"Lax"`):
```json
{"cookies":[{"sameSite":"Lax","name":"CONSENT","value":"PENDING+769","domain":".google.de","path":"/","expires":1709300312.782131,"httpOnly":false,"secure":true}],"origins":[]}
```

That's because in dotnet the `JsonStringEnumMemberConverter` is missing.

It should just use the `JsonExtensions.DefaultJsonSerializerOptions` then the `"url":null` wouldn't also be there I guess.

mxschmitt: I've created https://github.com/microsoft/playwright-dotnet/issues/2035.👍

PR #2033, opened 2022-02-28
chore: force brackets for if conditions

mxschmitt:
I don't know if we should do this.


pavelfeldman:
I have no preference here!

PR #2039, opened 2022-03-03
use DefaultJsonSerializerOptions in StorageStateAsync

### LINKED ISSUE #2035: StorageState serialization seems wrong
No, I mean that in JS you get the enum value as string whereas in dotnet you get it as a number.

dotnet
```c#
var playwright = await Microsoft.Playwright.Playwright.CreateAsync();
await page.GotoAsync("https://google.de");

System.Console.WriteLine(await page.Context.StorageStateAsync());
```
Returns the following (note the `"sameSite":1`):
```json
{"cookies":[{"name":"CONSENT","value":"PENDING\u002B096","url":null,"domain":".google.de","path":"/","expires":1.7092998E+09,"httpOnly":false,"secure":true,"sameSite":1}],"origins":[]}
```

---
JS
```js
import { chromium } from 'playwright';
(async () => {
    const browser = await chromium.launch();
    const page = await browser.newPage();
    await page.goto('https://google.de');

    console.log(JSON.stringify(await page.context().storageState()));

    await browser.close();
})();
```
Returns the following (note the `"sameSite":"Lax"`):
```json
{"cookies":[{"sameSite":"Lax","name":"CONSENT","value":"PENDING+769","domain":".google.de","path":"/","expires":1709300312.782131,"httpOnly":false,"secure":true}],"origins":[]}
```

That's because in dotnet the `JsonStringEnumMemberConverter` is missing.

It should just use the `JsonExtensions.DefaultJsonSerializerOptions` then the `"url":null` wouldn't also be there I guess.

_Originally posted by @campersau in https://github.com/microsoft/playwright-dotnet/pull/2020#discussion_r817705453_

campersau:
which serializes enum values to strings and does not serialize null values

Fixes #2035

### DIFF
--- a/src/Playwright.Tests/BrowserContextStorageStateTests.cs
+++ b/src/Playwright.Tests/BrowserContextStorageStateTests.cs
@@ -130,7 +130,14 @@ namespace Microsoft.Playwright.Tests
             var storageState = await Context.StorageStateAsync();
             StringAssert.Contains(@"""name"":""a"",""value"":""b""", storageState);
             StringAssert.Contains(@"""name"":""empty"",""value"":""""", storageState);
-            StringAssert.Contains(@"""sameSite"":""Lax""", storageState);
+            if (TestConstants.IsWebKit)
+            {
+                StringAssert.Contains(@"""sameSite"":""None""", storageState);
+            }
+            else
+            {
+                StringAssert.Contains(@"""sameSite"":""Lax""", storageState);
+            }
             StringAssert.DoesNotContain(@"""url"":null", storageState);
 
             await using var context2 = await Browser.NewContextAsync(new() { StorageState = storageState });
--- a/src/Playwright.Tests/BrowserContextStorageStateTests.cs
+++ b/src/Playwright.Tests/BrowserContextStorageStateTests.cs
@@ -109,5 +109,38 @@ namespace Microsoft.Playwright.Tests
             Assert.AreEqual("value1", await page2.EvaluateAsync<string>("localStorage['name1']"));
             Assert.AreEqual("username=John Doe", await page2.EvaluateAsync<string>("document.cookie"));
         }
+
+        [PlaywrightTest("browsercontext-storage-state.spec.ts", "should capture cookies")]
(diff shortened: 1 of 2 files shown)


src/Playwright.Tests/BrowserContextStorageStateTests.cs
```diff
@@ -109,5 +109,38 @@ public async Task ShouldRoundTripThroughTheFile()
...
+            {
+                const cookies = document.cookie.split(';');
+                return cookies.map(cookie => cookie.trim()).sort();
+            }"));
+
+            var storageState = await Context.StorageStateAsync();
+            StringAssert.Contains(@"""name"":""a"",""value"":""b""", storageState);
+            StringAssert.Contains(@"""name"":""empty"",""value"":""""", storageState);
+            StringAssert.Contains(@"""sameSite"":""Lax""", storageState);
+            StringAssert.DoesNotContain(@"""url"":null", storageState);
```
campersau, line 134: Added these two asserts to check for the behavior change which are not included upstream: https://github.com/microsoft/playwright/blob/71a4f69acf3e827402ba208939c677ea4a38bca4/tests/browsercontext-storage-state.spec.ts#L107-L143

mxschmitt:
~I'm still curious how this internally has happened. I did not introduce the code, but for me I thought we create a new options `JsonSerializerOptions`) which only have `camelCase` internally to reflect a similar JSON behaviour like JavaScript is having. Then the JSON serialisation and deserialisation should not be affected by some enum converter.~

~Or is this because it internally inside the connection or channel somewhere already got transferred?~

Okay the reason is because inside [our channel](https://github.com/microsoft/playwright-dotnet/blob/98db95b3b279f8654674d7434479625bb5d8cc76/src/Playwright/Transport/Channels/BrowserContextChannel.cs#L234-L235) we already parse it into a StorageState instance which is enum based. Then we try to deserialise it with the default json serializor which does not work with enums. Makes sense.

mxschmitt (approved):

mxschmitt:
Thank you very much @campersau

PR #2041, opened 2022-03-03
chore: support linux arm64

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2053

### DIFF
--- a/src/Playwright/build/Microsoft.Playwright.targets
+++ b/src/Playwright/build/Microsoft.Playwright.targets
@@ -5,13 +5,18 @@
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">%(RuntimePack.RuntimeIdentifier)</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">$(RuntimeIdentifier)</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">$(RuntimeIdentifiers)</PlaywrightPlatform>
-      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux'))">linux</PlaywrightPlatform>
+      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">linux-x64</PlaywrightPlatform>
+      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'Arm64'">linux-arm64</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Windows'))">win</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('OSX'))">osx</PlaywrightPlatform>
     </PropertyGroup>
     <ItemGroup Condition="'$(PlaywrightPlatform)' != 'none'">
-      <_PlaywrightCopyItems Include="$(MSBuildThisFileDirectory)..\.playwright\node\linux\**" Condition="$(PlaywrightPlatform.Contains('linux'))">
-        <PlaywrightFolder>node\linux\</PlaywrightFolder>
+       <!-- 'linux' to stay backwards compatible -->
+      <_PlaywrightCopyItems Include="$(MSBuildThisFileDirectory)..\.playwright\node\linux\**" Condition="$(PlaywrightPlatform.Contains('linux')) OR $(PlaywrightPlatform.Contains('linux-x64'))">
+        <PlaywrightFolder>node\linux-x64\</PlaywrightFolder>
+      </_PlaywrightCopyItems>
+      <_PlaywrightCopyItems Include="$(MSBuildThisFileDirectory)..\.playwright\node\linux-arm64\**" Condition="$(PlaywrightPlatform.Contains('linux-arm64'))">
+        <PlaywrightFolder>node\linux-arm64\</PlaywrightFolder>
       </_PlaywrightCopyItems>
       <_PlaywrightCopyItems Include="$(MSBuildThisFileDirectory)..\.playwright\node\win32_x64\**" Condition="$(PlaywrightPlatform.Contains('win'))">
         <PlaywrightFolder>node\win32_x64\</PlaywrightFolder>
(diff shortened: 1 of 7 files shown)


utils/docker/Dockerfile.focal
```diff
@@ -30,7 +30,7 @@ RUN mkdir /ms-playwright && \
     cd /ms-playwright-agent && \
     dotnet new console && \
     echo '<?xml version="1.0" encoding="utf-8"?><configuration><packageSources><add key="local" value="/tmp/"/></packageSources></configuration>' > nuget.config && \
-    dotnet add package Microsoft.Playwright && \
+    dotnet add package Microsoft.Playwright --prerelease && \
```
mxschmitt, line 33: when our current version was a e.g. next-1 version it didn't install the expected version inside Docker. This fixes it.

src/Playwright/build/Microsoft.Playwright.targets
```diff
@@ -3,14 +3,18 @@
...
-      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux'))">linux</PlaywrightPlatform>
+      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">linux</PlaywrightPlatform>
+      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'Arm64'">linux-arm64</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Windows'))">win</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('OSX'))">osx</PlaywrightPlatform>
     </PropertyGroup>
     <ItemGroup Condition="'$(PlaywrightPlatform)' != 'none'">
       <_PlaywrightCopyItems Include="$(MSBuildThisFileDirectory)..\.playwright\node\linux\**" Condition="$(PlaywrightPlatform.Contains('linux'))">
```
campersau, line 12: `linux-arm64` includes `linux` so on linux it there will be both folders.

src/Playwright/build/Microsoft.Playwright.targets
```diff
@@ -3,14 +3,18 @@
   <Target Name="CopyPlaywrightFilesToOutput" AfterTargets="BeforeBuild">
     <PropertyGroup>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">%(RuntimePack.RuntimeIdentifier)</PlaywrightPlatform>
-      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux'))">linux</PlaywrightPlatform>
+      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">linux</PlaywrightPlatform>
```
campersau, line 6: Maybe use this instead?
```suggestion
      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' != 'Arm64'">linux</PlaywrightPlatform>
```

src/Playwright/build/Microsoft.Playwright.targets
mxschmitt: yup, not sure if there is a good way around that without breaking people. (Very good catch, actually wanted to add a comment for it in this minute)

src/Playwright/build/Microsoft.Playwright.targets
mxschmitt: Not sure, better to have no driver at all rather than having one for a different architecture on e.g. arm (e.g. normal arm). wdyt?

campersau: The current behavior is that the linux one will be chosen regardless of the architecture so it is "backwardcompatible" 😄

mxschmitt: but having a not functioning driver since you can't run a driver for a different arch is also bad. 😄

src/Playwright/build/Microsoft.Playwright.targets
mxschmitt: I'm considering renaming linux to linux-amd64, not sure how many people it would break and if semver would allow it. Probably not, but the feature with specifying the arch is pretty new and most of the people probably "let the driver version autoselect".

campersau: Note that dotnet / msbuild uses `linux-x64` instead: https://docs.microsoft.com/dotnet/core/rid-catalog#linux-rids

mxschmitt: yeah I'd vote to use these rids instead of our custom values which we added a few weeks ago (linux, mac, windows).

campersau: We added `linux` / `win` / `osx` which are valid dotnet / msbuild values. Except for `none` which is the only custom one we currently use.
Specifiying the architecture is optional, see https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json:

```
    win-x64  |  win-x86  |  ...
          \     /
            win
             |
            any
```
```
    linux-x64  |  linux-arm64  |  ...
          \     /
           linux
             |
            any
```
```
    osx-x64  |  osx-arm64  |  ...
          \     /
            osx
             |
            any
```

mxschmitt: thanks!

src/Playwright/build/Microsoft.Playwright.targets
```diff
@@ -5,13 +5,18 @@
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">%(RuntimePack.RuntimeIdentifier)</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">$(RuntimeIdentifier)</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">$(RuntimeIdentifiers)</PlaywrightPlatform>
-      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux'))">linux</PlaywrightPlatform>
+      <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('Linux')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">linux-x64</PlaywrightPlatform>
```
yury-s, line 8: nit: should the values be x64 and arm64 respectively?

yury-s (approved):

src/Playwright/build/Microsoft.Playwright.targets
mxschmitt: These are the values if no PlaywrightPlatform is set, then its automatically determined from the host OS.

PR #2060, opened 2022-03-10
devops: various fixes and enhancements

mxschmitt:
The Docker pipeline was broken. This fixes it.

### DIFF
--- a/.github/workflows/test_docker.yml
+++ b/.github/workflows/test_docker.yml
@@ -3,6 +3,8 @@ on:
   push:
     paths:
       - '.github/workflows/test_docker.yml'
+      - 'src/Common/Version.props'
+      - '**/*.csproj'
     branches:
       - main
   pull_request:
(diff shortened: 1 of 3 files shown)


utils/docker/Dockerfile.focal
```diff
@@ -30,7 +30,7 @@ RUN mkdir /ms-playwright && \
     cd /ms-playwright-agent && \
     dotnet new console && \
     echo '<?xml version="1.0" encoding="utf-8"?><configuration><packageSources><add key="local" value="/tmp/"/></packageSources></configuration>' > nuget.config && \
-    dotnet add package Microsoft.Playwright && \
+    dotnet add package Microsoft.Playwright --prerelease && \
```
mxschmitt, line 33: the line above means: "hey use the /tmp/ directory as a package source". This line installs playwright, but since the local version is sometimes a pre-release, it did not use it, it went to nuget.org instead. That change fixes it so its picking up prereleases too.

.github/workflows/tests.yml
```diff
@@ -39,12 +39,12 @@ jobs:
         if: matrix.os != 'ubuntu-latest'
         env:
           BROWSER: ${{ matrix.browser }}
-        run: dotnet test ./src/Playwright.Tests/Playwright.Tests.csproj -c Debug
+        run: dotnet test ./src/Playwright.Tests/Playwright.Tests.csproj -c Debug -f net6.0
```
mxschmitt, line 42: before tests were running on netstandard2.0 and net6.0. We test it on a different pipeline, so lets revert this how it previously was.

.github/workflows/test_docker.yml
```diff
@@ -19,16 +21,19 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 5.0.x
+        dotnet-version: |
```
yury-s, line 24: Does it mean we'll run with each of them ?

yury-s (approved):

.github/workflows/test_docker.yml
mxschmitt: no. its just what's required to run them. 3.1 is required because our test-server is ASP.Net based as of today and it requires 3.1.

PR #2072, opened 2022-03-17
chore: do not redirect stdout/stderr in CLI

mxschmitt:
Current state/implementation (ToT) when invoking the CLI:

```
CreateNoWindow = true,
RedirectStandardError = true,
RedirectStandardInput = true,
RedirectStandardOutput = true,
```

(properties when calling the Playwright CLI via our ps1 script or Playwright CLI.

This ends up that the stdout/stderr stream does end up in `OutputDataReceived` and `ErrorDataReceived` which has the downside that `isTTY` on `Node.js` side is `false`. -> No progress bar when downloading browsers.

New implementation:

```
CreateNoWindow = false,
RedirectStandardError = false,
RedirectStandardInput = false,
RedirectStandardOutput = false,
```

(leaving defaults there so its easier understandable). When specifying these values to false, `they` get automatically written to `Process.StandardOutput` and friends. Thats what we want. On Windows this does only work when `CreateNoWindow=false`, after testing I saw no issue with it keeping it on false (tested on Windows).

See here for reference: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.redirectstandardoutput?view=net-6.0

### DIFF
--- a/src/Playwright.Tests/CLITests.cs
+++ b/src/Playwright.Tests/CLITests.cs
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 
-using System;
+using System.Diagnostics;
 using System.IO;
 using Microsoft.Playwright.NUnit;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


nohwnd:
Running processes correctly and grabbing their output is a bit fiddly on .NET. But mostly I am not sure about the intent here, or how exactly Playwright works. 

 I don't like that this changes the behavior for everyone that uses Playwright.Program.Main, is "everybody" using that? How is this done in normal tests that are not testing the cli? Are they using Playwright.Program.Main under the cover, or some other way of starting up?

---

What I see is happening now is that: 

**Old:** 
You start your test. Test calls to Playwright.Program.Main (in-process), that resolves playwright.cmd (or playwright.sh) via Driver which calls node.exe with the provided parameters. So there are 2 process hops (dotnet -> cmd -> node.exe).

**In the new approach:** 

You start your test, that calls ExecutePlaywrightPs1 (in-process), which calls pwsh.exe that calls Playwright.Program.Main (in-process) -> that resolves playwright.cmd (or playwright.sh) via Driver which calls node.exe with the provided parameters.

So there are 4 process hops dotnet -> pwsh -> cmd -> node.exe.

Additionally pwsh will itself on startup, and will load user profile, which can take few ms, or few seconds (you did not specify -noprofile). So using powershell as a glorified dotnet host resolver is a bit costly. 

**Ideally:** 

You would only do 1 hop, dotnet -> node.exe. But maybe this issue is only impacting these CLI tests, and normally you start node only 1 time, so the overhead of calling also cmd is tiny.

dgozman (approved):

Issue #2064, opened 2022-03-15
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:

PR #2089, opened 2022-04-08
chore: roll Playwright to 1.21.0-beta-1649396150000

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2064

### DIFF
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -372,25 +372,7 @@ namespace Microsoft.Playwright.Transport
                 return new TimeoutException(error.Message);
             }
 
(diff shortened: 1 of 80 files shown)


src/Playwright/Transport/Connection.cs
```diff
@@ -372,25 +372,7 @@ private Exception CreateException(PlaywrightServerError error)
                 return new TimeoutException(error.Message);
             }
 
-            if (error.Message.Contains("Browser closed") || error.Message.Contains("Target closed") || error.Message.Contains("The page has been closed."))
```
yury-s, line 375: Why is it changing?

yury-s (approved):

src/Playwright/Transport/Connection.cs
mxschmitt: It was like logic which did not have any effect. Drive-by cleanup.

PR #2099, opened 2022-04-14
Use ItemGroup for copying playwright.ps1

### LINKED ISSUE #2098: [Question]: run playwright.ps1 after `dotnet publish`
Hi,

I'm trying to run the playwright.ps1 instead of the Microsoft.Playwright.CLI.

We create an artifact of the published application (with the tests) and run that. We do something like: `dotnet publish --ouput .dist --configuration Release --no-restore)`. From that artifact we create an docker container. 

In the publish there is no playwright.ps1 file. I tried copying that, but it failed when running as it probably needs more files.

How could I fix that the  playwright.ps1 is in my publish folder? I prefer no to create an artifact of the bin folder. 

PS: we hoped we did not need the playwright.ps1 when using `-p:PlaywrightPlatform=linux` on `dotnet build`. But we get sometimes, but not always, *Looks like Playwright Test or Playwright was just installed or updated.*

campersau:
which copies files correctly when using `publish`.

- `CopyRuntimeConfigToOutput` is no longer needed as there is no runtime config for `netstandard2.0`. (It was needed for NET 3.1/5/6 etc.)
- `PlaywrightBuildCleanup` is no longer needed because `Content` files are automatically part of the `clean` target.
- `PlaywrightPlatform` does not affect `playwright.ps1`.

Fixes #2098

### DIFF
--- a/src/Playwright/build/Microsoft.Playwright.targets
+++ b/src/Playwright/build/Microsoft.Playwright.targets
@@ -35,25 +35,13 @@
         <Visible>false</Visible>
         <Pack>false</Pack>
       </Content>
+      <Content Include="$(MSBuildThisFileDirectory)playwright.ps1">
+        <Link>playwright.ps1</Link>
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        <PublishState>Included</PublishState>
+        <Visible>false</Visible>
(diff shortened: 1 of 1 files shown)


mxschmitt:
Thank you @campersau!

PR #2119, opened 2022-04-28
chore: fix updating JSHandle previews

mxschmitt:
This fixes a flaky firefox test and fixes updating previews for JSHandles what the test was testing.


src/Playwright/Transport/Channels/PreviewUpdatedEventArgs.cs
```diff
@@ -1,33 +0,0 @@
...
- * SOFTWARE.
- */
-
-using System;
-
-namespace Microsoft.Playwright.Transport.Channels
-{
-    internal class PreviewUpdatedEventArgs : EventArgs
-    {
-        public string Preview { get; set; }
```
mxschmitt, line 31: We tend to drop this EventArgs classes for simple data structures to reduce the complexity.

mxschmitt:
can't repro anymore what I was actually fixing, so closing for now.

PR #2121, opened 2022-05-02
chore: do not rely on auto camel case for channel json initializers

mxschmitt:


kblok:
What property was the one breaking?

mxschmitt:
> What property was the one breaking?

APIRequestContext

dgozman (approved):

PR #2142, opened 2022-05-13
chore: roll Playwright to 1.22

mxschmitt:


src/Playwright.Tests/PageEvaluateTests.cs
```diff
@@ -388,18 +386,6 @@ public async Task ShouldWorkForCircularObject()
             Assert.NotNull(result);
         }
 
-        [PlaywrightTest("page-evaluate.spec.ts", "should be able to throw a tricky error")]
-        [Ignore("v1.22 blocker")]
-        public async Task ShouldBeAbleToThrowATrickyError()
```
dgozman, line 393: What's up with this one?

src/Playwright.Tests/PageEvaluateTests.cs
```diff
@@ -633,11 +619,10 @@ private enum TestEnum
         }
 
         [PlaywrightTest(Description = "https://github.com/microsoft/playwright-dotnet/issues/1706")]
-        [Ignore("v1.22 blocker")] // actually this should not be possible what the test tries to achieve. No access to the internal data structure!
         public async Task ShouldNotReturnDisposedJsonElement()
         {
             var result = await Page.EvaluateAsync<JsonElement?>("()=> [{a:1,b:2},{a:1,b:2}]");
-            Assert.AreEqual("[{\"o\":[{\"k\":\"a\",\"v\":{\"n\":1}},{\"k\":\"b\",\"v\":{\"n\":2}}]},{\"o\":[{\"k\":\"a\",\"v\":{\"n\":1}},{\"k\":\"b\",\"v\":{\"n\":2}}]}]", result.ToString());
+            Assert.AreEqual("[{\"o\":[{\"k\":\"a\",\"v\":{\"n\":1}},{\"k\":\"b\",\"v\":{\"n\":2}}],\"id\":2},{\"o\":[{\"k\":\"a\",\"v\":{\"n\":1}},{\"k\":\"b\",\"v\":{\"n\":2}}],\"id\":3}]", result.ToString());
```
dgozman, line 625: Why do we test internal representation? Let's remove this test.

src/Playwright.Tests/SelectorMiscTests.cs
```diff
@@ -99,102 +99,51 @@ public async Task ShouldWorkWithLayoutSelectors()
...
-            Assert.AreEqual(await Page.EvalOnSelectorAllAsync<string>("div >> near=\"#id7\"", "els => els.map(e => e.id).join(',')"), "id0,id5,id3,id6");
             Assert.AreEqual(await Page.EvalOnSelectorAllAsync<string>("div:near(#id0)", "els => els.map(e => e.id).join(',')"), "id3,id6,id7,id8,id1,id5");
-            Assert.AreEqual(await Page.EvalOnSelectorAllAsync<string>("div >> near=\"#id0\"", "els => els.map(e => e.id).join(',')"), "id3,id6,id7,id8,id1,id5");
             Assert.AreEqual(await Page.EvalOnSelectorAllAsync<string>("div:near(#id6)", "els => els.map(e => e.id).join(',')"), "id0,id3,id7");
-            Assert.AreEqual(await Page.EvalOnSelectorAllAsync<string>("div >> near=\"#id6\"", "els => els.map(e => e.id).join(',')"), "id0,id3,id7");
             Assert.AreEqual(await Page.EvalOnSelectorAllAsync<string>("div:near(#id6, 10)", "els => els.map(e => e.id).join(',')"), "id0");
-            Assert.AreEqual(await Page.EvalOnSelectorAllAsync<string>("div >> near=\"#id6\",10", "els => els.map(e => e.id).join(',')"), "id0");
```
dgozman, line 189: When porting next time, it is not necessary to port all string selector tests, since for the port a string is just a string after all :)

dgozman (approved):

src/Playwright.Tests/PageEvaluateTests.cs
mxschmitt: We'll follow-up with this one in a different PR.

src/Playwright.Tests/PageEvaluateTests.cs
mxschmitt: The upstream test makes no sense anymore:

```js
it('should be able to throw a tricky error', async ({ page }) => {
  const windowHandle = await page.evaluateHandle(() => window);
  const errorText = await windowHandle.jsonValue().catch(e => e.message); // [0]
  console.log(errorText, typeof errorText);
  const error = await page.evaluate(errorText => {
    throw new Error(errorText);
  }, errorText).catch(e => e);
  expect(error.message).toContain(errorText);
});
```

Previously before the circular reference support, [0] was throwing because `window` had circular references. But now it's not throwing anymore, and it's just `ref: <Window>` which makes the whole test kinda unnecessary. because there are no exception messages to compare with.

PR #2149, opened 2022-05-23
chore: generate object types from protocol

mxschmitt:
Generator changes: https://github.com/microsoft/playwright/pull/14342

### DIFF
--- a/src/Playwright/Core/StorageState.cs
+++ b/src/Playwright/Core/StorageState.cs
@@ -25,6 +25,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Playwright.Transport.Protocol;
 
 namespace Microsoft.Playwright.Core
 {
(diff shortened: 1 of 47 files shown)


src/Playwright/Transport/Channels/PageChannel.cs
```diff
@@ -105,7 +105,7 @@ internal override void OnMessage(string method, JsonElement? serverParams)
                     Popup?.Invoke(this, new() { Page = serverParams?.GetProperty("page").ToObject<PageChannel>(Connection.DefaultJsonSerializerOptions).Object });
                     break;
                 case "pageError":
-                    PageError?.Invoke(this, serverParams?.GetProperty("error").GetProperty("error").ToObject<SerializedError>(Connection.DefaultJsonSerializerOptions));
```
yury-s, line 108: Let's add a test for this.

yury-s (approved):

src/Playwright/Transport/Channels/PageChannel.cs
mxschmitt: there are tests so I discovered that I had to adjust this to make the test pass.

src/Playwright/Core/StorageState.cs
```diff
@@ -38,7 +39,7 @@ internal class StorageState : IEquatable<StorageState>
         /// <summary>
         /// List of local storage per origin.
         /// </summary>
-        public ICollection<StorageStateOrigin> Origins { get; set; } = new List<StorageStateOrigin>();
+        public ICollection<OriginStorage> Origins { get; set; } = new List<OriginStorage>();
```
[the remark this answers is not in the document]
mxschmitt: Its only used internally, externally we expose it only as a string:

image

PR #2158, opened 2022-05-27
chore: add local library playground

mxschmitt:
Testing some actual user code without needing to build the nuget locally and installing it in a non trivial way in a new folder was hard.

We have now a small little playground where we can run user-code which uses the Playwright package right from the source-code.


yury-s (approved):
Maybe convert it into something like examples project? Worked pretty well in java.

Issue #1611, opened 2021-07-11
[Feature]: Provide a .NET Docker image


mxschmitt:
### Feature request

Provide a Docker image to make Playwright for .NET work with Docker out of the box. Similarly to what was done for Playwright for Java.

LGinC:
```dockerfile
FROM mcr.microsoft.com/playwright
RUN apt-get update; \
    apt-get install wget apt-transport-https && \
    wget https://packages.microsoft.com/config/ubuntu/20.10/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
    dpkg -i packages-microsoft-prod.deb && \
    rm packages-microsoft-prod.deb && \
    apt-get update && apt-get install -y dotnet-runtime-5.0 && \
    apt-get clean
WORKDIR /app
COPY . .
ENTRYPOINT ["dotnet", "xxx.dll"]
```
.net 5.0,  emmm.... size of image is 1.8G

h0730303779:
> ```dockerfile
> FROM mcr.microsoft.com/playwright
> RUN apt-get update; \
>     apt-get install wget apt-transport-https && \
>     wget https://packages.microsoft.com/config/ubuntu/20.10/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
>     dpkg -i packages-microsoft-prod.deb && \
>     rm packages-microsoft-prod.deb && \
>     apt-get update && apt-get install -y dotnet-runtime-5.0 && \
>     apt-get clean
> WORKDIR /app
> COPY . .
> ENTRYPOINT ["dotnet", "xxx.dll"]
> ```
> 
> .net 5.0, emmm.... size of image is 1.8G

![image](https://user-images.githubusercontent.com/16299810/132552045-ca2e7d8e-f230-406b-a73e-a2695a1f0dc3.png)

这个dockerfile 能用么 我直接扑街啊.

pablopioli:
The official image is based on Ubuntu 20.04, what I use is

```
FROM mcr.microsoft.com/playwright:focal AS playwright
RUN apt-get update \
    && apt-get install -y wget \
    && wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb \
    && dpkg -i packages-microsoft-prod.deb \
    && apt-get update \
	&& apt-get install dumb-init \
    && apt-get install -y apt-transport-https \
[...]

PR #2173, opened 2022-06-09
Include publishing ARM64 in docker manifest

JakeStevenson:
This should also include arm64 in the docker manifest when published, allowing it to be used on systems with that architecture.  Should help with the later comments on https://github.com/microsoft/playwright-dotnet/issues/1611

### DIFF
--- a/utils/docker/publish_docker.sh
+++ b/utils/docker/publish_docker.sh
@@ -111,4 +111,5 @@ publish_docker_manifest () {
 }
 
 publish_docker_images_with_arch_suffix focal amd64
-publish_docker_manifest focal amd64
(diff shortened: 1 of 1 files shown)


mxschmitt:
Thank you for your contribution! It requires to change the internal part which distributes the driver into the nuget package. Something like that: https://github.com/microsoft/playwright-dotnet/pull/2041

But since the demand was not that high I stopped digging into it.

JakeStevenson:
Yeah, it did seem a bit too easy.  I'll go ahead and close this one until the more complex issue is tackled.

mxschmitt:
This should be fixed now and be available under: `playwright.azurecr.io/public/playwright/dotnet:next`.

PR #2180, opened 2022-06-17
chore: support circular objects

mxschmitt:


src/Playwright.Tests/PageEvaluateHandleTests.cs
```diff
@@ -100,15 +102,33 @@ public async Task ShouldAcceptMultipleNestedHandles()
...
+                var wasCorrectlySerialized = await Page.EvaluateAsync<bool>("a => a.b === a", a);
+                Assert.True(wasCorrectlySerialized);
+            }
+
+            {
+                // lists
+                var a = new List<object>();
+                a.Add(a);
+                var wasCorrectlySerialized = await Page.EvaluateAsync<bool>("a => window.a = a.length === 1 && a[0] && a[0] === a", a);
+                await Page.PauseAsync();
```
pavelfeldman, line 128: Ditto

src/Playwright/Transport/Converters/ProtocolSerializer.cs
```diff
@@ -115,54 +112,66 @@ public object Serialize(object value)
                 return new { r = new { p = regex.ToString(), f = regex.Options.GetInlineFlags() } };
             }
 
-            if (value is IDictionary dictionary && dictionary.Keys.OfType<string>().Any())
+            if (value is ExpandoObject)
             {
-                _visited.Add(value);
+                var o = new List<object>();
+                id = ++visitorInfo.LastId;
+                visitorInfo.Visited.Add(value, id);
```
pavelfeldman, line 119: Visited is a Dictionary (hash table), and you are using a complex (potentially recursive) value as a key. In Java that would stack overflow unless identity hash code and comparison are used, so you'll need something like https://stackoverflow.com/questions/8946790/how-to-use-an-objects-identity-as-key-for-dictionaryk-v for all your Dictionaries.

src/Playwright/Transport/Converters/EvaluateArgumentValueConverter.cs
```diff
@@ -173,27 +182,37 @@ internal static object ParseEvaluateResult(JsonElement result, Type t)
...
         {
             if (parsed == null)
             {
                 return null;
             }
 
+            if (visited.ContainsKey(parsed))
+            {
+                return visited[parsed];
+            }
```
campersau, line 207:
```suggestion
            if (visited.TryGetValue(parsed, out var value))
            {
                return value;
            }
```

src/Playwright/Transport/Converters/EvaluateArgumentValueConverter.cs
```diff
@@ -37,22 +37,19 @@
...
-            if (_visited.Contains(value))
+            int id;
+            if (value == null)
             {
-                throw new JsonException("Argument is a circular structure");
+                return new { v = "null" };
             }
 
-            if (value == null)
+            if (visitorInfo.Visited.ContainsKey(value.GetHashCode()))
```
campersau, line 50:
```suggestion
            if (visitorInfo.Visited.TryGetValue(value.GetHashCode(), out var ref))
```

src/Playwright/Transport/Converters/EvaluateArgumentValueConverter.cs
```diff
@@ -37,22 +37,19 @@
...
             {
-                return new { v = "null" };
+                return new Dictionary<string, object> { ["ref"] = visitorInfo.Visited[value.GetHashCode()] };
```
campersau, line 52:
```suggestion
                return new Dictionary<string, object> { ["ref"] = ref };
```

campersau (commented):
Nit: Avoid multiple dictionary lookups.

src/Playwright/Transport/Converters/EvaluateArgumentValueConverter.cs
```diff
@@ -311,13 +332,21 @@ private static object ParseEvaluateResultToExpando(JsonElement result)
...
+                    list.Add(ParseEvaluateResultToExpando(item, refs));
                 }
                 return list.ToArray();
             }
             return null;
         }
+
+        internal class VisitorInfo
+        {
+            public Dictionary<object, int> Visited { get; set; }
```
pavelfeldman, line 347: This still suggests hashcode computation, we need an identity hash map instead.

pavelfeldman (approved):

PR #2186, opened 2022-06-27
chore: add new 1.23 networking

mxschmitt:


src/Playwright/Core/Request.cs
```diff
@@ -39,24 +40,23 @@ internal class Request : ChannelOwnerBase, IChannelOwner<Request>, IRequest
     {
         private readonly RequestChannel _channel;
         private readonly RequestInitializer _initializer;
-        private readonly RawHeaders _headers;
+        private readonly RawHeaders _provisionalHeaders;
```
pavelfeldman, line 43: nit: keep the name as is, we will update its value along the way.

src/Playwright/Core/Request.cs
```diff
@@ -134,16 +165,20 @@ public async Task<RequestSizesResult> SizesAsync()
...
         public async Task<IReadOnlyList<Header>> HeadersArrayAsync()
-            => (await GetRawHeadersAsync().ConfigureAwait(false)).HeadersArray;
+            => (await ActualHeadersAsync().ConfigureAwait(false)).HeadersArray;
 
         public async Task<string> HeaderValueAsync(string name)
-            => (await GetRawHeadersAsync().ConfigureAwait(false)).Get(name);
+            => (await ActualHeadersAsync().ConfigureAwait(false)).Get(name);
 
-        private Task<RawHeaders> GetRawHeadersAsync()
+        private Task<RawHeaders> ActualHeadersAsync()
```
pavelfeldman, line 176: nit: InnerHeaders

pavelfeldman (approved):

src/Playwright/Core/Request.cs
mxschmitt: lets keep in sync with upstream

src/Playwright/Core/Request.cs
mxschmitt: its like upstream now.

Issue #2185, opened 2022-06-27
[BUG] StdIOTransport Shown in Traces


bjharo:
**Context:**
- Playwright Version: 1.22
- Operating System: Windows, Mac
- .NET version: .NET 6
- Browser: Chromium

**Code Snippet**

```cs
await context.RouteAsync(new Regex(@"https://\w*\.pendo\..*"), r => r.AbortAsync("blockedbyclient"));
```

**Describe the bug**

I'm attempting to use routing to abort certain requests. This is working however, I find my trace files are now filled with a lot entries for StdIOTransport.ProcessStream. I would guess that an entry is generated every time a route is requests is evaluated to see if it matches a defined route. This reduces the usability of trace files and makes it harder to local actual test actions. It would be more useful to only put entries in the trance for requests that matched a route and were acted upon.

<img width="1007" alt="Screen Shot 2022-06-27 at 8 51 28 AM" src="https://user-images.githubusercontent.com/3472778/175946679-eef94e75-cc0c-4e66-a126-249dc7e29624.png">

PR #2193, opened 2022-06-29
fix: apiName calculation when route is used

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2185

### DIFF
--- a/src/Playwright/Core/Route.cs
+++ b/src/Playwright/Core/Route.cs
@@ -97,7 +97,7 @@ namespace Microsoft.Playwright.Core
                 async () =>
                 {
                     await RaceWithPageCloseAsync(_channel.ContinueAsync(url: options.Url, method: options.Method, postData: options.PostData, headers: options.Headers)).ConfigureAwait(false);
-                    return 0;
+                    return 42; // We need to return something to make generics work.
                 },
                 @internal).ConfigureAwait(false);
         }
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -462,18 +462,17 @@ namespace Microsoft.Playwright.Transport
                 {
                     break;
                 }
-                if (methodName.StartsWith("<", StringComparison.InvariantCultureIgnoreCase))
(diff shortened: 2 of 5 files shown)


src/Playwright/Transport/Connection.cs
```diff
@@ -437,43 +437,37 @@ internal async Task<T> WrapApiCallAsync<T>(Func<Task<T>> action, bool isInternal
...
                 var sf = st.GetFrame(i);
                 string fileName = sf.GetFileName();
-                string cSharpNamespace = sf.GetMethod().ReflectedType?.Namespace;
-                bool playwrightInternal = cSharpNamespace != null &&
-                                          (cSharpNamespace == "Microsoft.Playwright" ||
-                                          cSharpNamespace.StartsWith("Microsoft.Playwright.Core", StringComparison.InvariantCultureIgnoreCase) ||
-                                          cSharpNamespace.StartsWith("Microsoft.Playwright.Transport", StringComparison.InvariantCultureIgnoreCase) ||
-                                          cSharpNamespace.StartsWith("Microsoft.Playwright.Helpers", StringComparison.InvariantCultureIgnoreCase));
+                bool playwrightInternal = IsPlaywrightInternalNamespace(sf.GetMethod().ReflectedType?.Namespace);
                 if (string.IsNullOrEmpty(fileName) && !playwrightInternal)
```
dgozman, line 446: Move this to line 466?

dgozman (approved):

src/Playwright/Core/Route.cs
```diff
@@ -97,7 +97,7 @@ internal async Task InnerContinueAsync(bool @internal = false)
                 async () =>
                 {
                     await RaceWithPageCloseAsync(_channel.ContinueAsync(url: options.Url, method: options.Method, postData: options.PostData, headers: options.Headers)).ConfigureAwait(false);
-                    return 0;
+                    return 42; // We need to return something to make generics work.
```
kblok, line 100: I think a `return true;` would look prettier?

PR #2212, opened 2022-07-14
chore: consider runsettings TestRunParameters in context/browser creation

mxschmitt:
Relates https://github.com/microsoft/playwright-dotnet/issues/1730 we'll follow-up with docs to close it.

### DIFF
--- a/src/Playwright.MSTest/ContextTest.cs
+++ b/src/Playwright.MSTest/ContextTest.cs
@@ -35,9 +35,12 @@ namespace Microsoft.Playwright.MSTest
         [TestInitialize]
         public async Task ContextSetup()
         {
-            Context = await NewContextAsync(ContextOptions).ConfigureAwait(false);
+            Context = await NewContextAsync(ContextOptions()).ConfigureAwait(false);
         }
 
-        public virtual BrowserNewContextOptions? ContextOptions => null;
+        public virtual BrowserNewContextOptions ContextOptions()
+        {
+            return null!;
+        }
     }
 }
--- a/src/Playwright.NUnit/PlaywrightTest.cs
+++ b/src/Playwright.NUnit/PlaywrightTest.cs
@@ -22,7 +22,6 @@
  * SOFTWARE.
  */
 
-using System;
 using System.Threading.Tasks;
 using NUnit.Framework;
 
@@ -30,7 +29,7 @@ namespace Microsoft.Playwright.NUnit
 {
     public class PlaywrightTest : WorkerAwareTest
     {
-        public static string BrowserName => (Environment.GetEnvironmentVariable("BROWSER") ?? Microsoft.Playwright.BrowserType.Chromium).ToLower();
+        public string BrowserName { get; internal set; } = null!;
 
         private static readonly Task<IPlaywright> _playwrightTask = Microsoft.Playwright.Playwright.CreateAsync();
 
@@ -41,8 +40,8 @@ namespace Microsoft.Playwright.NUnit
         public async Task PlaywrightSetup()
         {
             Playwright = await _playwrightTask.ConfigureAwait(false);
+            BrowserName = ParsedSettings.BrowserName;
             BrowserType = Playwright[BrowserName];
-            Assert.IsNotNull(BrowserType, $"The requested browser ({BrowserName}) could not be found - make sure your BROWSER env variable is set correctly.");
         }
 
         public ILocatorAssertions Expect(ILocator locator) => Assertions.Expect(locator);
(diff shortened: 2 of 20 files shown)


src/Playwright.NUnit/PlaywrightTest.cs
```diff
@@ -22,15 +22,14 @@
...
-using System;
 using System.Threading.Tasks;
 using NUnit.Framework;
 
 namespace Microsoft.Playwright.NUnit
 {
     public class PlaywrightTest : WorkerAwareTest
     {
-        public static string BrowserName => (Environment.GetEnvironmentVariable("BROWSER") ?? Microsoft.Playwright.BrowserType.Chromium).ToLower();
+        public string BrowserName { get; internal set; } = null!;
```
yury-s, line 32: This is a breaking change, isn't it? I believe we should support both, probably giving priority to the env variable.

src/Playwright.Tests/TestConstants.cs
```diff
@@ -33,17 +33,36 @@ namespace Microsoft.Playwright.Tests
...
+
+        internal static bool IsChromium
+        {
+            get
+            {
+                return BrowserName == BrowserType.Chromium;
+            }
+        }
 
-        internal static readonly bool IsWebKit = BrowserName == BrowserType.Webkit;
```
yury-s, line 44: nit: Why did it change? The previous code was more concise.

src/Playwright/Core/Shared/RunSettingsParser.cs
```diff
@@ -0,0 +1,176 @@
...
+                    return;
+                }
+            }
+            var type = Nullable.GetUnderlyingType(property.PropertyType) ?? property.PropertyType;
+            switch (type)
+            {
+                case Type t when t == typeof(string):
+                    property.SetValue(options, value);
+                    break;
+                case Type t when type == typeof(bool):
```
yury-s, line 85: `Type t when t ==` for consistency with other branches? I'm not actually sure what `Type t` would yield in this particular expression, is this case covered with a test?

src/Playwright.NUnit/PlaywrightTest.cs
yury-s: Ok, I see that it moved to the settings parser.

yury-s (approved):

src/Playwright.MSTest/ContextTest.cs
```diff
@@ -35,9 +35,12 @@ public class ContextTest : BrowserTest
...
         {
-            Context = await NewContextAsync(ContextOptions).ConfigureAwait(false);
+            Context = await NewContextAsync(ContextOptions()).ConfigureAwait(false);
         }
 
-        public virtual BrowserNewContextOptions? ContextOptions => null;
+        public virtual BrowserNewContextOptions ContextOptions()
+        {
+            return null!;
+        }
```
kblok, line 44: Why do we need this change?

src/Playwright.Tests/TestConstants.cs
mxschmitt: BrowserName gets set lazily, so we need a getter which gets executed.

This is only for our internal tests too.

src/Playwright.MSTest/ContextTest.cs
mxschmitt: Align it with nunit harness.

src/Playwright.Tests/TestConstants.cs
```diff
@@ -33,17 +33,36 @@ namespace Microsoft.Playwright.Tests
...
     {
-        public static string BrowserName => NUnit.PlaywrightTest.BrowserName;
+        public static string BrowserName { get; set; } = null!;
 
         public const int DefaultTestTimeout = 30_000;
         public const int DefaultTimeout = 10_000;
         public const int DefaultTaskTimeout = 5_000;
 
-        public const string AboutBlank = "about:blank";
+        internal static bool IsWebKit
```
campersau, line 42: Nit: you can use the shorthand syntax if you want:
```suggestion
        internal static bool IsWebKit => BrowserName == BrowserType.Webkit;
```

src/Playwright/Core/Shared/RunSettingsParser.cs
mxschmitt: will follow up with better tests.

src/Playwright.Tests/TestConstants.cs
mxschmitt: sweet.

PR #2217, opened 2022-07-21
test: add test harness integration tests

mxschmitt:


src/Playwright.TestingHarnessTest/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {},
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@playwright/test": "^1.23.4",
+    "fast-xml-parser": "^4.0.9"
+  }
```
rwoll, line 13:
```suggestion
  "name": "playwright.testingharnesstest",
  "private": true,
  "devDependencies": {
    "@playwright/test": "^1.23.4",
    "fast-xml-parser": "^4.0.9"
  }
```

src/Playwright.TestingHarnessTest/playwright.config.ts
```diff
@@ -0,0 +1,11 @@
+import type { PlaywrightTestConfig } from '@playwright/test';
+
+const config: PlaywrightTestConfig = {
+  testDir: './tests',
+  timeout: 2 * 60 * 1_000,
+  workers: 1,
+  reporter: 'list',
```
rwoll, line 7: Do you want `list` in CI, or `dot`?

src/Playwright.TestingHarnessTest/tests/mstest.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+    '.runsettings': `
+      <?xml version="1.0" encoding="utf-8"?>
+      <RunSettings>
+        <RunConfiguration>
+          <EnvironmentVariables>
+            <DEBUG>pw:api</DEBUG>
+          </EnvironmentVariables>
+        </RunConfiguration>
+      </RunSettings>`,
+  }, 'dotnet test --settings=.runsettings');
```
rwoll, line 39: This seems pretty painful to author, edit, and debug. Instead of putting string inline, consider changing `runTest` to just take a path to a directly with all these files. That way, your editor can help with the verbose XML and C#.

Our JS test harness suffers the same issue, but for some reason I don't mind it in JS, but I do mind it here. Perhaps it's because there's not so much extra syntax.

rwoll (approved):
Overall, seems fine, but I'd encourage you to not use the inline XML and C# strings. Not a blocker though.

src/Playwright.TestingHarnessTest/playwright.config.ts
mxschmitt: since we have little tests which take long, list is helpful there!

PR #2222, opened 2022-07-21
chore(roll): roll to 1.24.0

mxschmitt:


rwoll (approved):
looks good assuming the CI passes. (NB: Looks like your branch name refers to an old version: [roll/1.21-release](https://github.com/mxschmitt/playwright-dotnet/tree/roll/1.21-release). Doesn't make a difference here but was a bit confusing!)

PR #2246, opened 2022-08-11
chore: new test harness options

mxschmitt:
This makes the following work:

```xml
<?xml version="1.0" encoding="utf-8"?>
<RunSettings>
  <Playwright>
    <Browser>chromium</Browser>
    <ExpectTimeout>10000</ExpectTimeout>
    <LaunchOptions>
      <Headless>false</Headless>
      <Channel>msedge</Channel>
    </LaunchOptions>
     <!-- The following is not yet implemented -->
    <ActionTimeout>10000</ActionTimeout>
    <Retries>3</Retries>
    <Trace>OnFirstRetry</Trace>
  </Playwright>
</RunSettings>
```

instead of `TestRunParameters` which gives us more flexibility for the future and is easier to use when overriding it via CLI args. We are aware that we might break users, but we didn't announce it yet so it should not harm too many.


src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+                return BrowserType.Chromium;
+            }
+            private set { }
+        }
+
+        public static float? ExpectTimeout
+        {
+            get
+            {
+                if (_settings == null)
```
mxschmitt, line 63: in ALL the getters `_settings` can be null if there is no `RUNSETTINGS` file given. We fallback then to defaults.

src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+                browserName != BrowserType.Webkit)
+            {
+                throw new ArgumentException($"Invalid browser name: {browserName}");
+            }
+        }
+
+        public void Load(XmlReader reader)
+        {
+            _settings = new PlaywrightSettingsXml(reader);
+        }
```
kblok, line 107:
```suggestion
            => _settings = new PlaywrightSettingsXml(reader);
```

src/Playwright.TestAdapter/PlaywrightSettingsXml.cs
```diff
@@ -0,0 +1,147 @@
...
+                        break;
+                    default:
+                        Console.Error.WriteLine($"Playwright RunSettings Parsing Error: Playwright>{reader.Name} is not implemented");
+                        break;
+                }
+            }
+        }
+
+        private static T ParseXmlIntoClass<T>(XmlReader reader)
+        where T : class, new()
```
kblok, line 72:
```suggestion
        private static T ParseXmlIntoClass<T>(XmlReader reader) where T : class, new()
```

src/Playwright.TestAdapter/PlaywrightSettingsXml.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            var property = options.GetType().GetProperty(key);
+            if (property == null)
+            {
+                return;
+            }
+            var type = Nullable.GetUnderlyingType(property.PropertyType) ?? property.PropertyType;
+            switch (type)
+            {
+                case Type t when t == typeof(string):
```
kblok, line 97:
```
case string stringValue:
case bool boolValue:
```
Won't work?

mxschmitt: compile does not allow it.

src/Playwright.TestAdapter/Playwright.TestAdapter.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ReleaseVersion>0.0.0</ReleaseVersion>
```
yury-s, line 6: is it expected to be all zeros?

src/Playwright.TestAdapter/PlaywrightSettingsXml.cs
```diff
@@ -0,0 +1,147 @@
...
+                    continue;
+                }
+                switch (reader.Name)
+                {
+                    case "Browser":
+                        reader.Read();
+                        Browser = reader.Value;
+                        break;
+                    case "LaunchOptions":
+                        LaunchOptions = ParseXmlIntoClass<BrowserTypeLaunchOptions>(reader);
```
yury-s, line 54: Shouldn't this go under Browser section?

src/Playwright.TestAdapter/PlaywrightSettingsXml.cs
```diff
@@ -0,0 +1,147 @@
...
+                        break;
+                    case "Headless":
+                        reader.Read();
```
yury-s, line 57: Ditto?

yury-s (approved):

src/Playwright.TestAdapter/Playwright.TestAdapter.csproj
mxschmitt: yes and fixed.

Issue #2218, opened 2022-07-21
[BUG] Nunit fails parallel test in same file when overriding BrowserNewContextOptions ContextOptions and using [Parallelizable(ParallelScope.All)]


j-meds:
- Playwright Version: 1.23
- Operating System: Windows
- .NET version: 6
- Browser: Chromium

I'm getting Flaky test when running two similar test in parallel, using [Parallelizable(ParallelScope.All)]. The test always pass when running with [Parallelizable(ParallelScope.Self)], which runs one at time. I 

**Code Snippet**

```cs
[Parallelizable(ParallelScope.All)]
public class UnionReportTests : PageTest
{

    public override BrowserNewContextOptions ContextOptions()
    {
        return new BrowserNewContextOptions()
        {
            StorageStatePath = "./state.json",
        };
    }

    // Testing Admin union reports view
    [Test]
    public async Task Unions_Reports_View()
    {
        // TEST PASS IN PARALLEL WHEN I CREATE A NEW BROWSER CONTEXT
        // Create a new context with the saved storage state.
        //var context = await Browser.NewContextAsync(new BrowserNewContextOptions
        //{
        //    StorageStatePath = "state.json"
        //});


        //var Page = await context.NewPageAsync();

        var unionPage = new UnionReportPage(Page);

        await unionPage.NavigateToUnionPage();
        // User should be on/Reports/Unions
        StringAssert.Contains("/Reports/Unions", Page.Url);

        // Table should be available
        await Expect(Page.Locator(".p-datatable#unionsReportTable")).ToBeVisibleAsync();
        await Page.WaitForSelectorAsync(".p-datatable-tbody > tr");
        int rowCount = await Page.Locator(".p-datatable-tbody > tr").CountAsync();

        Assert.True(rowCount > 0, "The union table should have more than one row");

        unionPage.searchTable("Boilermakers Local 92");

        rowCount = await Page.Locator(".p-datatable-tbody > tr").CountAsync();

        // the table should only have one record
        Assert.True(rowCount == 1, "The union table should have one row");
    }
   
     //THIS IS THE SAME TEST WITH A DIFFERENT NAME
    // Testing Admin union reports view
    [Test]
    public async Task Unions_Reports_Create()
    {
        TEST PASS IN PARALLEL WHEN I CREATE A NEW BROWSER CONTEXT
        // Create a new context with the saved storage state.
        //var context = await Browser.NewContextAsync(new BrowserNewContextOptions
        //{
[...]

PR #2248, opened 2022-08-11
feat: Introduce nunit analyzers

kblok:
This PR will introduce static analyzers in the NUnit library.
The first analyzer `PWNunit1001`, will enforce `ParallelScope.Self` in the `Parallelizable` attribute. 
This will open the door to new analyzers that will help us share good practices with the users.

The code is inspired on [ParallelizableUsageAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/ParallelizableUsage/ParallelizableUsageAnalyzer.cs) from [nunit.analyzers](https://github.com/nunit/nunit.analyzers).

Related to #2218

### DIFF
--- a/src/Playwright.TestingHarnessTest/Playwright.TestingHarnessTest.csproj
+++ b/src/Playwright.TestingHarnessTest/Playwright.TestingHarnessTest.csproj
@@ -19,6 +19,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Playwright.MSTest\Playwright.MSTest.csproj" />
-    <ProjectReference Include="..\Playwright.NUnit\Playwright.NUnit.csproj" />
+    <ProjectReference Include="..\Playwright.NUnit\Playwright.NUnit.csproj" OutputItemType="Analyzer"/>
   </ItemGroup>
 </Project>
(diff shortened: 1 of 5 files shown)


src/Playwright.TestingHarnessTest/Playwright.TestingHarnessTest.csproj
```diff
@@ -19,6 +19,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Playwright.MSTest\Playwright.MSTest.csproj" />
-    <ProjectReference Include="..\Playwright.NUnit\Playwright.NUnit.csproj" />
+    <ProjectReference Include="..\Playwright.NUnit\Playwright.NUnit.csproj" OutputItemType="Analyzer"/>
```
kblok, line 22: We need to add this because we are using the project instead of the nuget.

src/Playwright.TestingHarnessTest/tests/nunit.spec.ts
```diff
@@ -335,3 +335,81 @@ test.describe('Expect() timeout', () => {
     expect(result.stdout).toContain("LocatorAssertions.ToHaveTextAsync with timeout 123ms")
   });
 });
+
+test.describe('Parallelizable analyzer', () => {
```
Meir017, line 339: consider testing the analyzer using pure C# and verifying the analyzer creates a diagnostic - https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/tutorials/how-to-write-csharp-analyzer-code-fix#build-unit-tests

kblok: Yeah, I thought about that. But as we already have all the test harness set up and the overall goal is pretty much the same, I went that way.
I would have loved the Test harness suite to be a C# project :)

PR #2265, opened 2022-08-22
chore: add workaround for packages.config

mxschmitt:


nohwnd (approved):
As suggested in the last line of buildTransitive proposal: https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior#solution

> To construct a package which allows build assets to flow transitively, package author will need to **put all these build assets in /buildTransitive as well as /build folder** to make the package **compatible with packages.config**.

By experimentation I saw that the target is executed only once, so this is all we should need to do.

PR #2241, opened 2022-08-08
fix: make sure it works with packages.config (.NET 4.8)

BeniFreitag:
Extended Playwright.CreateAsync() with optional driversPath param.

Fixes #1842, fixes #2240, fixes #2004

`Playwright.CreateAsync()` fails in .NET 4.8 with "PlaywrightException: Driver not found" and there's currently no workaround.

This change allows to specify the driversPath, to a path where the runtime can pick the .playwright-folder:
```c#
Playwright.CreateAsync(driversPath: "path/.playwright")
```

The implementation is inspired by [PlaywrightSharp](https://github.com/hardkoded/playwright-sharp). If driversPath is specified, it uses that path, otherwise it tries to find the .playwright-folder automatically (as before).

A working example with ASP.Net 4.8 can be found at https://github.com/BeniFreitag/Playwright-AspNet-4.8/tree/develop

Issue #2240, opened 2022-08-05
[BUG]: .NET Framework 4.8 Driver not found in Temporary ASP.NET Files


BeniFreitag:
**Context:**
- Playwright Version: 1.24.1
- Operating System: Windows 11 (and Windows Server 2019)
- .NET version: .NET 4.8
- Browser: All

**Code Snippet**

Sample-Project with ASP.Net Website and ASP.Net Webapplication at https://github.com/BeniFreitag/Playwright-AspNet-4.8

**Describe the bug**

Followup of #1842 and #1872. When the ASP.Net Application runs, DLLs are copied to _Temporary ASP.NET Files_ automatically. [`Microsoft.Playwright.Helpers.Driver.GetExecutablePath()`](https://github.com/microsoft/playwright-dotnet/blob/main/src/Playwright/Helpers/Driver.cs#L35-L58) looks for a .playwright folder there, but the ASP.NET ecosystem/runtime doesn't copy this folder to _Temporary ASP.NET Files_, so it always fails with error:

```
Driver not found: C:\Windows\Microsoft.NET\Framework64\v4.0.30319\Temporary ASP.NET Files\vs\6c77f81d\ac62943a\assembly\dl3\.playwright\node\win32_x64\playwright.cmd

Stack Trace:
[PlaywrightException: Driver not found: C:\Windows\Microsoft.NET\Framework64\v4.0.30319\Temporary ASP.NET Files\vs\6c77f81d\ac62943a\assembly\dl3\.playwright\node\win32_x64\playwright.cmd]
   Microsoft.Playwright.Helpers.Driver.GetExecutablePath() in /_/src/Playwright/Helpers/Driver.cs:57
   Microsoft.Playwright.Transport.StdIOTransport.GetProcess() in /_/src/Playwright/Transport/StdIOTransport.cs:119
   Microsoft.Playwright.Transport.StdIOTransport..ctor() in /_/src/Playwright/Transport/StdIOTransport.cs:44
   Microsoft.Playwright.<CreateAsync>d__0.MoveNext() in /_/src/Playwright/Playwright.cs:44
   System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() +32
   System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) +62
   System.Runtime.CompilerServices.TaskAwaiter`1.GetResult() +32
[...]

Issue #2004, opened 2022-02-10
[BUG] Playwright does not work with packages.config (.NET 4.8)


Jojoshua:
**Context:**
- Playwright Version: 1.18.1
- Operating System: Win 10
- .NET version: 4.8
- Browser: N/A

**Code Snippet**

```cs
 static async Task Main(string[] args)
        {
            await Go().ConfigureAwait(false);
        }

        public static async Task Go()
        {
            using (var playwright = await Playwright.CreateAsync())
            {
                var browser = await playwright.Chromium.LaunchAsync();
                var page = await browser.NewPageAsync();
                await page.GotoAsync("https://playwright.dev/dotnet");
                await page.ScreenshotAsync(new PageScreenshotOptions { Path = "screenshot.png" });
                await browser.DisposeAsync();
            }
        }
```

**Describe the bug**
Driver is not found upon load. Clone this project for complete reproduction

https://github.com/Jojoshua/Playwright4.8InstallIssue

mxschmitt:
I'd recommend to migrate to PackageReference, then it should work as expected: https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference

Jojoshua:
> I'd recommend to migrate to PackageReference, then it should work as expected: https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference

This worked for my plain console app but it did not work for my asp.net project. When I tried to convert it to package references an error said my project was not eligible for upgrading.

campersau:
@Jojoshua These resources might help you with the migration:
https://github.com/NuGet/docs.microsoft.com-nuget/issues/860#issuecomment-409207305
https://stackoverflow.com/questions/57581514/vs-2015-to-2017-migrate-to-package-reference-failed/65701746

Looks like https://github.com/NuGet/Home/issues/5877 is still not completely implemented...

Jojoshua:
> @Jojoshua These resources might help you with the migration: [NuGet/docs.microsoft.com-nuget#860 (comment)](https://github.com/NuGet/docs.microsoft.com-nuget/issues/860#issuecomment-409207305) https://stackoverflow.com/questions/57581514/vs-2015-to-2017-migrate-to-package-reference-failed/65701746
> 
> Looks like [NuGet/Home#5877](https://github.com/NuGet/Home/issues/5877) is still not completely implemented...

[...]

PR #2266, opened 2022-08-22
chore: respect CodeBase when locating driver entrypoint

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/pull/2241
Fixes https://github.com/microsoft/playwright-dotnet/issues/2004
Fixes https://github.com/microsoft/playwright-dotnet/issues/2240

### DIFF
--- a/src/Playwright/Helpers/Driver.cs
+++ b/src/Playwright/Helpers/Driver.cs
@@ -37,16 +37,18 @@ namespace Microsoft.Playwright.Helpers
             DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
             if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, "Microsoft.Playwright.dll")))
             {
-#if NETFRAMEWORK
+                string assemblyLocation = null;
+#pragma warning disable SYSLIB0012 // 'Assembly.CodeBase' is obsolete: 'Assembly.CodeBase and Assembly.EscapedCodeBase are only included for .NET Framework compatibility.
                 var assemblyLocationFileUrl = typeof(Playwright).Assembly.CodeBase;
-                if (string.IsNullOrEmpty(assemblyLocationFileUrl))
+#pragma warning restore SYSLIB0012 // 'Assembly.CodeBase' is obsolete: 'Assembly.CodeBase and Assembly.EscapedCodeBase are only included for .NET Framework compatibility.
+                if (!string.IsNullOrEmpty(assemblyLocationFileUrl))
                 {
-                    throw new PlaywrightException("Unable to find Microsoft.Playwright.dll");
+                    assemblyLocation = new Uri(assemblyLocationFileUrl).LocalPath;
                 }
-                var assemblyLocation = new Uri(assemblyLocationFileUrl).LocalPath;
-#else
-                var assemblyLocation = typeof(Playwright).Assembly.Location;
-#endif
+                else
+                {
+                    assemblyLocation = typeof(Playwright).Assembly.Location;
+                }
                 assemblyDirectory = new FileInfo(assemblyLocation).Directory;
             }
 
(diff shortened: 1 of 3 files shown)


src/Playwright/Helpers/Driver.cs
```diff
@@ -37,7 +37,17 @@ internal static string GetExecutablePath()
             DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
             if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, "Microsoft.Playwright.dll")))
             {
+#if NETFRAMEWORK
                 var assemblyLocation = typeof(Playwright).Assembly.Location;
+#else
+
+                var assemblyLocationFileUrl = typeof(Playwright).Assembly.CodeBase;
+                if (string.IsNullOrEmpty(assemblyLocationFileUrl))
```
mxschmitt, line 45: I'm not sure if I should throw there or instead fallback to  `typeof(Playwright).Assembly.Location;´.

mxschmitt: Looks fine, since we only do it with .net framework: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.codebase?view=net-6.0 and its always set there.

src/Playwright/Helpers/Driver.cs
```diff
@@ -37,7 +37,19 @@ internal static string GetExecutablePath()
             DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
             if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, "Microsoft.Playwright.dll")))
             {
+#if NETFRAMEWORK
```
campersau, line 40: I am not sure if that works for a project targeting `netstandard2.0` only. Also I guess that this should be negated?

kblok: Yup. This will always be false.

mxschmitt: Yeah, it's wrong, we should somehow check during the runtime if the application is `NETFRAMEWORK` since its always netstandard.

nohwnd: Is the value different between .NET and .NET Framework? Maybe you can just always use CodeBase? 

You could check on runtime, but it is a bit annoying because the versions don't correspond exactly to the .NET versions. https://stackoverflow.com/questions/1826688/get-current-net-clr-version-at-runtime

nohwnd (approved):

src/Playwright/Helpers/Driver.cs
```diff
@@ -37,7 +37,18 @@ internal static string GetExecutablePath()
...
+                var assemblyLocationFileUrl = typeof(Playwright).Assembly.CodeBase;
+#pragma warning restore SYSLIB0012 // 'Assembly.CodeBase' is obsolete: 'Assembly.CodeBase and Assembly.EscapedCodeBase are only included for .NET Framework compatibility.
+                if (!string.IsNullOrEmpty(assemblyLocationFileUrl))
+                {
+                    assemblyLocation = new Uri(assemblyLocationFileUrl).LocalPath;
+                }
+                else
+                {
+                    assemblyLocation = typeof(Playwright).Assembly.Location;
+                }
```
campersau, line 51:
```suggestion
                string assemblyLocation;
                var assembly = typeof(Playwright).Assembly;
#pragma warning disable SYSLIB0012 // 'Assembly.CodeBase' is obsolete: 'Assembly.CodeBase and Assembly.EscapedCodeBase are only included for .NET Framework compatibility.
                if (Uri.TryCreate(assembly.CodeBase, UriKind.Absolute, out var codeBase) && codeBase.IsFile)
#pragma warning restore SYSLIB0012 // 'Assembly.CodeBase' is obsolete: 'Assembly.CodeBase and Assembly.EscapedCodeBase are only included for .NET Framework compatibility.
                {
                    assemblyLocation = codeBase.LocalPath;
                }
                else
                {
                    assemblyLocation = assembly.Location;
                }
```

Slightly more paranoid version as `new Uri` could throw.
The same logic is also used by `Microsoft.Extensions.DependencyModel`: 
https://github.com/dotnet/runtime/blob/e55c908229e36f99a52745d4ee85316a0e8bb6a2/src/libraries/Microsoft.Extensions.DependencyModel/src/DependencyContextLoader.cs#L161

But they are using `CodeBase` as a fallback instead and don't prefer it over `Assembly.Location` like here.
https://github.com/dotnet/runtime/blob/e55c908229e36f99a52745d4ee85316a0e8bb6a2/src/libraries/Microsoft.Extensions.DependencyModel/src/DependencyContextLoader.cs#L130-L171

PR #2273, opened 2022-08-25
chore: fix dotnet build did not pack the same results as dotnet pack

### LINKED ISSUE #2271: [internal] dotnet build does not produce the same nuget as dotnet pack
e.g. `playwright.ps1` is missing. Not a huge deal or blocker but would be nice if they would produce the same.

Attemps to fix it were https://github.com/microsoft/playwright-dotnet/pull/2267 and https://github.com/microsoft/playwright-dotnet/pull/2268.

cc @campersau

campersau:
Removes `Microsoft.Playwright.targets` import from `Playwright.csproj` and inline `DedupeDriver` using `ItemGroup`s with `CopyToOutputDirectory` / `Pack` attributes instead of using `Copy`. Similar to `Microsoft.Playwright.targets`.

Supersedes #2267 and #2268

Fixes #2271

### DIFF
--- a/src/Playwright/Playwright.csproj
+++ b/src/Playwright/Playwright.csproj
@@ -24,7 +24,6 @@
   <Import Project="../Common/Version.props" />
   <Import Project="../Common/Dependencies.props" />
   <Import Project="../Common/SignAssembly.props" />
-  <Import Project="build/Microsoft.Playwright.targets" />
   <ItemGroup>
     <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core" Version="1.0.0">
       <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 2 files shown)


src/Playwright/Playwright.csproj
```diff
@@ -35,33 +34,21 @@
...
-    <Copy SourceFiles="@(DriverShell)" DestinationFiles="@(DriverShell-&gt;'.playwright\node\linux-x64\%(RecursiveDir)%(Filename)%(Extension)')" />
-    <Copy SourceFiles="@(DriverShell)" DestinationFiles="@(DriverShell-&gt;'.playwright\node\linux-arm64\%(RecursiveDir)%(Filename)%(Extension)')" />
-    <Copy SourceFiles="@(DriverShell)" DestinationFiles="@(DriverShell-&gt;'.playwright\node\mac\%(RecursiveDir)%(Filename)%(Extension)')" />
-    <Copy SourceFiles="@(DriverCmd)" DestinationFiles="@(DriverCmd-&gt;'.playwright\node\win32_x64\%(RecursiveDir)%(Filename)%(Extension)')" />
-  </Target>
   <ItemGroup>
-    <None Include=".playwright\**" Pack="true" PackagePath=".playwright" />
+    <None Remove=".drivers\**" />
+    <None Include=".drivers\linux\package\**" Link=".playwright\package\%(RecursiveDir)%(Filename)%(Extension)" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\package" />
+    <None Include=".drivers\linux\LICENSE" Pack="true" PackagePath=".playwright\node" />
```
campersau, line 40: `LICENSE` is only packed and not copied to output directory since it isn't technically needed there. Please let me know if it should also be copied.

campersau:
These could now be removed since the `.playwright` directory is only created in the `bin` output:
https://github.com/microsoft/playwright-dotnet/blob/df8d6de48491bff3f077c25c17ee24f3e7cc77ca/.gitignore#L356
https://github.com/microsoft/playwright-dotnet/blob/df8d6de48491bff3f077c25c17ee24f3e7cc77ca/.gitignore#L354

nohwnd (approved):

nohwnd:
Approved, I think it is better that the build of the package does not depend on the target that it ships, because then it is hard to keep both local build of `Playwright.csproj` working and not breaking the behavior for users in their build.

mxschmitt:
> These could now be removed since the `.playwright` directory is only created in the `bin` output:
> 
> https://github.com/microsoft/playwright-dotnet/blob/df8d6de48491bff3f077c25c17ee24f3e7cc77ca/.gitignore#L356
> 
> 
> https://github.com/microsoft/playwright-dotnet/blob/df8d6de48491bff3f077c25c17ee24f3e7cc77ca/.gitignore#L354

Lets remove them as well?

mxschmitt (approved):
Great PR, thanks!

PR #2279, opened 2022-08-29
chore: cleanup connection and align with upstream

campersau:
The same object got tryadded 3 times before:
- `OnObjectCreated` which wasn't really used.
- `CreateRemoteObject`
- `ChannelOwnerBase`

Now only `ChannelOwnerBase` adds objects which aligns with upstream / java.

---

`__adopt__` was wrong as it adopted the object itself, now it adopts the child correctly.

---

Throw exception when callback is not found.

---

Remove some #pragma warnings since netstandard2.0 is the only target.

### DIFF
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -139,9 +139,7 @@ namespace Microsoft.Playwright.Transport
                         object obj = propertyDescriptor.GetValue(args);
                         if (obj != null)
                         {
-#pragma warning disable CA1845 // Use span-based 'string.Concat' and 'AsSpan' instead of 'Substring
                             string name = propertyDescriptor.Name.Substring(0, 1).ToLowerInvariant() + propertyDescriptor.Name.Substring(1);
-#pragma warning restore CA2000 // Use span-based 'string.Concat' and 'AsSpan' instead of 'Substring
                             sanitizedArgs.Add(name, obj);
                         }
                     }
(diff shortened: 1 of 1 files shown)


src/Playwright/Transport/Connection.cs
```diff
@@ -247,30 +241,31 @@ internal void Dispatch(PlaywrightServerMessage message)
...
+                @object.Channel?.OnMessage(message.Method, message.Params);
             }
             catch (Exception ex)
             {
                 DoClose(ex);
             }
         }
 
-        private void CreateRemoteObject(string parentGuid, ChannelOwnerType type, string guid, JsonElement? initializer)
+        private IChannelOwner CreateRemoteObject(string parentGuid, ChannelOwnerType type, string guid, JsonElement? initializer)
```
mxschmitt, line 268:
```suggestion
        private IChannelOwner CreateRemoteObject(string parentGuid, ChannelOwnerType type, string guid, JsonElement? initializer)
```
```suggestion
        private void CreateRemoteObject(string parentGuid, ChannelOwnerType type, string guid, JsonElement? initializer)
```
Do I see it correctly that the return value is unused?

campersau: Yes, the return value is unused. But that is also the case in javascript and java 😄 
I can change it if you want, maybe this should be done everywhere?

https://github.com/microsoft/playwright/blob/d56b27962207de38f3b4427bcf4f2a3d9f8b1077/packages/playwright-core/src/client/connection.ts#L188-L294

https://github.com/microsoft/playwright-java/blob/4d278c391e3c50738ddea6c3e324a4bbbf719d86/playwright/src/main/java/com/microsoft/playwright/impl/Connection.java#L225-L326

mxschmitt (approved):

PR #2281, opened 2022-08-29
fix: TracingStopChunkOptions can be null

### LINKED ISSUE #2280: [BUG]: Tracing.StopChunkAsync causes NullReferenceException due to incorrect handling of optional TracingStopOptions options parameter
Microsoft.Playwright 1.25.0

https://github.com/microsoft/playwright-dotnet/blob/4dc7ff349a59ac07e9200b05f9a6406a0fcec9a2/src/Playwright/Core/Tracing.cs#L63-L66

Passing null as the option causes an NullReferenceException due to the `options.Path` Property access. In the [original typescript](https://github.com/microsoft/playwright/blob/e975aef961055c67cc9e4a25ffb0357bdca23ff1/packages/playwright-core/src/client/tracing.ts#L42-L44) version this method does an optional property access

```ts
  async stopChunk(options: { path?: string } = {}) {
    await this._doStopChunk(options.path);
  }
```

So I guess the fix would be to also do a conditional access here.

campersau:
Fixes #2280

### DIFF
--- a/src/Playwright.Tests/TracingTests.cs
+++ b/src/Playwright.Tests/TracingTests.cs
@@ -125,6 +125,57 @@ namespace Microsoft.Playwright.Tests
 
         }
 
+        [PlaywrightTest("tracing.spec.ts", "should work with multiple chunks")]
+        public async Task ShouldWorkWithMultipleChunks()
+        {
+            using var tmp = new TempDirectory();
+
+            await Context.Tracing.StartAsync(new()
+            {
+                Screenshots = true,
+                Snapshots = true,
+            });
+
+            var page = await Context.NewPageAsync();
+            await page.GotoAsync(Server.Prefix + "/frames/frame.html");
+
+            await Context.Tracing.StartChunkAsync();
+            await page.SetContentAsync("<button>Click</button>");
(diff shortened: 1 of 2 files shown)


src/Playwright.Tests/TracingTests.cs
```diff
@@ -125,6 +125,57 @@ public async Task ShouldCollectTwoTraces()
...
+            await Context.Tracing.StopChunkAsync(new TracingStopChunkOptions { Path = traceFile2 });
+
+            await Context.Tracing.StartChunkAsync();
+            await page.HoverAsync("'Click'");
+            await Context.Tracing.StopChunkAsync(); // Should stop without a path.
+
+            {
+                var (events, resources) = ParseTrace(traceFile1);
+                Assert.AreEqual("context-options", events[0].Type);
+                Assert.GreaterOrEqual(events.Where(x => x.Metadata?.ApiName == "Page.SetContentAsync").Count(), 1);
```
mxschmitt, line 160: can we use `GetActions` there to assert it? since line 161 and 162 are the same. Like we have upstream.

mxschmitt (approved):

Issue #2272, opened 2022-08-25
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [ ] https://github.com/microsoft/playwright/commit/622c73cc1e677834ebdc27e95c89ff101bdccf54 (feat: expect(locator).toHaveAttribute to assert attribute presence (#16767))
  - [x] https://github.com/microsoft/playwright/commit/a07a4a25a26e2fa6a976fb2ff174f25b24414da6 (chore: make parent scope explicit (#16819))
  - [ ] https://github.com/microsoft/playwright/commit/f30ac1d67827fbcf3f8be7a851ef81b761842c52 (fix: support toHaveAttribute(name, options) (#16941))

PR #2287, opened 2022-09-01
chore(roll): roll Playwright to 1.26.0-alpha-sep-1-2022

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2272


src/Playwright.NUnit/PlaywrightTestAttribute.cs
```diff
@@ -100,6 +100,10 @@ private static string GenerateOutputLog(List<TestResult> failedResults, int retr
                     output += $"\nFailing test runs:\n";
                     foreach (var result in failedResults)
                     {
+                        if (result == null)
```
mxschmitt, line 103: drive-by linting fix, since result and result.Message could be null.

src/Playwright.Tests/PageGotoTests.cs
```diff
@@ -583,10 +582,11 @@ public async Task ShouldReturnWhenNavigationIsComittedIfCommitIsSpecified()
                 // Write enought bytes of the body to trigge response received event.
                 var str = "<title>" + new string('a', 4100);
                 await context.Response.WriteAsync(str);
-                await context.Response.BodyWriter.FlushAsync();
+                await context.Response.Body.FlushAsync();
+                await Task.Delay(5_000);
```
yury-s, line 586: does it have to be so long?

yury-s (approved):

PR #2297, opened 2022-09-07
chore(nunit): make PlaywrightTestAttribute inherit from NUnit.TestAttribute

mxschmitt:


dgozman (approved):
Does this need a codegen change as well?

PR #2302, opened 2022-09-10
chore: align connection close with upstream impl

mxschmitt:
Apply the upstream https://github.com/microsoft/playwright/pull/9524 patch.

Drive-by fix specific connect waiter.cs tests. For the actual fox of them see the comment related to `callback.Value.TaskCompletionSource.Task.IgnoreException();`.


yury-s (approved):
Is there a test that covers this?

mxschmitt:
> Is there a test that covers this?

Mostly yeah, previously they were flaky and we have CDP disconnection tests in various states.

PR #2315, opened 2022-09-14
chore: make retry attributes internal

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2316


src/Playwright.MSTest/Services/BrowserService.cs
```diff
@@ -27,7 +27,7 @@
 
 namespace Microsoft.Playwright.MSTest.Services;
 
-public class BrowserService : IWorkerService
+internal class BrowserService : IWorkerService
```
mxschmitt, line 30: drive-by: should not be public.

src/Playwright.NUnit/BrowserService.cs
```diff
@@ -27,7 +27,7 @@
 
 namespace Microsoft.Playwright.NUnit;
 
-public class BrowserService : IWorkerService
+internal class BrowserService : IWorkerService
```
mxschmitt, line 30: drive-by: should not be public.

pavelfeldman (approved):

dgozman (approved):

Issue #2316, opened 2022-09-14
[Feature] Add Retries


mxschmitt:
Retries got reverted since there were bugs in it which would prevent users from using it efficiently.

Open items:

- Make sure it works with timeouts
- Make sure SetUp and TearDown runs as expected
- Add more tests for edge cases
- Revert https://github.com/microsoft/playwright-dotnet/pull/2315
- Revert https://github.com/microsoft/playwright/commit/59c32bf2c6470645e9903eb7903425dcc4cf23f8

bheemreddy181:
Really looking forward into this - we are missing retries at this moment

bheemreddy181:
Can we also make retries differ based on the tag - like say if we have API tests then set retries to 0 and set retries to number n for web tests ?

mxschmitt:
Retries would be for the whole project, so you can define `<Retries>1</Retries>` inside your runsettings file or when defining it via `--settings` to `dotnet test`. This is the same way how it's done for Node.js.

Just a question out of curiosity, are you using NUnit or MSTest?

bheemreddy181:
We are using Nunit - makes sense you can ignore the tag request. It’s better to have different run settings for different tag runs

nicolaiarocci:
Looking forward to this one too.

masaplecas:
Is there any ETA about this one?

ArcadeMode:
Is there any news on this? We're enjoying using playwright-dotnet a lot but retries would save us a lot of effort on tests that fail due to random failures.

badsyntax:
Has anyone tried using test framework retries eg [NUnit retries](https://docs.nunit.org/articles/nunit/writing-tests/attributes/retry.html)?

gauravkhuraana:
Dotnet is lagging compared to nodejs in many perspective. Can we please bring it to speed and bring on these features

ashleybissell:
Do we have any timescales for when this might be looked at again? It appears to be a popular issue and would be a big help to a lot of users.

mxschmitt:
@ertiusmeo we are actively exploring options for NUnit and MSTest and collaborate with the MSTest maintainers. (no eta unfortunately)

Ozkanil:
Hello @mxschmitt is there any update on retry support for .net? Thanks!

jfuinsure:
Any updates?

mattvb91:
[...]

PR #2319, opened 2022-09-19
chore: make retries work with custom timeouts

mxschmitt:
Differences between NUnit and MSTest:

- NUnit will call the `afterEach` when a timeout is reached, this is not possible in MSTest since the `base.Execute()` takes care of calling `beforeEach/afterEach`.

Both implementations will:
- throw if test framework timeout < Playwright test timeout
- gracefully close the BrowserContext when the test timeout has reached
- making reached timeouts compatible with retries

Relates https://github.com/microsoft/playwright-dotnet/issues/2316

### DIFF
--- a/src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
+++ b/src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
@@ -1,0 +1,115 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Playwright.TestAdapter;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Playwright.MSTest;
+
+internal class PlaywrightTimeoutAttribute : TestMethodAttribute
+{
+    static internal Func<Task>? ContextCloseHookOnTimeout { get; set; } = null;
+
+    public override TestResult[] Execute(ITestMethod testMethod)
+    {
+        var timeout = PlaywrightSettingsProvider.TestTimeout;
+        if (timeout == null || Debugger.IsAttached)
+        {
+            return base.Execute(testMethod);
+        }
+        if (CheckThatPlaywrightTimeoutIsSmallerThanNUnitTimeout(testMethod, timeout.Value) is var result && result != null)
+        {
+            return result;
+        }
+
+        return ExecuteWithTimeoutAsync(testMethod, timeout.Value).GetAwaiter().GetResult();
+    }
+
+    private async Task<TestResult[]> ExecuteWithTimeoutAsync(ITestMethod testMethod, int timeout)
+    {
+        // 1. Race Test execution against the specified timeout.
+        // In MSTest base.Execute will internally run the TestInitialize and TestCleanup. In case of a timeout, we don't
+        // run TestCleanup, only TestInitialize. To get good error messages, we close the browser context inside ContextCloseHookOnTimeout.
+        var testExecutionTask = Task.Run(() => base.Execute(testMethod));
+        var timeoutTask = Task.Delay(timeout);
+        await Task.WhenAny(testExecutionTask, timeoutTask);
+
+        // 2. Close the BrowserContext
+        if (ContextCloseHookOnTimeout != null)
+        {
+            await ContextCloseHookOnTimeout().ConfigureAwait(false);
+        }
+
+        // 3. If timeout was reached, we need to wait for the test execution to finish, since Page.Click could hang.
+        if (timeoutTask.IsCompleted)
+        {
+            // 3.1 Wait for the test execution to complete (let e.g. Page.ClickAsync throw)
+            // and give the test 5 seconds to finish.
+            await Task.WhenAny(testExecutionTask, Task.Delay(5000));
+        }
+
+        // 4. If the test execution is still running, we know it was a timeout.
+        if (!testExecutionTask.IsCompleted)
+        {
+            return new TestResult[]
+            {
+                new TestResult()
+                {
+                    Outcome = UnitTestOutcome.Timeout,
+                    TestFailureException = new Exception($"Test timed out after {timeout}ms.")
+                },
+            };
+        }
+
+        return await testExecutionTask;
+    }
+
+    private TestResult[]? CheckThatPlaywrightTimeoutIsSmallerThanNUnitTimeout(ITestMethod testMethod, int playwrightTestTimeout)
+    {
+        var timeout = MSTestProvider.TestTimeout;
+        var timeoutAttribute = testMethod.MethodInfo.GetCustomAttributes(typeof(TimeoutAttribute), false).FirstOrDefault();
+        if (timeoutAttribute != null)
+        {
+            timeout = ((TimeoutAttribute)timeoutAttribute).Timeout;
+        }
+
+        if (timeout != null && timeout != 0 && playwrightTestTimeout > timeout)
+        {
+            return new TestResult[]
+            {
+                new TestResult()
+                {
+                    Outcome = UnitTestOutcome.Failed,
+                    TestFailureException = new Exception($"Playwright test timeout ({playwrightTestTimeout}ms) is larger than MSTest timeout ({timeout}ms).")
+                },
+            };
+        }
+        return null;
+    }
+}
--- a/src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
+++ b/src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Playwright.MSTest;
 
 internal class PlaywrightTimeoutAttribute : TestMethodAttribute
 {
-    static internal Func<Task>? ContextCloseHookOnTimeout { get; set; } = null;
+    static internal Func<Task>? ContextCloseHookOnTimeoutAsync { get; set; } = null;
 
     public override TestResult[] Execute(ITestMethod testMethod)
     {
@@ -43,7 +43,7 @@ internal class PlaywrightTimeoutAttribute : TestMethodAttribute
         {
             return base.Execute(testMethod);
         }
-        if (CheckThatPlaywrightTimeoutIsSmallerThanNUnitTimeout(testMethod, timeout.Value) is var result && result != null)
+        if (CheckThatPlaywrightTimeoutIsSmallerThanNUnitTimeout(testMethod, timeout.Value) is TestResult[] result)
         {
             return result;
         }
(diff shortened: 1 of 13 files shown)


src/Playwright.NUnit/TimeoutCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+        // cancel by that the underlying BrowserContext calls.
+        if (timeoutTask.IsCompleted)
+        {
+            // 2.1 Call the test teardown handlers.
+            await RunSetUpOrTearDownMethod(context, isTearDown: true);
+            TearDownRan = true;
+
+            // 2.2 Wait for the test execution to complete (let e.g. Page.ClickAsync throw)
+            // and give the test 5 seconds to finish.
+            await Task.WhenAny(testExecutionTask, Task.Delay(5000));
```
mxschmitt, line 52: look at this

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+                }
+            }
+            finally
+            {
+                CleanupMethods.Clear();
+            }
+
+            // 2.2 Wait for the test execution to complete (let e.g. Page.ClickAsync throw)
+            // and give the test 5 seconds to finish.
+            await Task.WhenAny(testExecutionTask, Task.Delay(5000));
```
dgozman, line 80: Drop this one please.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        if (timeout != null && timeout != 0 && playwrightTestTimeout > timeout)
+        {
+            result = new TestResult[]
+            {
+                new TestResult()
+                {
+                    Outcome = UnitTestOutcome.Failed,
+                    TestFailureException = new Exception($"Playwright test timeout ({playwrightTestTimeout}ms) is larger than MSTest timeout ({timeout}ms).")
```
dgozman, line 114: Is it smaller or larger? How should it be? I find naming here somewhat confusing.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+internal class PlaywrightTimeoutAttribute : TestMethodAttribute
+{
+    static internal List<Func<Task>> CleanupMethods { get; set; } = new List<Func<Task>>();
+
+    public override TestResult[] Execute(ITestMethod testMethod)
+    {
+        var timeout = PlaywrightSettingsProvider.TestTimeout;
+        if (timeout == null || Debugger.IsAttached)
+        {
+            return base.Execute(testMethod);
```
dgozman, line 44: This misses `CleanupMethods.Clear()`.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+        await Task.WhenAny(testExecutionTask, timeoutTask);
+
+        // 2. If timeout was reached, call the tearDownHandlers and
+        // cancel by that the underlying BrowserContext calls.
+        if (timeoutTask.IsCompleted)
+        {
+            // 2.1 Call the test teardown handlers.
+            try
+            {
+                foreach (var cleanupMethod in CleanupMethods)
```
dgozman, line 68: Shouldn't we call these when timeout did not happen?

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+                {
+                    await cleanupMethod();
```
dgozman, line 70: So this only cleans up our context, but does not run other `[TestCleanup]` methods. Is there a way to do that?

src/Playwright.TestAdapter/MSTestProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+using System.Xml;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+
+namespace Microsoft.Playwright.TestAdapter;
+
+/// <summary>
+/// MSTest does not provide a way to get the global timeout value during a test execution.
```
dgozman, line 32: Can we get a timeout for the current test instead?

src/Playwright.NUnit/TimeoutCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+            return;
+        }
+
+        context.CurrentResult = context.CurrentTest.MakeTestResult();
+        context.CurrentResult.RecordException(new TimeoutException(timeoutExceptionMessage));
+        return;
+    }
+
+    protected async Task RunSetUpOrTearDownMethod(TestExecutionContext context, bool isSetUp = false, bool isTearDown = false)
```
dgozman, line 74: I feel uncomfortable with all the internal details of NUnit we are touching here.

src/Playwright.NUnit/TimeoutCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+        if (PlaywrightTimeoutIsSmallerThanNUnitTimeout(context, timeout.Value, out var result))
+        {
+            return result!;
+        };
+
+        ExecuteWithTimeoutAsync(context, timeout.Value).GetAwaiter().GetResult();
+        return context.CurrentResult;
+    }
+
+    private async Task ExecuteWithTimeoutAsync(TestExecutionContext context, int timeout)
```
dgozman, line 34: This needs a comment that explains whether this function runs setups and teardowns.

src/Playwright.NUnit/PlaywrightTestAttribute.cs
```diff
@@ -76,6 +83,13 @@ public override TestResult Execute(TestExecutionContext context)
                 {
                     break;
                 }
+                // NUnit takes care about running TearDown for the last iteration. PlaywrightTestAttribute inherits from
+                // the TimeoutCommand which runs TearDown for us in a case of a timeout.
+                if (!TearDownRan)
```
dgozman, line 88: I think merging `RetryTestCommand` and `TimeoutCommand` could lead to easier code, wdyt?

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
mxschmitt: I can rephrase it to e.g. `Playwright.TestTimeout is larger than MSTest.TestTimeout, make sure to have a larger Playwright timeout otherwise test execution gets terminated`

src/Playwright.NUnit/PlaywrightTestAttribute.cs
mxschmitt: I made an attempt, seems better already.

src/Playwright.TestAdapter/MSTestProvider.cs
mxschmitt: Its unfortunately not possible: https://cs.github.com/microsoft/testfx/blob/e5bc0e46f6d2cc8f0465596ff77ff18bb6680253/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs#L718

not exposed to the public, after a discussion with Jakub this seems the best way to determine it.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
mxschmitt: we do it already here: https://github.com/mxschmitt/playwright-dotnet/blob/45c4e554b95a422c79afa40b106b5167b83cfb78/src/Playwright.MSTest/ContextTest.cs#L44-L51

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            return result!;
+        }
+
+        return ExecuteWithTimeoutAsync(testMethod, timeout.Value).GetAwaiter().GetResult();
+    }
+
+    private async Task<TestResult[]> ExecuteWithTimeoutAsync(ITestMethod testMethod, int timeout)
+    {
+        // 1. Race Test execution against the specified timeout.
+        var testExecutionTask = Task.Run(() => base.Execute(testMethod));
```
mxschmitt, line 64: dig into if we can replace the method to handle the timeout inside so teardown will run after a timeout happened.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                CleanupMethods.Clear();
+            }
+
+            // 2.2 Wait for the test execution to complete (let e.g. Page.ClickAsync throw)
+            // and give the test 5 seconds to finish.
+            await Task.WhenAny(testExecutionTask, Task.Delay(5000));
+        }
+
+        if (testExecutionTask.IsCompleted)
```
mxschmitt, line 90: swap with below

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private async Task<TestResult[]> ExecuteWithTimeoutAsync(ITestMethod testMethod, int timeout)
+    {
+        // 1. Race Test execution against the specified timeout.
+        // In MSTest base.Execute will internally run the TestInitialize and TestCleanup. In case of a timeout, we don't
+        // run TestCleanup, only TestInitialize. To get good error messages, we close the browser context inside ContextCloseHookOnTimeout.
+        var testExecutionTask = Task.Run(() => base.Execute(testMethod));
+        var timeoutTask = Task.Delay(timeout);
+        await Task.WhenAny(testExecutionTask, timeoutTask);
```
nohwnd, line 61: This returns the task that completed that you don't capture. I am not sure if you don't use it on purpose because it made the code read worse, or if you don't know about it. 

That said I don't see any problem with checking the `.IsCompleted` on the task. The worst thing that I can see happen is that you await the test again in 3. if the timeout and test both completed because ContextCloseHookOnTimeout took enough time for both of them to complete. Wrapping a completed task into another task and awaiting it again has not functional issue, other than a tiny tiny added overhead.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        // 2. Close the BrowserContext
+        if (ContextCloseHookOnTimeout != null)
+        {
+            await ContextCloseHookOnTimeout().ConfigureAwait(false);
```
nohwnd, line 66: Would be nice to see some kind of cancellation of the close hook, but maybe I am overthinking it.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        if (timeout != null && timeout != 0 && playwrightTestTimeout > timeout)
+        {
+            return new TestResult[]
+            {
+                new TestResult()
+                {
+                    Outcome = UnitTestOutcome.Failed,
+                    TestFailureException = new Exception($"Playwright test timeout ({playwrightTestTimeout}ms) is larger than MSTest timeout ({timeout}ms).")
```
nohwnd, line 109: Maybe [InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception) could be used here.

It would be nice to see the more verbose explanation you proposed above to know why Playwright timeout should be larger. I would also use longer rather than larger, as we are referring to time.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,115 @@
...
+namespace Microsoft.Playwright.MSTest;
+
+internal class PlaywrightTimeoutAttribute : TestMethodAttribute
+{
+    static internal Func<Task>? ContextCloseHookOnTimeout { get; set; } = null;
+
+    public override TestResult[] Execute(ITestMethod testMethod)
+    {
+        var timeout = PlaywrightSettingsProvider.TestTimeout;
+        if (timeout == null || Debugger.IsAttached)
```
nohwnd, line 42: I get the intent of not timing out under debugger, but it makes it impossible to debug the timeout flow. It would be nice to have a ENV variable that you can set and make it go into the async flow. You will have to pump up your timeouts in the test if you want to be able to go step through it, but to me it looks useful when you are assisting someone with their issue related to timeouts.

But feel free to ignore my suggestions if you think feature creep is bigger issue than not being able to debug this.

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Playwright.TestAdapter;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Playwright.MSTest;
+
+internal class PlaywrightTimeoutAttribute : TestMethodAttribute
+{
+    static internal Func<Task>? ContextCloseHookOnTimeout { get; set; } = null;
```
nohwnd, line 37: Rename ContextCloseHookOnTimeoutAsync? I can see it being invoked but not awaited being an easy error to make in your code.

src/Playwright.NUnit/ContextTest.cs
```diff
@@ -41,4 +41,11 @@ public async Task ContextSetup()
     {
         Context = await NewContext(ContextOptions()).ConfigureAwait(false);
     }
+
+    [TearDown]
+    public async Task ContextTearDown()
+    {
+        await Context.CloseAsync().ConfigureAwait(false);
+        Context = null!;
```
nohwnd, line 49: I could not find why this is done, and it seems a bit obscure to force break your nullability. Are you checking it for null somewhere, or are you just trying to get it garbage collected earlier?

src/Playwright.NUnit/PlaywrightTestAttribute.cs
```diff
@@ -130,6 +144,117 @@ private static string Indent(string text, int indent)
...
+
+        private (TestResult, bool) ExecuteWithTimeout(TestExecutionContext context)
+        {
+            var timeout = PlaywrightSettingsProvider.TestTimeout;
+            if (timeout == null || Debugger.IsAttached)
+            {
+                return (innerCommand.Execute(context), false);
+            }
+
+            if (CheckThatPlaywrightTimeoutIsSmallerThanNUnitTimeout(context, timeout.Value) is var result && result != null)
```
nohwnd, line 156: nit: if you don't use `var` the is check implies not null. `is TestResult result` would automatically mean that `result` has value. 

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7NAExAGoAfAAQCYAGAWAChqBGRthxgFRgGcMAlXjgA2GAARQxAXjF4RwgNzsKzAJwAKSQEseY7n0E8R42EdEBKJRyZU9vAUNFiA3gF8gA==

mxschmitt: cool

src/Playwright.MSTest/PlaywrightTimeoutAttribute.cs
mxschmitt: I'll follow-up.

PR #2326, opened 2022-09-22
chore: commit automated VS changes

mxschmitt:


src/Playwright/Playwright.csproj
```diff
@@ -38,15 +38,15 @@
...
+    <None Include=".drivers\linux\node" Link=".playwright\node\linux-x64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-x64" />
+    <None Include=".drivers\linux-arm64\node" Link=".playwright\node\linux-arm64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-arm64" />
+    <None Include=".drivers\mac\node" Link=".playwright\node\mac\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\mac" />
+    <None Include=".drivers\win32_x64\node.exe" Link=".playwright\node\win32_x64\node.exe" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\win32_x64" />
+    <None Include="Scripts\playwright.sh" Link=".playwright\node\linux-x64\playwright.sh" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-x64" />
+    <None Include="Scripts\playwright.sh" Link=".playwright\node\linux-arm64\playwright.sh" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-arm64" />
```
nohwnd, line 46: Looks okay, if those things are ending up in the same place as they were before.

src/Playwright.Examples/Playwright.Examples.csproj
```diff
@@ -7,6 +7,7 @@
     <Nullable>enable</Nullable>
     <LangVersion>10</LangVersion>
     <NoWarn>CA2007</NoWarn>
+    <ReleaseVersion>0.0.0</ReleaseVersion>
```
nohwnd, line 10: Not sure what this means, to set dll version we use <Version> <FileVersion> <AssemblyVersion> (list here https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#generateassemblyinfo ) this is probably VS mac specific?

mxschmitt: Playwright.Examples is just an internal project which nobody does consume except we for testing sometimes. Should be fine then.

PR #2330, opened 2022-09-26
test: use Usings.cs for tests / ImplicitUsings

mxschmitt:


src/.editorconfig
```diff
@@ -170,6 +170,7 @@ indent_size = 2
 end_of_line = lf
 
 [*.cs]
+dotnet_diagnostic.IDE0005.severity = error
```
nohwnd, line 173: Not a big fan of erroring out because of non-functional problems. We often make stuff like this (extra usings) a warning, and in CI build we use warnAsError

nohwnd (approved):

PR #2333, opened 2022-10-03
feat: pass additional environment variables

wodyjowski:


mxschmitt (commented):
Instead of having only env, it would be better to have a class as options like `PlaywrightCreateOptions` which does similar things like our other options: https://github.com/microsoft/playwright-dotnet/blob/main/src/Playwright/API/Generated/Options/PageGotoOptions.cs

So in the future we can add more needed options if necessary.

mxschmitt:
Are you interested in continuing to work on this patch? Otherwise I'd close it.

wodyjowski:
Yes, I'll do it tomorrow

wodyjowski:
@mxschmitt how about that?

src/Playwright/Configuration/PlaywrightCreateOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    public DebugMode DebugMode { get; set; }
+
+    /// <summary>
+    /// Defines the level of detail in debug logs outputted by Playwright.<br />
+    /// "*" - displays all available logs.<br />
+    /// "pw:browser*" - filters the logs starting with <i>pw:browser</i>.<br />
+    /// Empty string disables logging.
+    /// </summary>
+    public string LoggingConfiguration { get; set; }
```
mxschmitt, line 55: Instead of going the full managed mode, I would recommend to make it more flexible like that.

```suggestion
    /// <summary>
    /// Allows setting custom environment variables.
    /// </summary>
    public Dictionary<string, string> EnvironmentVariables { get; set; }
```

mxschmitt:
Closing since this looks stale.

PR #2337, opened 2022-10-06
chore(roll): roll Playwright to 1.28.0-alpha-oct-6-2022

mxschmitt:


pavelfeldman (commented):
Don't do two things at a time. First remove extra using statements, then apply the roll so that it was possible to review it!

src/Playwright/Core/Page.cs
```diff
@@ -1160,4 +1160,40 @@ await Context.RecordIntoHarAsync(har, this, new()
         }).ConfigureAwait(false);
         await harRouter.AddPageRouteAsync(this).ConfigureAwait(false);
     }
+
+    public ILocator GetByAltText(string text, PageGetByAltTextOptions options = null)
```
pavelfeldman, line 1164: Page implementation should delegate to the mainframe instead.

src/Playwright/Core/Locator.cs
```diff
@@ -51,22 +54,23 @@ public Locator(Frame parent, string selector, LocatorLocatorOptions options = nu
 
         if (options?.HasTextRegex != null)
         {
-            var jsRegex = $"/{options.HasTextRegex.ToString()}/{options.HasTextRegex.Options.GetInlineFlags()}";
-            _selector += $" >> has={JsonSerializer.Serialize("text=" + jsRegex, serializerOptions)}";
+            var textSelector = "text=" + EscapeForTextSelector(options.HasTextRegex, false);
```
pavelfeldman, line 57: nit: You don't need the branching anymore since EscapeForTextSelector takes care of it

src/Playwright/Core/Locator.cs
```diff
@@ -338,4 +342,249 @@ private async Task WithElementAsync(Func<IElementHandle, float?, Task> callback,
...
+    internal static void SetTestIdAttribute(string attributeName)
+        => _testIdAttributeName = attributeName;
+
+    internal static string GetByTestIdSelector(string testId)
+           => GetByAttributeTextSelector(_testIdAttributeName, testId, exact: true);
+
+    internal static string GetByAttributeTextSelector(string attrName, string text, bool? exact)
+    {
+        var exactFlag = (exact == true) ? "s" : "i";
+        return $"internal:attr=[{attrName}={EscapeForAttributeSelector(text)}{exactFlag}]";
```
pavelfeldman, line 391: Nit: you should follow up with upstream's exact as a part of escapeForAttributeSelector and also no 's' by default.

src/Playwright/Core/Locator.cs
```diff
@@ -338,4 +342,249 @@ private async Task WithElementAsync(Func<IElementHandle, float?, Task> callback,
...
+        {
+            props.Add(new List<string> { "include-hidden", options.IncludeHidden.ToString() });
+        }
+        if (options.Level != null)
+        {
+            props.Add(new List<string> { "level", options.Level.ToString() });
+        }
+        if (options.NameString != null)
+        {
+            props.Add(new List<string> { "name", options.NameString });
```
pavelfeldman, line 456: You need to escape the string and regex, use EscapeForAttributeSelector here

src/Playwright/Core/Locator.cs
```diff
@@ -338,4 +342,249 @@ private async Task WithElementAsync(Func<IElementHandle, float?, Task> callback,
...
+        }
+        else if (options.NameRegex != null)
+        {
+            props.Add(new List<string> { "name", $"/{options.NameRegex}/{options.NameRegex.Options.GetInlineFlags()}" });
```
pavelfeldman, line 460: ditto

src/Playwright/Core/Locator.cs
```diff
@@ -338,4 +342,249 @@ private async Task WithElementAsync(Func<IElementHandle, float?, Task> callback,
...
+        return $"role={role}{string.Concat(props.Select(p => $"[{p[0]}={p[1]}]"))}";
+    }
+
+    private static string EscapeForAttributeSelector(string value)
+    {
+        // TODO: this should actually be
+        //   cssEscape(value).replace(/\\ /g, ' ')
+        // However, our attribute selectors do not conform to CSS parsing spec,
+        // so we escape them differently.
+        return $"\"{value.Replace("\"", "\\\"")}\"";
```
pavelfeldman, line 475: should get `bool exact`

src/Playwright/Core/Locator.cs
mxschmitt: following up with role stuff.

PR #2361, opened 2022-10-31
chore: use ResolveAssemblyReferences instead of BeforeBuild

### LINKED ISSUE #2354: [BUG] .playwright folder not copied on dotnet publish from v1.19.0 onwards
**Context:**
- Playwright Version: was using v1.18.1 then tried v1.19.0 and found the issue in v1.19.0
- Operating System: Linux (github actions) but can confirm issue on Windows
- .NET version: 6.0
- Browser: n/a but Chromium
- Extra: noticed issue in github actions (linux ubuntu) but can also confirm issue on windows 10

So I've been running playwright chromium in Azure Functions (Linux isolated consumption plan) successfully for a while now and have created [a template](https://github.com/MrSimonC/Playwright-In-Azure-Function-Template) which was working until I tried to use it recently. When using playwright-dotnet v1.27.1 I came across this new error at runtime:

`Microsoft.Playwright.PlaywrightException: Driver not found: /home/.playwright/node/linux/playwright.sh`
`at Microsoft.Playwright.Helpers.Driver.GetExecutablePath()`

I've done a thorough investigation on this:

* Tracing dates back to a working version, I found the template/code
  * Worked at tag 1.18.1
  * Stopped working tag 1.19.0

So comparing those tags:

* There were no code changes in Driver.GetExecutablePath()
  * Yet it blows up here - meaning the logic is sound, but it's not finding files it's looking for
* Running a few github actions with folder outputs, I can see that `.playwright` is included in the publish folder in v1.18.1 but not v1.19.0
* Since `Driver.cs` is doing a `Path.Combine(driversPath, ".playwright", "node", platformId, runnerName);` [here](https://github.com/microsoft/playwright-dotnet/blob/858354b1cafdf9c5a4e674159182ac88ac185d1c/src/Playwright/Helpers/Driver.cs#L84) ... it means we need it in the published folder of our app.
* There *were* changes to the `Microsoft.Playwright.targets` file from #1955 
* I think the [removal of this line](https://github.com/microsoft/playwright-dotnet/pull/1955/files#diff-e3082407b006296971300d96ce43253f11b1754c334c6b7e7b81191d0a5a37fbL29) (and associated line 26 above it) in #1955 is the culprit which means that we don't see the `.playwright` included in the publish folder when doing `dotnet publish`.
* Even on windows, doing `dotnet restore`, `dotnet build` and `dotnet publish` locally against v1.18.1 and v1.19.0 I can confirm the `.playwright` folder is missing after the publish step in the publish folder between these versions.

Could I please ask if

a) you think my logic is sound?  
b) someone could raise a PR against this to fix it?

campersau:
Which will run the `CopyPlaywrightFilesToOutput` target also on `publish --no-build`.

Fixes #2354

### DIFF
--- a/src/Playwright/build/Microsoft.Playwright.targets
+++ b/src/Playwright/build/Microsoft.Playwright.targets
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Target Name="CopyPlaywrightFilesToOutput" AfterTargets="BeforeBuild">
+  <Target Name="CopyPlaywrightFilesToOutput" AfterTargets="ResolveAssemblyReferences">
     <PropertyGroup>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">%(RuntimePack.RuntimeIdentifier)</PlaywrightPlatform>
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == ''">$(RuntimeIdentifier)</PlaywrightPlatform>


mxschmitt (approved):
LGTM, thanks!

quldude:
@mxschmitt @campersau will this fix create new nuget package for Microsoft.Playwright.NUnit?

PR #2364, opened 2022-11-02
Environment Variabl PLAYWRIGHT_DRIVER_PATH

### LINKED ISSUE #2363: [Question]: Is it possible to configure a different node.exe path outside of the compiled Playwright DLL?
### Your question

In Driver.CS, the GetExecutablePath returns the path of the playwright.cmd file. The playwright.cmd file starts node runtime.

If I leave the playwright.cmd file in its original folder and update the playwright.cmd file to open node.exe from a different folder path such as a centralized location that is not within the app folder, I assumed that this would allow playwright to start with a centralized path to node.exe, However, this doesn't work so there must be other circumstances preventing this.

**I left this path/file in place and modified the cmd file to point to a centralized .playwright folder that contains node.exe**
C:\Users\userID\source\repos\MsPlaywrightTestHarness\\.playwright\node\win32_x64\playwright.cmd

**An example of a centralized folder that contains the node files that I pointed to is below -** 
C:\Users\userID\AppData\Roaming\automations\\.playwright

Is it possible to configure a different node.exe path outside of the compiled Playwright DLL? If so, what other file needs to be updated other than playwright.cmd?

From previous questions/responses on this topic, I know this issue isn't on the radar to be addressed, I'm just trying to find options that I can use to change the path where node is launched from so that we don't have multiple instances of this node folder for every playwright automation that we install on an end-user desktop. Currently we decompile the playwright DLL, make changes to allow that path of our choice (a customized "GetExecutablePath") then recompile the DLL which has been a pain to manage. 

Playwright has been a great automation tool for us compared to other tools, we'd just like to streamline the installation process while removing multiple instances of node.

Thanks

feiyun0112:
fix  #2363

### DIFF
--- a/src/Playwright/Helpers/Driver.cs
+++ b/src/Playwright/Helpers/Driver.cs
@@ -34,7 +34,13 @@ internal static class Driver
 {
     internal static string GetExecutablePath()
     {
-        DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
+        string baseDirectory = Environment.GetEnvironmentVariable("PLAYWRIGHT_DRIVER_PATH");
+        if (string.IsNullOrEmpty(baseDirectory))
+        {
+            baseDirectory = AppContext.BaseDirectory;
+        }
+
+        DirectoryInfo assemblyDirectory = new(baseDirectory);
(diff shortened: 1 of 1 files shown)


mxschmitt:
Since the issue it fixes seems fixed and the issue author happy, I'm closing it for now. Passing an env variable seems more like a workaround rather than an actual implementational fix.

PR #2376, opened 2022-11-08
fix: make evaluate work with serializing Guids

### LINKED ISSUE #2374: [Bug] [Regression] [since 1.23] Parsing Guid values (IPage.EvaluateAsync<Guid>) using System.Convert class throws
This created very problematic regression for parsing Guid values.
`Convert.ChangeType("{08d15efd-bd4f-4244-b46e-2238964e0e36}", typeof(Guid), System.Globalization.CultureInfo.InvariantCulture)` throws

_Originally posted by @maw136 in https://github.com/microsoft/playwright-dotnet/issues/2178#issuecomment-1306371464_

feiyun0112:
Fixes #2374

### DIFF
--- a/src/Playwright/Helpers/Driver.cs
+++ b/src/Playwright/Helpers/Driver.cs
@@ -34,7 +34,13 @@ internal static class Driver
 {
     internal static string GetExecutablePath()
     {
-        DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
+        string baseDirectory = Environment.GetEnvironmentVariable("PLAYWRIGHT_DRIVER_PATH");
+        if (string.IsNullOrEmpty(baseDirectory))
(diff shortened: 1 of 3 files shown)


src/Playwright/Helpers/Driver.cs
```diff
@@ -34,7 +34,13 @@ internal static class Driver
 {
     internal static string GetExecutablePath()
     {
-        DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
+        string baseDirectory = Environment.GetEnvironmentVariable("PLAYWRIGHT_DRIVER_PATH");
```
pavelfeldman, line 37: Looks like a stray change.

mxschmitt (approved):

PR #2384, opened 2022-11-10
fix: parsing of runsettings file

mxschmitt:


src/Playwright.TestAdapter/PlaywrightSettingsXml.cs
```diff
@@ -74,13 +74,15 @@ private static T ParseXmlIntoClass<T>(XmlReader reader) where T : class, new()
         var options = new T();
         while (reader.Read())
         {
-            if (reader.NodeType != XmlNodeType.Element)
+            if (reader.NodeType == XmlNodeType.EndElement)
```
aslushnikov, line 77: you should go until the end element of a given tag

aslushnikov (approved):

PR #2385, opened 2022-11-14
chore(roll): roll Playwright to 1.28.0-alpha-nov-11-2022

mxschmitt:


src/Playwright/Core/EventUtils.cs
```diff
@@ -0,0 +1,107 @@
...
+            UpdateEventSubscription(eventName, true);
+        }
+        if (!_eventNameToHandlers.ContainsKey(eventName))
+        {
+            _eventNameToHandlers[eventName] = handler;
+        }
+        else
+        {
+            _eventNameToHandlers[eventName] = Delegate.Combine(_eventNameToHandlers[eventName], handler);
+        }
```
campersau, line 75: Nit: also use `TryGetValue` here

```suggestion
        if (!_eventNameToHandlers.TryGetValue(eventName, out var handlers))
        {
            _eventNameToHandlers[eventName] = handler;
        }
        else
        {
            _eventNameToHandlers[eventName] = Delegate.Combine(handlers, handler);
        }
```

dgozman (approved):

PR #2389, opened 2022-11-16
chore: roll to 1.28.0-beta-1668538774000

mxschmitt:
- Add GetByRole Exact
- align selectors handling with upstream
- remove legacy namespace declaration in some of the channels (the rest is good already) moved out into https://github.com/microsoft/playwright-dotnet/pull/2391.

### DIFF
--- a/src/Playwright/Core/Locator.cs
+++ b/src/Playwright/Core/Locator.cs
@@ -374,7 +374,7 @@ internal class Locator : ILocator
         => ((ILocator)this).Locator(GetByRoleSelector(role, new(options)));
 
     public ILocator GetByTestId(string testId)
-        => ((ILocator)this).Locator(GetByTestIdSelector(testId));
+        => ((ILocator)this).Locator(GetByTestIdSelector(TestIdAttributeName(), testId));
 
     public ILocator GetByText(string text, LocatorGetByTextOptions options = null)
         => ((ILocator)this).Locator(GetByTextSelector(text, options?.Exact));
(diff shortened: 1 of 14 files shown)


src/Playwright/Transport/Protocol/Generated/PlaywrightInitializer.cs
```diff
@@ -45,7 +45,7 @@ internal class PlaywrightInitializer
     public List<DeviceDescriptorEntry> DeviceDescriptors { get; set; }
 
     [JsonPropertyName("selectors")]
-    public Core.Selectors Selectors { get; set; }
+    public Core.SelectorsOwner Selectors { get; set; }
```
mxschmitt, line 48: This will get overridden in the next protocol roll, need to think how we should do it. Maybe just a hacky if inside our Generator.

src/Playwright/Core/Locator.cs
```diff
@@ -388,11 +388,13 @@ public ILocator GetByTitle(string text, LocatorGetByTitleOptions options = null)
...
 
+    internal static string TestIdAttributeName() => _testIdAttributeName;
+
     internal static void SetTestIdAttribute(string attributeName)
         => _testIdAttributeName = attributeName;
 
-    internal static string GetByTestIdSelector(string testId)
-        => GetByAttributeTextSelector(_testIdAttributeName, testId, exact: true);
+    internal static string GetByTestIdSelector(string testIdAttributeName, string testId)
+        => $"internal:testid=[${testIdAttributeName}={EscapeForAttributeSelector(testId, true)}]";
```
dgozman, line 397:
```suggestion
        => $"internal:testid=[{testIdAttributeName}={EscapeForAttributeSelector(testId, true)}]";
```

src/Playwright/Core/Selectors.cs
```diff
@@ -22,32 +22,72 @@
...
-    private readonly SelectorsChannel _channel;
+    private readonly HashSet<SelectorsOwner> _channels = new();
+    private readonly List<SelectorsRegisterParams> _registrations = new();
 
-    internal Selectors(IChannelOwner parent, string guid) : base(parent, guid)
+    public async Task RegisterAsync(string name, SelectorsRegisterOptions options = default)
     {
-        _channel = new(guid, parent.Connection, this);
+        options ??= new SelectorsRegisterOptions();
+        var source = ScriptsHelper.EvaluationScript(options?.Script, options?.Path);
```
dgozman, line 41: Why `options?` here, given the previous line?

src/Playwright/Transport/Protocol/Generated/PlaywrightInitializer.cs
dgozman: We can just rename `Selectors` to `SelectorsAPI` and `SelectorsOwner` to `Selectors`.

dgozman (approved):

PR #2390, opened 2022-11-16
chore: review partial classes and merging / removing them

mxschmitt:


src/Playwright/Transport/Channels/ArtifactChannel.cs
```diff
@@ -20,83 +20,71 @@
...
-    internal class ArtifactChannelImpl : Channel<Artifact>
+    internal class ArtifactChannel : Channel<Artifact>
     {
-        public ArtifactChannelImpl(string guid, Connection connection, Artifact owner) : base(guid, connection, owner)
+        public ArtifactChannel(string guid, Connection connection, Artifact owner) : base(guid, connection, owner)
         {
         }
 
-        internal virtual async Task<string?> PathAfterFinishedAsync()
+        internal async Task<string> PathAfterFinishedAsync()
```
dgozman, line 38: Why return type change?

dgozman (approved):

src/Playwright/Transport/Channels/ArtifactChannel.cs
mxschmitt: `#nullable enable` which we have enabled in our generated types, but by default everything is nullable in c#.

PR #2396, opened 2022-11-17
chore: make SendMessageToServerAsync payload type more strict

mxschmitt:


src/Playwright/Transport/Connection.cs
```diff
@@ -118,30 +117,11 @@ public void Dispose()
...
-                {
-                    if (kv.Value != null)
-                    {
-                        sanitizedArgs.Add(kv.Key, kv.Value);
-                    }
-                }
-            }
-            else
-            {
-                foreach (PropertyDescriptor propertyDescriptor in TypeDescriptor.GetProperties(args))
```
mxschmitt, line 135: We before were able to pass anything, this caused some issues sometimes when passing `Dictionary<string, string>` instead of `Dictionary<string, object>` which were not serialised correctly.

src/Playwright/Transport/Channels/ArtifactChannel.cs
```diff
@@ -46,9 +47,9 @@ internal async Task SaveAsAsync(string path)
         => await Connection.SendMessageToServerAsync<JsonElement>(
             Guid,
             "saveAs",
-            new
+            new Dictionary<string, object>
```
dgozman, line 50: This looks like a `Dictionary<string, string>`

src/Playwright/Transport/Channels/Root.cs
```diff
@@ -39,9 +40,9 @@ internal Root(IChannelOwner parent, Connection connection, string guid) : base(p
 
     internal async Task<PlaywrightImpl> InitializeAsync()
     {
-        var args = new
+        var args = new Dictionary<string, object>
```
dgozman, line 43: `<string, string>`?

src/Playwright/Transport/Connection.cs
dgozman: Can we add a test?

dgozman (approved):

src/Playwright/Transport/Channels/Root.cs
mxschmitt: Parameters are always `<string, object>` ensured on the type level.

src/Playwright/Transport/Channels/ArtifactChannel.cs
mxschmitt: ditto

src/Playwright/Transport/Connection.cs
mxschmitt: It's ensured on the type level, before we were just able to pass it now its saver.

There is still a chance that someone passes a class instance in an object, but less lower the chance. With we I mean there us.

PR #2398, opened 2022-11-17
chore: move SendMessageToServerAsync calls to ChannelOwner

mxschmitt:
https://github.com/microsoft/playwright-dotnet/pull/2397 gets merged before so only https://github.com/microsoft/playwright-dotnet/commit/e5b6c023bbc83550e7169d3fa0f2fe25b4b217b7 is relevant.


dgozman (approved):
This looks good, but it only makes sense if we are merging channels into channel owners. Let's wait with landing this until the merging path is clear.

PR #2419, opened 2022-12-06
Use .Net 7 SDK as base image

joshendriks:
As a followup for both #2411 and #2410 

Context:
I am using this image as part of a bitbucket pipeline where Is execute "xUnit driven playwright test" as follows:
```
image: nwwz/playwright-dotnet:v1.28.0-net7-jammy
       caches:
        - dotnetcore
       script:
        - dotnet dev-certs https --clean
        - dotnet dev-certs https --trust
        - export TESTCONTAINERS_RYUK_DISABLED=true
        - dotnet restore
        - cd tests/End2EndTests
        - dotnet test
       services:
        - docker
```

I had to replace the official playwright image (_mcr.microsoft.com/playwright/dotnet:v1.27.0-focal_) with my own (_nwwz/playwright-dotnet:v1.28.0-net7-jammy_) as part of migrating my app to .NET 7.

_nwwz/playwright-dotnet:v1.28.0-net7-jammy_ was built using the same change as made in this PR.

### DIFF
--- a/utils/docker/Dockerfile.jammy
+++ b/utils/docker/Dockerfile.jammy
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/sdk:6.0-jammy
+FROM mcr.microsoft.com/dotnet/sdk:7.0-jammy
 
 ARG DEBIAN_FRONTEND=noninteractive
 ARG TZ=America/Los_Angeles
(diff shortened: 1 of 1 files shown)


joshendriks:
I assume you mean the current latest LTS? At the time of writing 3.1 is still a current LTS. Ass of nov 2023 we will gain be in that situattion where there are two versions in LTS.

Regardless, I do think it makes sense to also support .NET 7. It is fully supported and production ready. Also .Net 7 projects do not build on .Net 6. The otherway around does work. So having .NET 7 would give a brroader support to the users

Both of the above reasons might be enough of a justification to support multiple .NET versions. Would you like me to create an issue to dicuss it there or would your statement above remain the same?

mxschmitt:
Yes you are right, I assume the [latest LTS](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle).

Definitely it makes sense to have a more rich matrix for different .net versions across the Docker images. Feel free to file a feature request as an issue, then we can do it based on the user requests.

Issue #2351, opened 2022-10-18
[Feature] API class-coverage, CSS and JS coverage


Zetanova:
The chromium feature CSS and JS coverage is only implemented in NodeJS
https://playwright.dev/docs/api/class-coverage

It is a great feature to reduce page size, but currently missing in playwridht-dotnet

nikokolevkolev:
Is there any progress on this ?

PR #2438, opened 2023-01-03
feat: Collect JS and CSS Coverage

hutchcodes:
Adds the ability to collect coverage information when using Chrome in the same way that Node version of Playwright does.

[Completes this ](https://github.com/microsoft/playwright-dotnet/issues/2351)


hutchcodes:
Abandoned in favor of just giving access to the CDP, allowing access to all of the Chrome DevTools
https://github.com/microsoft/playwright-dotnet/pull/2448

Issue #2009, opened 2022-02-12
[Feature] Add CDPSession


chenzanyu:
Node.js and Python have CDPSession implementations, can this function in DOTNET?
https://playwright.dev/docs/api/class-cdpsession

namilkimfree:
I want that feature.

304NotModified:
Correct link: https://playwright.dev/docs/api/class-cdpsession

wY1k:
I want that feature.

namilkimfree:
with stealth function

This function is required to use capture events, etc.

@mxschmitt

PR #2448, opened 2023-01-11
feat: add support for Chrome DevTools Protocol

hutchcodes:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2009.

### DIFF
--- a/src/Playwright.Tests/SessionTests.cs
+++ b/src/Playwright.Tests/SessionTests.cs
@@ -67,7 +67,7 @@ public class SessionTests : PageTestEx
         bool threwException = false;
         var client = await Page.Context.NewCDPSessionAsync(Page);
         await client.SendAsync("Runtime.enable");
-        var evalResponse = (await client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "1 + 2" }, { "returnByValue", true } }));
+        var evalResponse = await client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "1 + 2" }, { "returnByValue", true } });
 
         // TODO: Why is the result nested in a result?
         Assert.AreEqual(3, (int)evalResponse.Value.Deserialize<JsonNode>()["result"]["result"]["value"]);
--- a/src/Playwright.Tests/SessionTests.cs
+++ b/src/Playwright.Tests/SessionTests.cs
@@ -76,7 +76,7 @@ public class SessionTests : PageTestEx
 
         try
         {
-            await client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "'1 + 2'" }, { "returnByValue", "true" } });
+            await client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "1 + 2" }, { "returnByValue", "true" } });
         }
         catch (PlaywrightException ex)
         {
--- a/src/Playwright.Tests/SessionTests.cs
+++ b/src/Playwright.Tests/SessionTests.cs
@@ -50,11 +50,11 @@ public class SessionTests : PageTestEx
         await client.SendAsync("Network.enable");
 
         var events = new List<object>();
-        client.On("Network.requestWillBeSent", (eventArgs) =>
+        client.AddEventListener("Network.requestWillBeSent", (eventArgs) =>
         {
             events.Add(eventArgs);
         });
-
+        
         await Page.GotoAsync(Server.EmptyPage);
 
         Assert.AreEqual(1, events.Count);
@@ -64,26 +64,18 @@ public class SessionTests : PageTestEx
     [Skip(SkipAttribute.Targets.Firefox, SkipAttribute.Targets.Webkit)]
     public async Task ShouldDetachSession()
     {
-        bool threwException = false;
         var client = await Page.Context.NewCDPSessionAsync(Page);
         await client.SendAsync("Runtime.enable");
         var evalResponse = await client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "1 + 2" }, { "returnByValue", true } });
 
-        // TODO: Why is the result nested in a result?
-        Assert.AreEqual(3, (int)evalResponse.Value.Deserialize<JsonNode>()["result"]["result"]["value"]);
+        Assert.AreEqual(3, (int)evalResponse.Value.Deserialize<JsonNode>()["result"]["value"]);
 
         await client.DetachAsync();
 
-        try
-        {
-            await client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "1 + 2" }, { "returnByValue", "true" } });
-        }
-        catch (PlaywrightException ex)
-        {
-            threwException = true;
-            Assert.AreEqual("Target page, context or browser has been closed", ex.Message);
-        }
-        Assert.IsTrue(threwException);
+
+        var exceptions = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(
+            () => client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "'1 + 2'" }, { "returnByValue", "true" } }));
+        StringAssert.Contains("Target page, context or browser has been closed", exceptions.Message);
     }
 
     [PlaywrightTest("chromium/session.spec.ts", "should throw nice errors")]
@@ -92,15 +84,9 @@ public class SessionTests : PageTestEx
     {
         var client = await Page.Context.NewCDPSessionAsync(Page);
 
-        try
-        {
-            await TheSourceOfTheProblem();
-        }
-        catch (PlaywrightException ex)
-        {
-            Assert.IsTrue(ex.StackTrace.Contains("TheSourceOfTheProblem", StringComparison.InvariantCultureIgnoreCase), "StackTrace does not contain 'TheSourceOfTheProblem'");
-            Assert.IsTrue(ex.Message.Contains("ThisCommand.DoesNotExist", StringComparison.InvariantCultureIgnoreCase), "Exception message does not contain 'ThisCommand.DoesNotExist'");
-        }
+        var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => TheSourceOfTheProblem());
+        StringAssert.Contains("TheSourceOfTheProblem", exception.StackTrace);
+        StringAssert.Contains("ThisCommand.DoesNotExist", exception.Message);
 
         async Task TheSourceOfTheProblem()
         {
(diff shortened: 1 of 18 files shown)


src/Playwright/API/Generated/ICDPSession.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// emit any events and can't be used to send messages.
+    /// </para>
+    /// </summary>
+    Task DetachAsync();
+
+    /// <param name="method">Protocol method name.</param>
+    /// <param name="args">Optional method parameters.</param>
+    Task<JsonElement?> SendAsync(string method, Dictionary<string, object>? args = default);
+
+    void On(string eventName, Action<JsonElement?> handler);
```
mxschmitt, line 63: Was this method generated from the upstream docs? I think we need to add it there, so it does not get overridden when we roll in the future. This gives us also support for comments + makes it visible on playwright.dev.

src/Playwright/Core/CDPSession.cs
```diff
@@ -0,0 +1,90 @@
...
+using Microsoft.Playwright.Transport.Channels;
+
+#nullable enable
+
+namespace Microsoft.Playwright.Core;
+
+internal class CDPSession : ChannelOwnerBase, ICDPSession, IChannelOwner<CDPSession>
+{
+    private readonly CDPChannel _channel;
+    private readonly Dictionary<string, List<Action<JsonElement?>>> _eventListeners;
```
mxschmitt, line 39: Not sure if here a Delegate would fit better. Since it provides Remove and Combine methods.

src/Playwright/Core/CDPSession.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            _eventListeners[eventName].Add(handler);
+        }
+        else
+        {
+            _eventListeners.Add(eventName, new() { handler });
+        }
+    }
+
+    public Task<JsonElement> Send2Async(string method, Dictionary<string, object>? args = null) => throw new NotImplementedException();
```
mxschmitt, line 86: what is Send2Async?

src/Playwright/API/Generated/ICDPSession.cs
mxschmitt, line 63: Why JsonElement? is there a case that its null?

src/Playwright/Transport/Channels/CDPChannelEventArgs.cs
```diff
@@ -0,0 +1,37 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Darío Kondratiuk
```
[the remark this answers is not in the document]
kblok: 💔 😂

src/Playwright/Core/CDPSession.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public async Task<JsonElement?> SendAsync(string method, Dictionary<string, object>? args = null)
+    {
+        return await _channel.SendAsync(method, args);
+    }
+
+    private void OnCDPEvent(object sender, CDPChannelEventArgs e)
+    {
+        if (_eventListeners.ContainsKey(e.EventName))
```
campersau, line 65: Use `TryGetValue` instead to avoid double indexing.

src/Playwright/Core/CDPSession.cs
```diff
@@ -0,0 +1,90 @@
...
+            foreach (var listener in _eventListeners[e.EventName])
+            {
+                listener(e.EventParams);
+            }
+        }
+    }
+
+    public void On(string eventName, Action<JsonElement?> handler)
+    {
+        if (_eventListeners.ContainsKey(eventName))
```
campersau, line 76: Use `TryGetValue` instead to avoid double indexing.

src/Playwright.Tests/SessionTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    [Skip(SkipAttribute.Targets.Firefox, SkipAttribute.Targets.Webkit)]
+    public async Task ShouldDetachSession()
+    {
+        bool threwException = false;
+        var client = await Page.Context.NewCDPSessionAsync(Page);
+        await client.SendAsync("Runtime.enable");
+        var evalResponse = (await client.SendAsync("Runtime.evaluate", new Dictionary<string, object> { { "expression", "1 + 2" }, { "returnByValue", true } }));
+
+        // TODO: Why is the result nested in a result?
+        Assert.AreEqual(3, (int)evalResponse.Value.Deserialize<JsonNode>()["result"]["result"]["value"]);
```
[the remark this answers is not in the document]
mxschmitt: Modify the channel like this, then it works:

```csharp
    internal async Task<JsonElement?> SendAsync(string method, Dictionary<string, object>? args = null)
    {
        var newArgs = new Dictionary<string, object>() { { "method", method } };
        if (args != null)
        {
            newArgs["params"] = args;
        }
        var result = await Connection.SendMessageToServerAsync(
           Guid,
           "send",
           newArgs).ConfigureAwait(false);
        return result?.GetProperty("result");
    }
```

See how its done upstream: 

https://github.com/microsoft/playwright/blob/fa49803ca08aed6c8373a7101f54c49ad458ffd7/packages/playwright-core/src/client/cdpSession.ts#L46

src/Playwright/API/Generated/ICDPSession.cs
```diff
@@ -0,0 +1,66 @@
...
+/// </summary>
+public partial interface ICDPSession
+{
+    /// <summary>
+    /// <para>
+    /// Detaches the CDPSession from the target. Once detached, the CDPSession object won't
+    /// emit any events and can't be used to send messages.
+    /// </para>
+    /// </summary>
+    Task DetachAsync();
```
campersau, line 57: Would it be nicer to implement `IAsyncDisposable` instead? So you could write:

```c#
await using (var cdp = await browser.NewBrowserCDPSessionAsync())
{
    // ...
}

```

src/Playwright/API/Generated/ICDPSession.cs
campersau, line 63: Should there also be an `Off` method to remove the handler? Or return an `IDisposable` and when it is disposed remove the handler?

src/Playwright/Core/CDPSession.cs
hutchcodes: That was me fooling around trying to create some generic methods rather than just return JSonElement, but wasn't supposed to be in this PR. Removed.

campersau:
In another project, [`CefSharp`](https://github.com/cefsharp/CefSharp), we generated code based on the [CDP](https://chromedevtools.github.io/devtools-protocol/) of the browser which can be downloaded from `http://localhost:9222/json/protocol` which looks something like this: https://raw.githubusercontent.com/cefsharp/CefSharp/master/CefSharp/DevTools/DevToolsClient.Generated.netcore.cs
So the user can use actual types / events and sees the documentation instead passing strings / dictionaries around.

Would that also be something you would want to add / support?
(The CDP to csharp parser itself isn't public.)

src/Playwright/API/Generated/ICDPSession.cs
hutchcodes: We're getting a nullable passed to us from the Connection class.

hutchcodes: We changed from `On` to `AddEventListener` and also added a `RemoveEventListener`

src/Playwright/Transport/Channels/CDPChannelEventArgs.cs
```diff
@@ -0,0 +1,36 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Playwright.Transport.Channels;
+
+public class CDPChannelEventArgs
```
campersau, line 29: Should this be `internal`? Looks like everything else in this namespace is `internal`.
```suggestion
internal class CDPChannelEventArgs
```

src/Playwright.Tests/SessionTests.cs
```diff
@@ -0,0 +1,206 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+using System.Text.Json;
+using System.Text.Json.Nodes;
+
+namespace Microsoft.Playwright.Tests.Chromium;
```
mxschmitt, line 28: namspace looks wrong. Should be just `namespace Microsoft.Playwright.Tests;`

src/Playwright/Core/CDPSession.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            listener(e.EventParams);
+        }
+    }
+
+    public void AddEventListener(string eventName, Action<JsonElement?> eventHandler)
+    {
+        if (_eventListeners.TryGetValue(eventName, out Action<JsonElement?> listener))
+        {
+            _eventListeners[eventName] += eventHandler;
```
pavelfeldman, line 75: What's is Actions's +=, does it combine the delegates? Not a C# expert, but not following how multiple handlers work here.

src/Playwright/Core/CDPSession.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    private void OnCDPEvent(object sender, CDPChannelEventArgs e)
+    {
+        if (_eventListeners.TryGetValue(e.EventName, out Action<JsonElement?> listener))
+        {
+            listener(e.EventParams);
+        }
+    }
+
+    public void AddEventListener(string eventName, Action<JsonElement?> eventHandler)
```
pavelfeldman, line 71: Implementing add/removeEventListener is typically a tedious task with several gotchas and edge conditions (like removing self while dispatching, removing dispatched while dispatching, removing non-dispatched while dispatching, etc.). I wonder if we can stay away from this and offer a different API:

```csharp
session.event("Animation.animationCreated") += (_) => Console.WriteLine("Animation created!");
```

This reads and implements simpler to me.

pavelfeldman (commented):
Looks mostly good.

src/Playwright/Core/CDPSession.cs
hutchcodes: Yes, += combines and then in the remove -= removes a handler from the combined handlers.

src/Playwright/Core/CDPSession.cs
hutchcodes: I definitely like that syntax better and explored that for a bit but couldn't find a way to make it work. 

The problem is that the `+=` creates a new object that combines the original handler with the new handler and tries to assign that back to the original. So `A += B` is equivalent to `A = A + B`, but we can't assign to the return value of the method.

The only idea I had was to create a custom indexed class that would allow us to do `session. Event["Animation.animationCreated"]' and automatically generate the event handler with a noop if it didn't exist. We could then `+=` to combine with that noop.

I wasn't too keen to start creating what amounts to a custom dictionary implementation and I was worried that it might be confusing to consumers of the interface to have something that looks kind of like a dictionary but has an odd behavior of being both empty and every `session. Event[""] returns a value. I'm open to it though if everyone else thinks that's the way to go.

hutchcodes: Found a way to make this work by returning an object that implements a standard C# event allowing consumer to add event listeners in a standard C# way
`client.Event("Network.requestWillBeSent").OnEvent += myEventHandler;`

PR #2465, opened 2023-01-26
chore: Configure awaits in CDPSession

hutchcodes:
Lack of ConfigureAwait is causing an error that prevents generating a new package

### DIFF
--- a/src/Playwright/Core/CDPSession.cs
+++ b/src/Playwright/Core/CDPSession.cs
@@ -48,8 +48,7 @@ internal class CDPSession : ChannelOwnerBase, ICDPSession, IChannelOwner<CDPSess
 
     IChannel<CDPSession> IChannelOwner<CDPSession>.Channel => _channel;
 
-    public Task DetachAsync()
-        => _channel.DetachAsync();
+    public Task DetachAsync() => _channel.DetachAsync();
 
     public Task<JsonElement?> SendAsync(string method, Dictionary<string, object>? args = null)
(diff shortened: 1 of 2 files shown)


src/Playwright/Core/CDPSession.cs
```diff
@@ -50,12 +50,12 @@ public CDPSession(IChannelOwner parent, string guid) : base(parent, guid)
 
     public async Task DetachAsync()
     {
-        await _channel.DetachAsync();
+        await _channel.DetachAsync().ConfigureAwait(false);
```
campersau, line 53: Since the method just forwards, it doesn't need to use `async` / `await`:

```c#
    public Task DetachAsync()
    {
        return _channel.DetachAsync();
    }
```

src/Playwright/Core/CDPSession.cs
```diff
@@ -50,12 +50,12 @@ public CDPSession(IChannelOwner parent, string guid) : base(parent, guid)
...
     }
 
     public async Task<JsonElement?> SendAsync(string method, Dictionary<string, object>? args = null)
     {
-        return await _channel.SendAsync(method, args);
+        return await _channel.SendAsync(method, args).ConfigureAwait(false);
```
campersau, line 58: Since the method just forwards, it doesn't need to use `async` / `await`:

```c#
    public Task<JsonElement?> SendAsync(string method, Dictionary<string, object>? args = null)
    {
        return _channel.SendAsync(method, args);
    }
```

PR #2489, opened 2023-02-22
fix: Start install with no new window

### LINKED ISSUE #2462: [BUG] Console window appearing over WPF Application v1.30
**Context:**
- Playwright Version: 1.30.0
- Operating System: Windows
- .NET version: [e.g. .NET 7]
- Browser: FireFox
- Extra: [any specific details about your environment]

**Code Snippet**

```cs
Microsoft.Playwright.Program.Main(new[] {"install", "firefox"});
```

**Describe the bug**

I'm currently using playwright in my WPF application and upon my application startup I have be running the code on startup:

Microsoft.Playwright.Program.Main(new[] {"install", "firefox"});

Everything has been working great. However, with the latest update playwright version (1.30.0), when that code is run it now pops up a console window cmd.exe when the other versions previously were silent. 

Even if the browser is already installed that window pops up for a few seconds over my application. I was wondering if there was a way to revert back to the old behavior especially for a WPF Application as I prefer not to have a console window popping up in front of my application.

I reported this issue before from a previous version https://github.com/microsoft/playwright-dotnet/issues/2103#issue-1206735789

nohwnd:
Running Playwright.Main(new [] { "install" }) shows console window in WPF app, which does not show in normal test. In normal test we are already running under existing conhost which is started as hidden so we don't show extra window. In WPF app that uses WinExe and has no existing conhost a new console is shown for playwright.cmd. This PR makes sure that the window is hidden.

Fixes #2462

### DIFF
--- a/src/Playwright/Program.cs
+++ b/src/Playwright/Program.cs
@@ -52,6 +52,7 @@ public class Program
         var playwrightStartInfo = new ProcessStartInfo(pwPath, allArgs)
         {
             UseShellExecute = false,
+            CreateNoWindow = true,
         };
         foreach (var pair in Driver.GetEnvironmentVariables())
         {


nohwnd:
Not sure how to write a test for this, without it being super heavyweight, you will have to believe me I guess: 

BEFORE: 

image


AFTER:

image

nohwnd:
@mxschmitt  please advise on the driver failure. I guess I can just push the version up to 1.31.0? But is that something you see often?

nohwnd:
Okay so apparently this would partially https://github.com/microsoft/playwright-dotnet/pull/2451, which in turn was breaking progress from installation (because it was not inteactive anymore), and would not correctly forward the output. 

WindowStyle Hidden also does not work, becuase it has no effect. 

So unless progress stops requiring interactive shell there is no solution.

Jozkee:
> WindowStyle Hidden also does not work, becuase it has no effect.

FWIW we added support for `WindowStyle` with `UseShellExecute = false` on https://github.com/dotnet/runtime/pull/82662.

PR #2500, opened 2023-03-01
chore: add support for osx-arm64

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2498

This increases the nuget size by 19%.

### DIFF
--- a/src/Playwright/Playwright.csproj
+++ b/src/Playwright/Playwright.csproj
@@ -40,11 +40,13 @@
     <None Include=".drivers\linux\LICENSE" Pack="true" PackagePath=".playwright\node" />
     <None Include=".drivers\linux\node" Link=".playwright\node\linux-x64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-x64" />
     <None Include=".drivers\linux-arm64\node" Link=".playwright\node\linux-arm64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-arm64" />
-    <None Include=".drivers\mac\node" Link=".playwright\node\mac\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\mac" />
+    <None Include=".drivers\mac\node" Link=".playwright\node\darwin-x64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\darwin-x64" />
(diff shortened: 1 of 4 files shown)


yury-s (approved):

src/Playwright/build/Microsoft.Playwright.targets
```diff
@@ -18,6 +19,13 @@
       <_PlaywrightCopyItems Include="$(MSBuildThisFileDirectory)..\.playwright\node\linux-arm64\**" Condition="$(PlaywrightPlatform.Contains('linux-arm64'))">
         <PlaywrightFolder>node\linux-arm64\</PlaywrightFolder>
       </_PlaywrightCopyItems>
+      <!-- 'osx' to stay backwards compatible -->
+      <_PlaywrightCopyItems Include="$(MSBuildThisFileDirectory)..\.playwright\node\mac-x64\**" Condition="$(PlaywrightPlatform.Contains('osx')) OR $(PlaywrightPlatform.Contains('osx-x64'))">
+        <PlaywrightFolder>node\mac-x64\</PlaywrightFolder>
```
campersau, line 24: The folder in the nuget package is actually named `darwin-x64` which was correctly done in the csproj but not in this file. See #2531

src/Playwright/Playwright.csproj
```diff
@@ -40,11 +40,13 @@
     <None Include=".drivers\linux\LICENSE" Pack="true" PackagePath=".playwright\node" />
     <None Include=".drivers\linux\node" Link=".playwright\node\linux-x64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-x64" />
     <None Include=".drivers\linux-arm64\node" Link=".playwright\node\linux-arm64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\linux-arm64" />
-    <None Include=".drivers\mac\node" Link=".playwright\node\mac\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\mac" />
+    <None Include=".drivers\mac\node" Link=".playwright\node\darwin-x64\node" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath=".playwright\node\darwin-x64" />
```
campersau, line 43: Ah I see. It is confusion to use different folders for `Include` and `PackagePath`.
`.drivers\mac\node` vs `.playwright\node\darwin-x64`
It would be easier to use the same folder e.g. either `mac` or `darwin-x64` for both of them.

mxschmitt: my idea was to rewrite our custom naming from Node.js drivers to the standardised .NET naming.

Issue #2486, opened 2023-02-18
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [x] https://github.com/microsoft/playwright/commit/ae05808f4b5eca5c35e15f6f51856c163d95838f (chore: align js routing logic with other langs (#21010))
  - [x] https://github.com/microsoft/playwright/commit/06fc72b6ed4424013485672194082cd3a1b7dada (revert(#12706): also fix related bugs it introduced (#21070))
  - [x] https://github.com/microsoft/playwright/commit/3f8f2a0fdd43d42cf420e7aefd7399cf54c9e4f0 (chore: do not use library stack capturer in test runner (#21075))
  - [x] https://github.com/microsoft/playwright/commit/55c95a4463be14a6bd39ce3b447ac93c863f5f3e (chore: do not send stacks as a part of the call metainfo (#21089))
  - [x] https://github.com/microsoft/playwright/commit/8ccc02a9d39478399e18b2045c9e6ffa3e6c50f3 (chore: use client wall time in traces (#21163))
  - [x] https://github.com/microsoft/playwright/commit/3fa19e80ada5c44a8285ef841f8d8b8ba298f22a (chore: wrap expect call in a zone (#21191))
  - [x] https://github.com/microsoft/playwright/commit/e17e0e40f858b2a2b22c8f2d0d45eb494a8c1258 (feat(routeFromHAR): add arguments (#21223))
  - [x] https://github.com/microsoft/playwright/commit/27027658dc208a5202944313a310f9571a51e5ac (chore: fix all Proxy() to account for symbol properties (#21272))
  - [x] https://github.com/microsoft/playwright/commit/09ff7eaaf2ab40fcaf0abb85f0308a91e1460a97 (chore: throw on context.close() if it was closed externally (#21347))
  - [x] https://github.com/microsoft/playwright/commit/0c5d46bb94b6ab122ba1f8354f5bb9492eb3de0b (feat: allow chaining locators with Locator.locator(anotherLocator) (#21391))
  - [x] https://github.com/microsoft/playwright/commit/86ca7e3949c3cd1a40cbf9e5e28d825355823869 (chore: introduce manual promise race (#21358))
  - [x] https://github.com/microsoft/playwright/commit/cffb6ac269ab97ac63ac98d88244603e3349b793 (chore: move sever/isomorphic to utils/ to use it in client (#21445))
[...]

PR #2511, opened 2023-03-15
chore: roll Playwright to 1.32.0-alpha-mar-15-2023

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2486

### DIFF
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -45,6 +45,7 @@ internal class Connection : IDisposable
     private readonly ConcurrentDictionary<int, ConnectionCallback> _callbacks = new();
     private readonly Root _rootObject;
     private readonly TaskQueue _queue = new();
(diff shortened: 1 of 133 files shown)


src/Playwright/Transport/Connection.cs
```diff
@@ -45,6 +45,7 @@ internal class Connection : IDisposable
     private readonly ConcurrentDictionary<int, ConnectionCallback> _callbacks = new();
     private readonly Root _rootObject;
     private readonly TaskQueue _queue = new();
+    private readonly HashSet<List<ClientSideCallMetadata>> _stackCollectors = new();
```
yury-s, line 48: Does it compare lists by their address or bu their content?

yury-s (approved):

src/Playwright/Transport/Connection.cs
mxschmitt: by their address.

Issue #2509, opened 2023-03-10
[BUG] Playwright excludes null when serializing request 


AlexSimionCommify:
**Context:**
- Playwright Version: 1.31.1
- Operating System: Windows
- .NET version: NET 6
- Browser: N/A
- Extra: [any specific details about your environment]

**Code Snippet**

Class example:
public class Recipients
    {
        public string? Id { get; set; }
        public string? Msisdn { get; set; }
    }
    
  Request example:
"recipients": [
  {
    "id": "someGuid",
    "msisdn": "+someMsisdn"
  },

**Describe the bug**
When I send the request based on a class like the one above "DataObject"
public async Task<IAPIResponse> PostOptouts<T>(T payload) where T : class
if I put "id=null", this apparently is not sent as null, and its possibly excluded from the request

but if I use the Json string directly ("recipients"), "id=null" seems to be passed successfully 
public async Task<IAPIResponse> PostOptoutsMessageWithJson(string payload)

So the description would be: "When trying to pass a class object as the request to be sent in Playwright, nulls are excluded from the request, but when passing jsonDirectly, nulls are passed ok 
_response = await _optoutsService.PostOptouts<PostOptoutRequest>(_optoutsRequest);

mxschmitt:
Could you provide us a full reproduction which we can run locally to demonstrate the issue? A small self contained script would be the best.

AlexSimionCommify:
sadly no, this is what I can provide as testdata (what I thought was relevant)
if you have anything specific, I can describe in mode details or other

mxschmitt:
The code above does not use any of Playwright's methods or does it internally use https://playwright.dev/dotnet/docs/api/class-route#route-fetch?

AlexSimionCommify:
It uses it internally:
 public async Task<IAPIResponse> PostOptouts<T>(T payload) where T : class
        {
            var optionsWithHeaderAndPayload = RequestHelper.Get_ApiRequestOptions_DefaultHeadersAndPayload(payload);
            var response = await _request.PostAsync(_optoutsUrl, optionsWithHeaderAndPayload);
            return response;
        }
 where:
  public static APIRequestContextOptions Get_ApiRequestOptions_DefaultHeadersAndPayload<T>(T payload) where T : class
        {
            APIRequestContextOptions apiRequestOptions_DefaultHeaders = new APIRequestContextOptions()
            {
                Headers = GetDefaultHeaders(),
                DataObject = payload
            };
[...]

PR #2523, opened 2023-03-24
fix(fetch): null values in body were not serialised

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2509


src/Playwright/Transport/Channels/APIRequestContextChannel.cs
```diff
@@ -62,13 +63,15 @@ public APIRequestContextChannel(string guid, Connection connection, APIRequestCo
...
             ["params"] = parameters?.ToProtocol(),
             ["headers"] = headers?.ToProtocol(),
-            ["jsonData"] = jsonData,
             ["postData"] = postData != null ? Convert.ToBase64String(postData) : null,
             ["formData"] = formData?.ToProtocol(),
             ["multipartData"] = multipartData?.ToProtocol(),
-        };
-
-        var response = await Connection.SendMessageToServerAsync(Guid, "fetch", message).ConfigureAwait(false);
+        }.FilterNullValues();
```
campersau, line 69: Isn't the null value filtering already done here?
https://github.com/microsoft/playwright-dotnet/blob/ffb7c21272c3f078c95976e7df60a41d9c53fbb7/src/Playwright/Transport/Connection.cs#L133-L138

campersau: Ah you removed it, should have scrolled to the end before commenting 😆

src/Playwright/Helpers/DictionaryExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        return dictionary.Where(x => x.Value != null).ToDictionary(x => x.Key, x => FilterNullValuesInternal(x.Value)) as Dictionary<string, T>;
+    }
+
+    private static object FilterNullValuesInternal(object value)
+    {
+        if (value is ChannelBase)
+        {
+            return new Dictionary<string, object>() { { "guid", ((ChannelBase)value).Guid } };
+        }
```
campersau, line 46:
```suggestion
        if (value is ChannelBase channelBase)
        {
            return new Dictionary<string, object>() { { "guid", channelBase.Guid } };
        }
```

src/Playwright/Helpers/DictionaryExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            return null;
+        }
+        if (value is string || value is int || value is bool || value is double || value is float || value is long || value is short || value is byte || value is uint || value is ulong || value is ushort || value is sbyte || value is decimal)
+        {
+            return value;
+        }
+        if (value is IDictionary dictionary)
+        {
+            return dictionary.Keys.Cast<string>().Where(key => dictionary[key] != null).ToDictionary(key => key, key => FilterNullValuesInternal(dictionary[key]));
```
campersau, line 63: If you add a little helper like

```c#
private static IEnumerable<DictionaryEntry> EnumerateDictionaryEntries(IDictionary dict)
{
    foreach (DictionaryEntry item in dict)
    {
        yield return item;
    }
}
```

You can change it to this and avoid multiple indexes into the same key:

```suggestion
            return EnumerateDictionaryEntries(dictionary).Where(entry => entry.Value != null).ToDictionary(entry => entry.Key, entry => FilterNullValuesInternal(entry.Value));
```

mxschmitt:
We'll go with https://github.com/microsoft/playwright-dotnet/pull/2532 instead.

PR #2531, opened 2023-03-28
chore: use darwin folder on osx and replace contains with identity check

### LINKED ISSUE #2525: [BUG] PlaywrightPlatform seems broken
Context:

Playwright Version: 1.31.1
Operating System: Windows 10

Setting PlaywrightPlatform to linux-arm64 gets linux-arm64,linux-x64 and win32_x64 node installations in the output dir.
Setting PlaywrightPlatform to linux-x64 gets linux-x64 and win32_x64 node installations in the output dir.

campersau:
The item `<_PlaywrightPlatforms Include="$(PlaywrightPlatform)" />` will split the `PlaywrightPlatform` property by `;` and so each entry can be checked individually with `%(_PlaywrightPlatforms.Identity)` which is better than using `Contains`.

Fixes #2525

@mxschmitt It looks like with version 1.32.0 the `mac` package was renamed to `darwin` in the nuget and the changes done in #2500 look wrong to me for the `Microsoft.Playwright.targets` file.
Maybe this needs a 1.32.1 release?

### DIFF
--- a/src/Playwright/build/Microsoft.Playwright.targets
+++ b/src/Playwright/build/Microsoft.Playwright.targets
@@ -12,25 +12,23 @@
       <PlaywrightPlatform Condition="'$(PlaywrightPlatform)' == '' AND $([MSBuild]::IsOSPlatform('OSX')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'Arm64'">osx-arm64</PlaywrightPlatform>
     </PropertyGroup>
     <ItemGroup Condition="'$(PlaywrightPlatform)' != 'none'">
+      <_PlaywrightPlatforms Include="$(PlaywrightPlatform)" />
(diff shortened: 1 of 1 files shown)


campersau:
Items have [well known](https://learn.microsoft.com/visualstudio/msbuild/msbuild-well-known-item-metadata) metadata including `Identity` which just returns the thing provided in the `Include`. ([Note that `Include` will first split the provided string by `;` and then creates an item for each of the parts.](https://learn.microsoft.com/visualstudio/msbuild/how-to-select-the-files-to-build#to-declare-multiple-items))

---

To avoid confusion should I also update the csproj to use the same folder as the nuget (e.g. `darwin-x64`) or should we do it the other way around (e.g. `mac`)? (The same applies for `linux` as well.)

mxschmitt (approved):
LGTM, thanks!

PR #2533, opened 2023-03-30
fix: all assets are kept in the playwright upstream - delete them here

maw136:
Delete `assets` subdirectory of Playwright.Tests.TestServer as all assets are kept updated by `build.sh` script


maw136:

maw136:

________________________________

PR #2540, opened 2023-04-05
test: enable tests that use screenshots as they work.

maw136:
Enable tests that use screenshots as they work.
Removing a warning.

Fixes https://github.com/microsoft/playwright-dotnet/issues/2541

### DIFF
--- a/src/Playwright/Helpers/TaskHelper.cs
+++ b/src/Playwright/Helpers/TaskHelper.cs
@@ -204,6 +204,6 @@ internal static class TaskHelper
 
     public static void IgnoreException(this Task task)
     {
-        _ = task.ContinueWith(t => t.Exception.Handle(_ => true), CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default);
+        _ = task.ContinueWith(t => t.Exception!.Handle(_ => true), CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Playwright/Helpers/TaskHelper.cs
```diff
@@ -204,6 +204,6 @@ private static async Task<bool> TimeoutTask(Task task, TimeSpan timeout)
 
     public static void IgnoreException(this Task task)
     {
-        _ = task.ContinueWith(t => t.Exception.Handle(_ => true), CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default);
+        _ = task.ContinueWith(t => t.Exception!.Handle(_ => true), CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default);
```
mxschmitt, line 207: whats the background for this? Since it didn't threw a warning before. Which .NET version are you using?

maw136: I'm using .NET 7.0, I will recheck it in a moment.

maw136: Ok it is actually the Resharper that is complaining at this line.
The null-forgiving operator stops them.
image

maw136: Should I remove the change?

maw136: Rolled back

PR #2545, opened 2023-04-07
chore(locator): Optimize Locator's constructor

RenderMichael:
JsonSerializerOptions is a deceptively heavy type, with [official guidance](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/configure-options?pivots=dotnet-7-0) recommending re-use, etc.

This patch only initializes a JsonSerializerOptions when it is necessary in Locator's constructor.

### DIFF
--- a/src/Playwright/Core/Locator.cs
+++ b/src/Playwright/Core/Locator.cs
@@ -46,12 +46,6 @@ internal class Locator : ILocator
     {
         _frame = parent;
         _selector = selector;
-
-        var serializerOptions = new JsonSerializerOptions
-        {
-            Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
-        };
-
         if (options?.HasTextRegex != null)
         {
             _selector += $" >> internal:has-text={EscapeForTextSelector(options.HasTextRegex, false)}";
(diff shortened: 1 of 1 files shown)


src/Playwright/Core/Locator.cs
```diff
@@ -40,18 +40,17 @@ internal class Locator : ILocator
 {
     internal readonly Frame _frame;
     internal readonly string _selector;
+    private static readonly JsonSerializerOptions _locatorSerializerOptions = new()
```
maw136, line 43: Should statics be at the top of class?

maw136:
Added tracking issue for this: https://github.com/microsoft/playwright-dotnet/issues/2546

src/Playwright/Core/Locator.cs
RenderMichael: I was following the style of the following `_testIdAttributeName`, but if that's the preferred code style I can move this field as well as `_testIdAttributeName` to the top

RenderMichael:
> Added tracking issue for this: #2546

Updated PR description to close that issue

mxschmitt (approved):

PR #2577, opened 2023-05-13
Switching RequestFinished event handling to use IReponse object 

maw136:
Switching RequestFinished event handling to use IReponse object that is already there - it is more useful in this context.


mxschmitt:
Closing as part of https://github.com/microsoft/playwright-dotnet/pull/2578 since we don't want breaking changes.

Issue #2551, opened 2023-04-14
[Feature] Add `RunAndWaitForDialogAsync`


campersau:
There are a bunch of helpers like `RunAndWaitForPopupAsync` / `RunAndWaitForFileChooserAsync` but `RunAndWaitForDialogAsync` is missing. It would be great to have that so you could write:

```c#
var dialog = await page.RunAndWaitForDialogAsync(async () =>
{
    await page.EvaluateAsync("alert('1')");
});
await dialog.DismissAsync();
```

An additional `WaitForDialogAsync` could also be added for symmetry.

campersau:
I think this pattern might only work for `alert` but not so much for `prompt` / `confirm`, closing.

PR #2580, opened 2023-05-18
chore: roll Playwright to 1.34.0-alpha-may-18-2023

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2565


src/Playwright/API/Generated/IPage.cs
```diff
@@ -2595,6 +2599,27 @@ public partial interface IPage
...
+    /// <summary>
+    /// <para>
+    /// Performs action and waits for a <see cref="IDialog"/> to be created by in the context.
+    /// If predicate is provided, it passes <see cref="IDialog"/> value into the <c>predicate</c>
+    /// function and waits for <c>predicate(message)</c> to return a truthy value.
+    /// </para>
+    /// </summary>
+    /// <param name="action">Action that triggers the event.</param>
+    /// <param name="options">Call options</param>
+    Task<IDialog> RunAndWaitForDialogAsync(Func<Task> action, PageRunAndWaitForDialogOptions? options = default);
```
campersau, line 2621: #2551 thanks 👍

src/Playwright/Core/Waiter.cs
```diff
@@ -73,15 +73,20 @@ public void Dispose()
...
             }
 
+            var info = new Dictionary<string, object>
+            {
+                ["waitId"] = _waitId,
+                ["phase"] = "after",
+            };
+            if (!string.IsNullOrEmpty(_error))
+            {
+                info["error"] = _error;
```
mxschmitt, line 83: note: error was not allowed to have null, hence we need to change it. This was not caused by the roll afaik, just saw it in the protocol logs.

dgozman (approved):

Issue #2581, opened 2023-05-18
[BUG] Regex flags (RegexOptions) ignored when passed to ILocatorAssertions.ToHaveAttributeAsync()


cjohnson:
**Context:**
- Playwright Version: 1.33.0
- Operating System: Windows
- .NET version: .NET 6
- Browser: All

**Code Snippet**
```cs
public async Task ShouldSupportToHaveAttributeByRegexAsync()
{
    await Page.SetContentAsync("<img src=\"https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?queryString=true\"/>");
    await Expect(Page.Locator("img")).ToHaveAttributeAsync("src", new Regex("https://playwright.dev/media/photo", RegexOptions.IgnoreCase));
}
```

**Stack Trace**
```
Failed ShouldSupportToHaveAttributeByRegexAsync [7 s]
  Error Message:
   Microsoft.Playwright.PlaywrightException : Locator expected to have attribute 'src' matching regex 'https://playwright.dev/media/photo'
But was: 'https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?queryString=true'
Call log:
LocatorAssertions.ToHaveAttributeAsync with timeout 5000ms
waiting for Locator("img")
  locator resolved to <img src="https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?quer./>
  unexpected value "https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?queryString=true"
  locator resolved to <img src="https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?quer./>
  unexpected value "https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?queryString=true"
  locator resolved to <img src="https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?quer./>
  unexpected value "https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?queryString=true"
  locator resolved to <img src="https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?quer./>
  unexpected value "https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?queryString=true"
  locator resolved to <img src="https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?quer./>
  unexpected value "https://PLAYWRIGHT.dEV/mEDIa/photo.JPG?queryString=true"
[...]

PR #2583, opened 2023-05-20
Set  ExpectedTextValue.IgnoreCase base on Regex

feiyun0112:
#2581


mxschmitt:
Hi! This is one way of fixing this issue. I thought about instead doing it like we do upstream, don't send `null` values over the wire and change `bool` to `bool?` in the [generated classes](https://github.com/microsoft/playwright/blob/main/utils/generate_dotnet_channels.js#L65).

feiyun0112:
> Hi! This is one way of fixing this issue. I thought about instead doing it like we do upstream, don't send `null` values over the wire and change `bool` to `bool?` in the [generated classes](https://github.com/microsoft/playwright/blob/main/utils/generate_dotnet_channels.js#L65).

Can you explain in detail how I do this

Issue #2590, opened 2023-05-29
chore: small refinements in prep. for nuget update


maw136:
chore: small refinements in prep. for nuget update,
- code style,
- capture variable
- unify event handling

maw136:
Solved by https://github.com/microsoft/playwright-dotnet/pull/2591

PR #2591, opened 2023-05-29
chore: small refinements, Capture Task to fix future compile error

maw136:
Fixes: https://github.com/microsoft/playwright-dotnet/issues/2590

### DIFF
--- a/src/Playwright/Transport/StdIOTransport.cs
+++ b/src/Playwright/Transport/StdIOTransport.cs
@@ -36,6 +36,7 @@ internal class StdIOTransport : IDisposable
     private const int DefaultBufferSize = 1024;  // Byte buffer size
     private readonly Process _process;
     private readonly CancellationTokenSource _readerCancellationSource = new();
+    private readonly Task _getResponseTask;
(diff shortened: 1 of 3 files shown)


src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -54,7 +55,7 @@ internal StdIOTransport()
         };
         _process.BeginErrorReadLine();
 
-        ScheduleTransportTask(GetResponseAsync, _readerCancellationSource.Token);
+        _getResponseTask = ScheduleTransportTaskAsync(GetResponseAsync, _readerCancellationSource.Token);
```
mxschmitt, line 58: Can we add a test for this, was it actually fixing something? How did you find out about this?

maw136: After upgrade to newer analyzers there are existing issues detected, like this:
https://github.com/microsoft/playwright-dotnet/actions/runs/5114152936/jobs/9194105803?pr=2556#step:8:11

Issue #2592, opened 2023-05-29
chore: Rewrite WritableStreamImpl to WritableStreamWrapper and StreamImpl to StreamWrapper


maw136:
chore: Rewrite WritableStreamImpl to WritableStreamWrapper and StreamImpl to StreamWrapper keeping best practices and notes for implementers in mind:https://learn.microsoft.com/en-us/dotnet/api/system.io.stream?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.IO.Stream)%3Bk(DevLang-csharp)%26rd%3Dtrue&view=net-6.0

maw136:
Solved by https://github.com/microsoft/playwright-dotnet/pull/2593

mxschmitt:
Closing as per https://github.com/microsoft/playwright-dotnet/pull/2593#issuecomment-1609921595.

maw136:
This detects issues with current code:
https://github.com/microsoft/playwright-dotnet/actions/runs/5414490141/jobs/9841624728?pr=2556

PR #2593, opened 2023-05-29
chore: Rewrite WritableStreamImpl to WritableStreamWrapper and StreamImpl to StreamWrapper

maw136:
Solves: https://github.com/microsoft/playwright-dotnet/issues/2592

### DIFF
--- a/src/Playwright/Core/Artifact.cs
+++ b/src/Playwright/Core/Artifact.cs
@@ -72,7 +72,7 @@ internal class Artifact : ChannelOwnerBase, IChannelOwner<Artifact>
         {
             using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
             {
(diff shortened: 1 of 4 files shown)


mxschmitt:
Thanks for contributing to this project. Did you mean instead of _rewrite_ to _rename_ this class? I don't see any logical changes inside this PR, does it fix something? If you have a test which you could add, it would help a ton to understand the problem you are fixing in the first place.

maw136:
Yes, there should be no logical changes apart from the patterns and practices of deriving from `Stream` class.
I will add some tests, I assumed all the tests that exist test underneath this class, nevertheless it is a good practice to add specific tests.

maw136:
We actually encounter analysis errors:
https://github.com/microsoft/playwright-dotnet/actions/runs/5414490141/jobs/9841624728?pr=2556
with newer analyzers.
What's more, I think striving for best possible quality of code is important, even without any detected bugs. It's not like there is a product owner and budget, it's an open-source project accepting contribution.

PR #2594, opened 2023-05-30
chore: small refinements string constants extraction and use more readable String.Equals implementation

maw136:
chore: string constants extraction and use more readable String.Equals implementation)
- extracted some strings to constants on top of classes and methods
- use more readable Equals with regard to comparing constant strings with input

fixes https://github.com/microsoft/playwright-dotnet/issues/2595


mxschmitt:
We appreciate the PR but would like to keep it how it was to align it with the other language bindings.

Thanks for your understanding!

Gso-1:
Yo lo podría asesorar pero tenemos que llegar a un acuerdo comercial

PR #2597, opened 2023-05-30
feature: add cancellation token to variety of ...WaitFor... APIs

maw136:
fixes: https://github.com/microsoft/playwright/issues/24083
adds CancellationToken where feasible and extends Waiter class to provide CancellationToken as a source to cancel wait

### DIFF
--- a/src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
+++ b/src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
@@ -24,6 +24,7 @@
 
 using System;
 using System.Text.Json.Serialization;
+using System.Threading;
 
 #nullable enable
 
@@ -42,6 +43,7 @@ public class BrowserContextRunAndWaitForConsoleMessageOptions
 
         Predicate = clone.Predicate;
         Timeout = clone.Timeout;
+        CancellationToken = clone.CancellationToken;
     }
 
     /// <summary>
@@ -62,6 +64,14 @@ public class BrowserContextRunAndWaitForConsoleMessageOptions
     /// </summary>
     [JsonPropertyName("timeout")]
     public float? Timeout { get; set; }
+
+    /// <summary>
+    /// <para>
+    /// Cancellation token to provide cancellation of waiting for console message
+    /// </para>
+    /// </summary>
+    [JsonPropertyName("cancellationToken")]
+    public CancellationToken CancellationToken { get; set; }
 }
 
 #nullable disable
--- a/src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
+++ b/src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
@@ -24,7 +24,6 @@
 
 using System;
 using System.Text.Json.Serialization;
-using System.Threading;
 
 #nullable enable
 
@@ -43,7 +42,6 @@ public class BrowserContextRunAndWaitForConsoleMessageOptions
 
         Predicate = clone.Predicate;
         Timeout = clone.Timeout;
-        CancellationToken = clone.CancellationToken;
     }
 
     /// <summary>
@@ -64,14 +62,6 @@ public class BrowserContextRunAndWaitForConsoleMessageOptions
     /// </summary>
     [JsonPropertyName("timeout")]
     public float? Timeout { get; set; }
-
-    /// <summary>
-    /// <para>
-    /// Cancellation token to provide cancellation of waiting for console message
-    /// </para>
-    /// </summary>
-    [JsonPropertyName("cancellationToken")]
-    public CancellationToken CancellationToken { get; set; }
 }
 
 #nullable disable
--- a/src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
+++ b/src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
@@ -24,6 +24,7 @@
 
 using System;
 using System.Text.Json.Serialization;
+using System.Threading;
 
 #nullable enable
 
@@ -40,9 +41,14 @@ public class BrowserContextRunAndWaitForConsoleMessageOptions
             return;
         }
 
+        CancellationToken = clone.CancellationToken;
         Predicate = clone.Predicate;
         Timeout = clone.Timeout;
     }
+
+    /// <summary><para>Cancellation token to provide cancellation of waiting for console message</para></summary>
+    [JsonPropertyName("cancellationToken")]
+    public CancellationToken CancellationToken { get; set; } = default!;
 
     /// <summary>
     /// <para>
--- a/src/Playwright/Core/Waiter.cs
+++ b/src/Playwright/Core/Waiter.cs
@@ -119,7 +119,7 @@ internal class Waiter : IDisposable
 
     internal void RejectOnCancellation(CancellationToken cancellationToken)
     {
-        var cancelledTask = Task.FromCanceled(cancellationToken);
+        var cancelledTask = Task.Delay(-1, cancellationToken);
         RejectOn(cancelledTask, () => cancelledTask.Dispose());
     }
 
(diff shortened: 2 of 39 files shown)


src/Playwright/Core/Waiter.cs
```diff
@@ -117,6 +117,12 @@ internal void RejectImmediately(Exception exception)
         _immediateError = exception;
     }
 
+    internal void RejectOnCancellation(CancellationToken cancellationToken)
+    {
+        var cancelledTask = Task.FromCanceled(cancellationToken);
+        RejectOn(cancelledTask, () => cancelledTask.Dispose());
```
campersau, line 123: This should probably look something like this:

```suggestion
        if (!cancellationToken.CanBeCanceled) return;
        var cts = new TaskCompletionSource<bool>();
        var registration = cancellationToken.Register(state => cts.SetCanceled((CancellationToken)state), cancellationToken);
        RejectOn(cts.Task, () => registration.Dispose());
```

maw136: Thank you very much. I was fighting with this part. I found some approaches but nothing worked properly.
I am adding tests to cover cancellation scenarios.

src/Playwright/Core/Waiter.cs
```diff
@@ -218,6 +230,13 @@ internal async Task<T> WaitForPromiseAsync<T>(Task<T> task, Action dispose = nul
             Dispose();
             throw new TimeoutException(ex.Message + FormatLogRecording(_logs), ex);
         }
+        catch (TaskCanceledException ex)
+        {
+            dispose?.Invoke();
+            _error = ex.ToString();
+            Dispose();
+            return default;
```
campersau, line 238: To me it would be strange to just get the default value back instead of a canceled exception.
In .NET it is standard that the user needs to handle canceled exceptions as soon as they are passing cancellation tokens around.

src/Playwright/Core/Waiter.cs
```diff
@@ -218,6 +230,13 @@ internal async Task<T> WaitForPromiseAsync<T>(Task<T> task, Action dispose = nul
             Dispose();
             throw new TimeoutException(ex.Message + FormatLogRecording(_logs), ex);
         }
+        catch (TaskCanceledException ex)
```
campersau, line 233: Catch `OperationCanceledException` instead or both.
See https://github.com/dotnet/runtime/issues/83406#issue-1623962011

src/Playwright/Core/Waiter.cs
```diff
@@ -117,6 +117,18 @@ internal void RejectImmediately(Exception exception)
...
 
+    internal void RejectOnCancellation(CancellationToken cancellationToken)
+    {
+        if (!cancellationToken.CanBeCanceled)
+        {
+            return;
+        }
+
+        var cts = new TaskCompletionSource<bool>();
+        var registration = cancellationToken.Register(cts.SetCanceled, false);
```
campersau, line 128: If you let the cancellation throw instead (https://github.com/microsoft/playwright-dotnet/pull/2597#discussion_r1211064250) this should use the `Set Canceled(cancellationToken)` instead so users can inspect it if the catch it higher up in the stack.

src/Playwright/Core/Waiter.cs
```diff
@@ -117,6 +117,18 @@ internal void RejectImmediately(Exception exception)
         _immediateError = exception;
     }
 
+    internal void RejectOnCancellation(CancellationToken cancellationToken)
+    {
+        if (!cancellationToken.CanBeCanceled)
+        {
+            return;
+        }
+
```
campersau, line 126: Just an idea: maybe it would be good to call `cancellationToken.ThrowIfCancellationRequested()` (here or maybe higher up in the call stack) so that if it is already cancelled we don't do anything.

src/Playwright/Core/Waiter.cs
maw136: I was looking for an accepted general rule for that. Do you suggest having that specific exception bubble up or converting it (as by default) to PlaywrightException?

campersau: From my experience cancellation exceptions escpecially ones triggered by a user provided CancellationToken aren't wrapped / converted in other exceptions. So I would vote for letting the cancellation exception bubble up.

maw136: I've refactored into rethrow.

src/Playwright/Core/Waiter.cs
maw136: Yeah, missed that inheritance hierarchy. Fixed

src/Playwright/Core/Waiter.cs
maw136: Yes, that was my understanding of your comment as well, alas I have added the throw in the wrong place, fixed now :)

src/Playwright/Core/Waiter.cs
```diff
@@ -243,7 +267,11 @@ private static async Task WrapActionAsync(Func<Task> action, CancellationTokenSo
         }
         catch
         {
-            cts.Cancel();
+            if (!cts.IsCancellationRequested)
```
mxschmitt, line 270: Why this change? What breaks if we always cancel here?

src/Playwright/Core/Waiter.cs
```diff
@@ -218,6 +235,13 @@ internal async Task<T> WaitForPromiseAsync<T>(Task<T> task, Action dispose = nul
             Dispose();
             throw new TimeoutException(ex.Message + FormatLogRecording(_logs), ex);
         }
+        catch (OperationCanceledException ex)
+        {
+            dispose?.Invoke();
+            _error = ex.ToString();
+            Dispose();
+            throw;
```
mxschmitt, line 243: would be nice to throw the same exception but with formatted logs like in line 250.

src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
```diff
@@ -62,6 +64,14 @@ public BrowserContextRunAndWaitForConsoleMessageOptions(BrowserContextRunAndWait
...
     [JsonPropertyName("timeout")]
     public float? Timeout { get; set; }
+
+    /// <summary>
+    /// <para>
+    /// Cancellation token to provide cancellation of waiting for console message
+    /// </para>
+    /// </summary>
+    [JsonPropertyName("cancellationToken")]
+    public CancellationToken CancellationToken { get; set; }
```
mxschmitt, line 74: Did you add these manually? All these files in that folder are generated automatically. We should add csharp specific properties upstream [like here](https://github.com/search?q=repo%3Amicrosoft%2Fplaywright+RunAndWaitFor+language%3AMarkdown&type=code) so that the generator will pick it up.

maw136: Yes, manual addition. I will move it to the source for the generator.

mxschmitt: See [here](https://github.com/microsoft/playwright-dotnet/blob/main/CONTRIBUTING.md#generating-the-apis--rolling-the-driver-upstream) for some notes on how it works.

src/Playwright/Core/Waiter.cs
maw136: There is no need to cancel 'harder'. I have replicated the solution from the other place in this PR where this check is made, and I remembered that I have added this cancellation here some time ago. This is the correct solution for the overall problem. Any code should be written in the most proper/correct way. Simply working is not ever good enough.

src/Playwright/Core/Waiter.cs
maw136: With `OperationCanceledException` the suggested flow is to throw the original exception to have better understanding of the place of cancelation. https://devblogs.microsoft.com/premier-developer/recommended-patterns-for-cancellationtoken/

maw136:
Added PR for upstream:
https://github.com/microsoft/playwright/pull/23953

src/Playwright/API/Generated/Options/BrowserContextRunAndWaitForConsoleMessageOptions.cs
maw136: https://github.com/microsoft/playwright/pull/23953

maw136:
Used this generator for generated members: 
https://github.com/microsoft/playwright/pull/23953

mxschmitt:
Closing as per https://github.com/microsoft/playwright/issues/24083#issuecomment-1640761745.

Issue #2598, opened 2023-06-01
[BUG] BrowserLaunchOptions Arguments no longer work in 1.31.0


TeamMug:
**Context:**
- Playwright Version: [1.32.0 +]
- Operating System: [Windows]
- .NET version: [.NET 6]
- Browser: [All, Chromium, Firefox, WebKit]
- Extra: [Microsoft login provider, NUnit 3.13.3]

**Code Snippet**

```xml
<Playwright>
    <BrowserName>chromium</BrowserName>
    <ExpectTimeout>10000</ExpectTimeout>
    <LaunchOptions>
      <!--<Devtools>true</Devtools>-->
      <Headless>false</Headless>
      <SlowMo>1000</SlowMo>
      <Args>['--auth-server-allowlist=\"_\"']</Args>
      <!--<Channel>msedge</Channel>-->
    </LaunchOptions>
  </Playwright>
```

**Describe the bug**

When developing tests we use playwright for dotnet on a windows machine, started from within Visual Studio (2022)
 We use the Microsoft login functionality as a login provider.
 The way this works on Windows with Chrome, is that the username is populated with the username that is logged in to the current windows session.
 For example, When joel@local.machine is logged in, in windows, the Microsoft login provider presents a popup with joel@local.machine as the username and only asks for a password. 
 There is a way around this, by starting Chrome with a '--auth-server-allowlist=\"_\"' flag.
 The way to configure this in playwright version 1.31.0 is by using the <Args> option in our test.runsettings file. 
 
As of Playwright version 1.32 and higher the <Args> no longer functions.
Tests just wont start. If we comment out or remove the <Args> tag, tests run as expected. (but then we have the login issue because the Microsoft authentication pre populates the username with the login in Windows user)

mxschmitt:
Sounds like a regression which got introduced in https://github.com/microsoft/playwright-dotnet/pull/2499.

PR #2602, opened 2023-06-04
chore: fix parsing of LaunchOptions.Args

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2598


src/Playwright.TestingHarnessTest/tests/mstest/basic.spec.ts
```diff
@@ -289,6 +289,42 @@ test('should be able to to parse LaunchOptions.Proxy from runsettings', async ({
   httpServer.close();
 });
 
+test('should be able to to parse LaunchOptions.Args from runsettings', async ({ runTest }) => {
```
yury-s, line 292:
```suggestion
test('should be able to parse LaunchOptions.Args from runsettings', async ({ runTest }) => {
```

yury-s (approved):

PR #2620, opened 2023-06-18
Prevent TieredPGO inlining from breaking stack trace walking logic

### LINKED ISSUE #2617: [BUG] Stack inspection done by playwright is fragile and breaks with Dynamic PGO enabled
**Context:**

We enabled DynamicPGO in .NET 8 Preview 5 and are seeing reports of some failures in applications that use playwright.

See for instance https://github.com/dotnet/runtime/issues/87628

This should readily repro if you run release mode builds with .NET 8 Preview 5 and default runtime options.

**Describe the bug**

https://github.com/microsoft/playwright-dotnet/blob/ca7b02ac910e573666f48d502bfabb1c17639e68/src/Playwright/Transport/Connection.cs#L477-L502

is looking for a specific pattern of method names in the stack trace. This is fragile unless all the methods being looked for are marked with `[MethodImpl(MethodImplOptions.NoInlining)]`. Note that with PGO the JIT may now inline methods at virtual, interface, and delegate call sites.

I suspect the fix may be as simple as marking this method noinline

https://github.com/microsoft/playwright-dotnet/blob/ca7b02ac910e573666f48d502bfabb1c17639e68/src/Playwright/Transport/Connection.cs#L521-L528

but there may be more cases elsewhere in this code that need similar treatment.

slang25:
Fixes #2617 

I've recreated the issue locally on macOS on my M2 MBP, this fixes the exception, I need to check that the traces have the right API names captured still.

### DIFF
--- a/src/Playwright.Tests/ConventionTests.cs
+++ b/src/Playwright.Tests/ConventionTests.cs
@@ -1,0 +1,32 @@
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.Playwright.Tests;
+
+public class ConventionTests
+{
+    // To ensure that public method are not inlined by the nwe PGO guided tiered JIT,
+    // we need to mark them with [MethodImpl(MethodImplOptions.NoInlining)]
+    [Test]
+    public void EnsurePublicMethodsAreNotInlined()
+    {
+        var assembly = typeof(Playwright).Assembly;
+
+        var types = assembly.GetTypes()
+            .Where(t => t.IsClass && !t.IsAbstract)
+            .Where(t => t.BaseType != null && t.BaseType.Name == "ChannelOwnerBase");
+
+        foreach (var type in types)
+        {
+            var methods = type
+                .GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly)
+                .Where(m => !m.IsSpecialName)
+                .Where(m => m.Name != "Dispose" && m.Name != "DisposeAsync" && m.Name != "ToString");
+
+            foreach (var method in methods)
+            {
+                Assert.True(method.MethodImplementationFlags.HasFlag(MethodImplAttributes.NoInlining), $"{type.Name}.{method.Name} is not marked with [MethodImpl(MethodImplOptions.NoInlining)]");
+            }
+        }
+    }
+}
--- a/src/Playwright.Tests/ConventionTests.cs
+++ b/src/Playwright.Tests/ConventionTests.cs
@@ -30,7 +30,9 @@ public class ConventionTests
             {
                 if (!method.MethodImplementationFlags.HasFlag(MethodImplAttributes.NoInlining))
                 {
-                    failedAssertions.AppendLine(CultureInfo.InvariantCulture, $"{type.Name}.{method.Name} is not marked with [MethodImpl(MethodImplOptions.NoInlining)]");
+#pragma warning disable CA1305
+                    failedAssertions.AppendLine($"{type.Name}.{method.Name} is not marked with [MethodImpl(MethodImplOptions.NoInlining)]");
+#pragma warning restore CA1305
                 }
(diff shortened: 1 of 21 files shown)


src/Playwright/Transport/Connection.cs
```diff
@@ -536,6 +538,7 @@ private static bool IsPlaywrightInternalNamespace(string namespaceName)
             namespaceName.StartsWith("Microsoft.Playwright.Helpers", StringComparison.InvariantCultureIgnoreCase));
     }
 
+    [MethodImpl(MethodImplOptions.NoInlining)] // This method is also a stacktrace marker.
```
slang25, line 541: While not strictly necessary, it seems a good idea to make this explicit

slang25:
Unfortunately I've done some further testing and it looks like a lot of the method names captured in the trace will change to base types, which isn't desired, I'll look to improve this PR.

slang25:
Update:

So this PR fixes the `NullReferenceException`, but in order to get the right API name captured like this:
https://github.com/microsoft/playwright-dotnet/blob/ca7b02ac910e573666f48d502bfabb1c17639e68/src/Playwright.Tests/TracingTests.cs#L69-L75
we're going to need to sprinkle `NoInlining` over all of the public APIs. I've been thinking about other approaches that don't do stack frame walking, however some public APIs flow through other public APIs , for example `Page.GotoAsync → Frame.GotoAsync` ([see here](https://github.com/microsoft/playwright-dotnet/blob/ca7b02ac910e573666f48d502bfabb1c17639e68/src/Playwright/Core/Page.cs#L292-L293)), which makes that harder to do in a non-breaking way.

slang25:
So this is now working and tested with some traces. Unfortunately the cleanest way I could think to do this was add `[MethodImpl(MethodImplOptions.NoInlining)]` absolutely everywhere on the public APIs of anything that inherits `ChannelOwnerBase`, this is enforced with a convention test.

src/Playwright.Tests/ConventionTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        foreach (var type in types)
+        {
+            var methods = type
+                .GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly)
+                .Where(m => !m.IsSpecialName)
+                .Where(m => m.Name != "Dispose" && m.Name != "DisposeAsync" && m.Name != "ToString");
+
+            foreach (var method in methods)
+            {
+                Assert.True(method.MethodImplementationFlags.HasFlag(MethodImplAttributes.NoInlining), $"{type.Name}.{method.Name} is not marked with [MethodImpl(MethodImplOptions.NoInlining)]");
```
martincostello, line 28: If you put these messages into a list and then assert the list is empty, you could fail the test with everything that needs to be fixed at once in a single message that concatenates them.

slang25: Sure, tell me now after I reran it 50 times!

src/Playwright.Tests/ConventionTests.cs
```diff
@@ -0,0 +1,47 @@
+using System.Globalization;
+using System.Reflection;
+using System.Text;
+
+namespace Microsoft.Playwright.Tests;
+
+public class ConventionTests
+{
+    // To ensure that public method are not inlined by the new tiered PGO JIT mode,
+    // we need to mark them with [MethodImpl(MethodImplOptions.NoInlining)]
```
mxschmitt, line 10:
```suggestion
    // we need to mark them with [MethodImpl(MethodImplOptions.NoInlining)]
    // See https://github.com/microsoft/playwright-dotnet/issues/2617
```

mxschmitt:
@slang25 thank you so much for looking into this and @AndyAyersMS for filing it inside this repo with all the details!

slang25:
My pleasure @mxschmitt 🙂

PR #2625, opened 2023-06-27
fix: allow RegexOptions.Compiled in GetInlineFlags

### LINKED ISSUE #2615: [Question]: Should we support `RegexOptions.Compiled`?
### Your question

I use `context.RouteAsync(RouteFilterRegex, route => {...})` API, and I create `RouteFilterRegex` with option `RegexOptions.Compiled` in order to have better performance. But looks like, the following code https://github.com/microsoft/playwright-dotnet/blob/ca7b02ac910e573666f48d502bfabb1c17639e68/src/Playwright/Helpers/RegexOptionsExtensions.cs#L35-L55 refuse to accept compiled regex.
To me, I would expect that we can set a compiled regex in `RouteAsync`, could you show me why you do not support that, or maybe fix it?

And this is actually a **regression**, at least I could do this in 1.24.0.

jjshao-ms:
Fix #2615


mxschmitt (approved):
Awesome work, appreciate the PR!

PR #2632, opened 2023-07-10
fix: SetInputFilesAsync() memory leak

### LINKED ISSUE #2629: [BUG] SetInputFilesAsync() allocates extremely large amount of memory
**Context:**
- Playwright Version: 1.35
- Operating System: Windows
- .NET version: .NET 6 and .NET 7
- Browser: Chromium, Firefox, Webkit
- Extra: Also occurs with Selenium Grid integration (using environment variable)

**Code Snippet**

```cs
[TestFixture]
public class PlaywrightPOCTest
{
    private static IPlaywright playwright;
    private static IBrowser browser;
    private static IBrowserContext context;
    private static IPage page;

    [OneTimeSetUp]
    public static async Task DownloadBrowsers()
    {
        var exitCode = Microsoft.Playwright.Program.Main(new[] { "install" });
        if (exitCode != 0)
        {
            throw new Exception($"Playwright exited with code {exitCode}");
        }

        playwright = await Playwright.CreateAsync();
    }

    [OneTimeSetUp]
    public static async Task EdgeDriverInitialize()
    {
        browser = await playwright.Chromium.LaunchAsync(
            new BrowserTypeLaunchOptions 
            { Channel = "msedge", Headless = true, /*SlowMo = 2000 */}
            );

        context = await browser.NewContextAsync(); 

        page = await context.NewPageAsync();
    }

    [Test]
    public static async Task PlaywrightFileUpload()
    {
        var frame = page.MainFrame;

        await frame.GotoAsync(@"http://www.fileconvoy.com/");
        Console.WriteLine($"Memory usage before upload: {System.Diagnostics.Process.GetCurrentProcess().PrivateMemorySize64}");
        await frame.Locator("id=upfile_0").SetInputFilesAsync(@"C:\Users\testuser\Downloads\19588.pdf");
        Console.WriteLine($"Memory usage after upload: {System.Diagnostics.Process.GetCurrentProcess().PrivateMemorySize64}");
        await frame.Locator("id=readTermsOfUse").CheckAsync();
        await frame.GetByRole(AriaRole.Button).GetByText("Upload").ClickAsync();
        Console.WriteLine($"Memory usage after page navigation: {System.Diagnostics.Process.GetCurrentProcess().PrivateMemorySize64}");
    }

    [OneTimeTearDown]
    public static async Task EdgeDriverCleanup()
    {
        await context?.CloseAsync();
        await browser?.CloseAsync();
        playwright.Dispose();
    }
}
```

**Describe the bug**

Steps

-Download this ~30MB PDF from here: https://research.nhm.org/pdfs/19588/19588.pdf
-Update path in SetInputFilesAsync() method to your file location
-Run PlaywrightFileUpload() test

Expected Results:

I would expect memory usage to increase temporarily by the uploaded filesize (~30MB) and then go away or be cleaned up fairly soon after page navigation occurs.

Actual Results:

Memory usage jumps from about 40MB to >700MB just from the file upload and persists for quite a long time in other tests which do not involve file uploads. Sometimes I can get it to drop by forcing a GC using GC.Collect() several minutes later, but I definitely should not be doing that.

This has had an impact on long running scenarios where we are doing a file upload as the first dependency and then running several tests on a web app after the upload. It also affects concurrent runs in our framework due to out of memory issues. Fwiw, I took a memory snapshot inside VS, and it shows up as a large number of dead char[] objects which are outside of my code.

kelmelzer:
fixes #2629 by streaming any file upload(s) totalling over 5MB in size, skipping base64/byte[] conversion to string buffer

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,13 +109,13 @@ The resulting code will follow our style guides. This is also enforced in our CI
 Tests can either be executed in their entirety:
 
 ```bash
-dotnet test ./src/Playwright.Tests/Playwright.Tests.csproj --logger:"console;verbosity=detailed"
+dotnet test -f net6.0 ./src/Playwright.Tests/Playwright.Tests.csproj --logger:"console;verbosity=detailed"
 ```
 
 You can also specify a single test to run:
 
 ```bash
-dotnet test ./src/Playwright.Tests/Playwright.Tests.csproj --logger:"console;verbosity=detailed" --filter Playwright.Tests.TapTests
+dotnet test -f net6.0 ./src/Playwright.Tests/Playwright.Tests.csproj --logger:"console;verbosity=detailed" --filter Playwright.Tests.TapTests
 ```
 
 Additionally, you can use the Test Explorer if you're using Visual Studio.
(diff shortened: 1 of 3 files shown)


mxschmitt:
Hey! Sorry for the late review. This looks more like a workaround rather than fixing the actual issue when streaming the file over to the driver side. I hope to be able to investigate into it the next few days.

PR #2637, opened 2023-07-14
chore: use CultureInfo.InvariantCulture for BigInts

mxschmitt:
This fixes https://github.com/microsoft/playwright-browsers/actions/runs/5554767815/jobs/10144990687

PR #2242, opened 2022-08-09
chore: add runsettings option to set the expect-timeout

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2237

PR #1917, opened 2022-01-07
chore: bump dependencies

mxschmitt:
(all generated automatically by the Visual Studio Dependency Upgrade feature)

Extracted from #1884.

Failing bots are unrelated.

PR #2639, opened 2023-07-14
chore: enable Microsoft.CodeAnalysis.NetAnalyzers

campersau:
When this project changed from multitargeting to only targeting netstandard, it also lost all `Microsoft.CodeAnalysis.NetAnalyzers`.
See https://learn.microsoft.com/visualstudio/code-quality/install-net-analyzers?view=vs-2022

Ref: #2637

### DIFF
--- a/src/Common/Dependencies.props
+++ b/src/Common/Dependencies.props
@@ -18,6 +18,10 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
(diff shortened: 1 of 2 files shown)


mxschmitt:
Do we have to specify a Remove here? https://github.com/microsoft/playwright-dotnet/blob/db21766d06ad7ea93d47b9a6d02ce03f27ff9396/src/Playwright/Playwright.csproj#L57

campersau:
I guess we can remove the removes :)
Looks like in #1917 you accidentially updated the versions via visual studio which created duplicate entries in the csproj vs Dependencies.props and I guess visual studio added the removes for you.
In #2242 you removed these duplicate package references and also removed some of the removes but not all.

mxschmitt (approved):

PR #2648, opened 2023-07-26
fix: hanging CreateAsync() under .NET framework

### LINKED ISSUE #2517: [BUG] Playwright.CreateAsync() is hanging when it's executed by GitHub runner
**Context:**
- Playwright Version: 1.31.1
- Operating System: Windows
- .NET version: .NET Framework 4.7.2
- Browser: Not applicable because it's hanging before any browser interaction
- Extra: 

Please find repro: [https://github.com/op-13/pwtest](https://github.com/op-13/pwtest)

**Code Snippet**

The simple test solution is using VSTest and a simple GitHub actions workflow. The single file has only a dummy test:

```cs
    [TestClass]
    public class SimpleTestWithPlaywright
    {
        [TestMethod]
        public async Task TestMethod1()
        {
            using (var instance = await Playwright.CreateAsync())  // it is hanging in the CreateAsync() call
            {
                Assert.AreEqual(1, 1); // we will never reach this point when the GitHub runner executes this
            }
        }
    }
```

The relevant part of the yaml workflow:

```yml
    - name: Run tests
      shell: powershell
      run: |
        . "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe" (Get-ChildItem $currentDirectory -include *.dll -recurse | where {$_.name -match "SimpleTest.dll" -and $_.Directory -notMatch '.*\\obj\\.*'}).FullName
```

**Describe the bug**

The bug is that the code is hanging when it's being executed by the GitHub runner: [link to a sample workflow run](https://github.com/op-13/pwtest/actions/runs/4487804137/jobs/7891632342)
I have to cancel this manually otherwise it would run for hours (until it times out).

If I run the same command from the machine of the runner, it works fine:

image

It doesn't make any difference if I run the GitHub runner in an Administrator command prompt or not.
Tested on 5 different machines, including installations of Windows Server 2019 Standard and a Windows 10 Pro.

Also we tried to use Microsoft's [VSTest-action](https://github.com/microsoft/vstest-action), it was still hanging.

mxschmitt:
This "backports" the [following upstream](https://github.com/dotnet/corefx/pull/25990/files#diff-b26f2f6aa938959376f7e73072595ac23ad5e3b4d24b9be07a36efa452102d28) change by setting the value during Process.Start() which before internally called the Kernel to get the CodePage.

Fixes #2517

### DIFF
--- a/src/Playwright/Transport/StdIOTransport.cs
+++ b/src/Playwright/Transport/StdIOTransport.cs
@@ -25,6 +25,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Playwright.Helpers;
@@ -44,7 +45,7 @@ internal class StdIOTransport : IDisposable
     {
         _process = GetProcess();
         _process.StartInfo.Arguments = "run-driver";
-        _process.Start();
+        StartProcessWithUTF8IOEncoding(_process);
         _process.Exited += (_, _) => Close("Process exited");
         _process.ErrorDataReceived += (_, error) =>
         {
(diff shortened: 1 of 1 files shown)


src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -130,6 +131,34 @@ private static Process GetProcess()
         };
     }
 
+    /// <summary>
+    /// Starts the given process with UTF8-encoding for input and output without
+    /// BOM which breaks driver transport otherwise.
+    /// This function backports https://github.com/dotnet/runtime/issues/22051 to .NET Framework.
+    /// See also https://github.com/microsoft/playwright-dotnet/issues/2517.
```
dgozman, line 138: Let's also link to SO post for the record.

src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -130,6 +131,34 @@ private static Process GetProcess()
...
+    /// </summary>
+    /// <param name="process">The process to start.</param>
+    private static void StartProcessWithUTF8IOEncoding(Process process)
+    {
+        var encoding = new UTF8Encoding(false);
```
dgozman, line 143: Let's add a comment that we can switch to the new field once available.

dgozman (approved):

PR #2660, opened 2023-08-14
chore(roll): roll to Playwright 1.37.0

mxschmitt:


src/tools/Playwright.Tooling/DriverDownloader.cs
```diff
@@ -113,6 +113,7 @@ private async Task DownloadDriverAsync(DirectoryInfo destinationDirectory, strin
         }
 
         using var client = new HttpClient();
+        client.Timeout = TimeSpan.FromMinutes(5);
```
mxschmitt, line 116: note: this does not affect end-users, this gets only used before publishing and for local development.

aslushnikov (approved):

Issue #2658, opened 2023-08-11
[BUG] - Video.DeleteAsync hangs if page is never interacted with


thomhurst:
### System info
- Playwright Version: v1.36.0
- Operating System: Windows 11
- Browser: Chromium
- Other info:

### Context
We have some logic within a TestBase class to spawn up the browser/page/etc. for every test.

However some tests are created from TestSourceData, so are created dynamically, and some will check some condition, and decide they don't need to run, and then exit.

Then our test teardown (also in the base class) will run afterwards, and try to call DeleteAsync on any video file.

So when viewing the test, you'll see a browser pop up (because of the base setup), the test decides it doesn't need to run, and then it'll exit. 

However, it seems that if the page wasn't interacted with, that the call to Video.DeleteAsync hangs. It also seems to hang if I call Video.PathAsync too.

Reproduceable code below:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Test file (self-contained)**

```cs
using Microsoft.Playwright;

var pw = await Playwright.CreateAsync();

var chrome = await pw.Chromium.LaunchAsync(new BrowserTypeLaunchOptions
{
    Headless = false
});

var recordVideoDir = $"C:\\temp\\PWVideos\\{DateTime.Now.ToString("HHmmss")}";

Directory.CreateDirectory(recordVideoDir);

var context = await chrome.NewContextAsync(new BrowserNewContextOptions
{
    RecordVideoDir = recordVideoDir,
});

var page = await context.NewPageAsync();

await page.CloseAsync();

await page.Video!.DeleteAsync(); // or await page.Video.PathAsync();

pw.Dispose();
```

**Steps**
- [Run the test]
- [...]

**Expected**

Any video file should be deleted

**Actual**

Execution hangs indefinitely on the DeleteAsync / PathAsync call

PR #2664, opened 2023-08-15
Fix Video Hanging

thomhurst:
#2658 

I think the issue is the video is only newed up within a getter. If this happens after a page close, then the Close event handlers are never hit and therefore the TaskCompletionSource never has a result set


thomhurst:
@mxschmitt looks like the tests are flaky - Unchanged tests are intermittently failing?

Issue #2662, opened 2023-08-14
[BUG] Exception When Making A Request With A Multi-Part Form File


Xen0byte:
### System Info
- Playwright Version: 1.36.0
- Operating System: Windows 11
- Browser: N/A
- Other Info: The code is fairly specific, but the easy repro is to just fire a POST multi-part form request with a file against (I suspect) any endpoint that supports that. In fact, the endpoint probably doesn't even matter, because this exception occurs before even making the request.

### Source code

```cs
  using SixLabors.ImageSharp;

  public async Task GenerateImageAsync(string format, int width, int height)
  {
      Image<Rgba32> image = new(width, height);

      byte GetRandomByte() => Convert.ToByte(Random.Shared.Next(Convert.ToInt32(byte.MaxValue).IncrementBy(1)));

      for (int x = 0; x < width; x++)
          for (int y = 0; y < height; y++)
              image[x, y] = new Rgba32(GetRandomByte(), GetRandomByte(), GetRandomByte(), GetRandomByte());

      string name = $"IMAGE.TEMP.{format.ToUpper()}";

      Models.API.Media.Image data = new()
      {
          Name = name,
          Format = format.ToUpper(),
          ParentID = Guid.Parse(ScenarioContext.Get<string>("Media Folder GUID")),
          Properties = new[] { new ImageProperties() { Alias = "description", Value = $"Programmatically Generated @ {DateTime.UtcNow.ToString("O")}" } }
      };

      switch (format.ToUpper())
      {
          case "GIF":  await image.SaveAsGifAsync(name);  break;
          case "JPEG": await image.SaveAsJpegAsync(name); break;
          case "PNG":  await image.SaveAsPngAsync(name);  break;

          default: Assert.Fail($@"Unsupported Image Format ""{format}""; Supported Formats Are: GIF, JPEG, PNG"); break;
      };

      IFormData form = HTTPCallContext.CreateFormData();

      form.Set("file", new FilePayload() { Buffer = await File.ReadAllBytesAsync(name), Name = name, MimeType = $"image/{format.ToLower()}" }); // TODO: Just Use A MemoryStream And Don't Create The File Locally
      form.Set("data", JsonSerializer.Serialize(data));

      // HTTPCallContext is an IAPIRequestContext
      IAPIResponse response = await HTTPCallContext.PostAsync(ImagesEndpoint, new APIRequestContextOptions() { Form = form });

      File.Delete(name);

      if (response.Status is 200)
[...]

PR #2671, opened 2023-08-17
fix: throw when serializing FilePayload in Form Fetch request

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2662

This might affect Java as well.

### DIFF
--- a/src/Playwright.Tests/BrowserContextFetchTests.cs
+++ b/src/Playwright.Tests/BrowserContextFetchTests.cs
@@ -703,6 +703,21 @@ public class BrowserContextFetchTests : PageTestEx
         Assert.AreEqual(200, response.Status);
     }
 
+    [PlaywrightTest("browsercontext-fetch.spec.ts", "should not allow file payloads in Form parameter")]
+    public async Task ShouldNotAllowFilePayloadsInFormParameter()
+    {
+        var file = new FilePayload()
+        {
+            Name = "f.js",
+            MimeType = "text/javascript",
+            Buffer = Array.Empty<byte>()
+        };
+        var formData = Context.APIRequest.CreateFormData();
+        formData.Set("file", file);
+        var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => Context.APIRequest.PostAsync(Server.EmptyPage, new() { Form = formData }));
+        StringAssert.Contains("Form requests don't support file payloads, use MultiPart=formData instead.", exception.Message);
+    }
+
     [PlaywrightTest("browsercontext-fetch.spec.ts", "should serialize data to json regardless of content-type")]
     public async Task ShouldSerializeDatatoJsonRegardlessOfContentType()
     {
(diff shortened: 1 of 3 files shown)


src/Playwright/Core/FormData.cs
```diff
@@ -51,13 +51,17 @@ private IFormData SetImpl(string name, object value)
...
+    internal IList<object> ToProtocol(bool throwWhenSerializingFilePayloads = false)
     {
         var output = new List<object>();
         foreach (var kvp in Values)
         {
             if (kvp.Value is FilePayload file)
             {
+                if (throwWhenSerializingFilePayloads)
+                {
+                    throw new PlaywrightException("Form requests don't support file payloads, use MultiPart=formData instead.");
```
campersau, line 63: Should it be `Multipart` instead?

```suggestion
                    throw new PlaywrightException("Form requests don't support file payloads, use Multipart=formData instead.");
```

https://github.com/microsoft/playwright-dotnet/blob/657f415d3eaa05df2d3dbff0a56526aad285d6f1/src/Playwright/API/Generated/Options/APIRequestContextOptions.cs#L166-L167

src/Playwright.Tests/BrowserContextFetchTests.cs
```diff
@@ -703,6 +703,21 @@ public async Task ShouldSupportMultipartFormData()
...
+        var file = new FilePayload()
+        {
+            Name = "f.js",
+            MimeType = "text/javascript",
+            Buffer = Array.Empty<byte>()
+        };
+        var formData = Context.APIRequest.CreateFormData();
+        formData.Set("file", file);
+        var exception = await PlaywrightAssert.ThrowsAsync<PlaywrightException>(() => Context.APIRequest.PostAsync(Server.EmptyPage, new() { Form = formData }));
+        StringAssert.Contains("Form requests don't support file payloads, use MultiPart=formData instead.", exception.Message);
```
campersau, line 718:
```suggestion
        StringAssert.Contains("Form requests don't support file payloads, use Multipart=formData instead.", exception.Message);
```

src/Playwright/Core/FormData.cs
mxschmitt: Good catch!

src/Playwright/Core/FormData.cs
```diff
@@ -51,13 +51,17 @@ private IFormData SetImpl(string name, object value)
...
+    internal IList<object> ToProtocol(bool throwWhenSerializingFilePayloads = false)
     {
         var output = new List<object>();
         foreach (var kvp in Values)
         {
             if (kvp.Value is FilePayload file)
             {
+                if (throwWhenSerializingFilePayloads)
+                {
+                    throw new PlaywrightException("Form requests don't support file payloads, use Multipart=formData instead.");
```
yury-s, line 63: Doesn't it serialize just file name in case of url encoded params? In java we just send it as a json object which should make it clear on the receiving end that something was wrong with the encoding.

yury-s (approved):

src/Playwright/Core/FormData.cs
mxschmitt: Java serializes to JSON:

https://github.com/microsoft/playwright-java/blob/4204096c2495addb4acd373531a234dc8c62f6ba/playwright/src/main/java/com/microsoft/playwright/impl/Serialization.java#L363

in JS we are more strict:
```
  form?: { [key: string]: string|number|boolean; };
  multipart?: { [key: string]: string|number|boolean|fs.ReadStream|FilePayload; };
so we simply not allow it.
```

see here:

https://github.com/microsoft/playwright/blob/adc9b2d39025af39c741bd0deef9f884d6473dae/packages/playwright-core/src/client/fetch.ts#L348-L349

In Python we probably throw protocol wise: https://github.com/microsoft/playwright-python/blob/42c0bf19d7ae415552172d7c04cdb7afd9dad7fb/playwright/_impl/_fetch.py#L343

PR #2673, opened 2023-08-18
fix: video.PathAsync() should not hang after direct launch/close

mxschmitt:
The issue was that the artifact came after the page close event has fired. So it was waiting for a page.close event inside the Video.cs but it already arrived, so it stalled.

Fixes https://github.com/microsoft/playwright-dotnet/issues/2658
Closes https://github.com/microsoft/playwright-dotnet/pull/2664

### DIFF
--- a/src/Playwright/Core/Video.cs
+++ b/src/Playwright/Core/Video.cs
@@ -35,12 +35,12 @@ internal class Video : IVideo
     public Video(Page page, Connection connection)
     {
         _isRemote = connection.IsRemote;
+        page.Close += (_, _) => _artifactTcs.TrySetResult(null);
+        page.Crash += (_, _) => _artifactTcs.TrySetResult(null);
         if (page.IsClosed)
         {
             _artifactTcs.TrySetResult(null);
         }
-        page.Close += (_, _) => _artifactTcs.TrySetResult(null);
-        page.Crash += (_, _) => _artifactTcs.TrySetResult(null);
     }
 
     public async Task DeleteAsync()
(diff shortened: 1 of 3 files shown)


yury-s (approved):

src/Playwright/Core/Video.cs
```diff
@@ -35,14 +35,21 @@ internal class Video : IVideo
...
     {
         _isRemote = connection.IsRemote;
-        page.Close += (_, _) => _artifactTcs.TrySetCanceled();
-        page.Crash += (_, _) => _artifactTcs.TrySetCanceled();
+        if (page.IsClosed)
+        {
+            _artifactTcs.TrySetResult(null);
+        }
+        page.Close += (_, _) => _artifactTcs.TrySetResult(null);
+        page.Crash += (_, _) => _artifactTcs.TrySetResult(null);
```
campersau, line 43: There is still a potential race condition when the page is closed right after the `page.IsClosed` check and before the event handlers are added, so lets do the check after adding the event handlers.

```suggestion
        page.Close += (_, _) => _artifactTcs.TrySetResult(null);
        page.Crash += (_, _) => _artifactTcs.TrySetResult(null);
        if (page.IsClosed)
        {
            _artifactTcs.TrySetResult(null);
        }
```

campersau: Note that `IsClosed` is not set when the page crashed, so using `page.ClosedOrCrashedTcs` as in the other PR might be more correct as this would handle closed and crashed correctly.

mxschmitt: good catch, thanks!

PR #2674, opened 2023-08-20
fix: adjust way of detecting if its launched with attached Console

mxschmitt:
https://github.com/microsoft/playwright-dotnet/commit/657f415d3eaa05df2d3dbff0a56526aad285d6f1#commitcomment-124959681

https://stackoverflow.com/questions/70997254/c-sharp-console-application-is-there-a-way-to-detect-whether-the-exe-was-run-fr

### DIFF
--- a/src/Playwright/Transport/StdIOTransport.cs
+++ b/src/Playwright/Transport/StdIOTransport.cs
@@ -146,15 +146,7 @@ internal class StdIOTransport : IDisposable
         var originalInputEncoding = Console.InputEncoding;
         var originalOutputEncoding = Console.OutputEncoding;
 
-        var hasConsole = true;
-        try
-        {
-            var height = Console.WindowHeight;
-        }
-        catch
-        {
-            hasConsole = false;
-        }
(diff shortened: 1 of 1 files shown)


yury-s (approved):

kelmelzer:
@mxschmitt I think my suggestion was still always returning 0 for Console.CursorTop and Console.CursorLeft even if I was running as a console app, so we could still be messed up here. Maybe your original implementation was better using Console.WindowHeight?

I am getting wildly different results depending on build environment, too.

On my MacBook with VS in a sample MVC site loading the Playwright Dotnet library using .CreateAsync(), I get 0 for Console.CursorTop and Console.CursorLeft.

On my Windows dev machine running the same code, I get an IOException when I try to get those, so maybe you need to keep that try/catch in place when setting hasConsole?

mxschmitt:
The actual code only affects Windows, so thats why you probably don't experience any issues with macOS.

What do you think about this?

```csharp
        bool hasConsole = false;
        try
        {
            hasConsole = (Console.CursorLeft > 0) && (Console.CursorTop > 0);
        }
        catch
        {
        }
```

Would be awesome if you could test it!

Issue #2694, opened 2023-09-08
[Feature] Smaller docker image


lonix1:
## PROBLEM

The official [docker image](https://github.com/microsoft/playwright-dotnet/blob/main/utils/docker/Dockerfile.jammy) is a whopping 3GiB in size.

That is a real problem:
- We customise the image (e.g. update to net7.0) and add our own app in there. So we have dozens of extra "overlay2" layers each about 3GiB in size, which clogs up our server daily.
- Most operations are very slow, especially in CI. It takes time to copy / update / etc. such a large image and so every pipeline takes forever.

## SOLUTION

*Please consider releasing a leaner image.*

If possible then alpine; otherwise debian.

Also, it would be nice to have an image with only chromium (which would significantly reduce the size). That could be setup as a matrix docker build, so we could download per-browser images, or an all-in-one image. If not possible, then at least provide guidance how we can strip down the image to remove unnecessary browsers.

mxschmitt:
Supporting alpine is unfortunately outside of the scope for the project, since we don't compile browsers for this platform. Having a Docker image for Debian is tracked here: https://github.com/microsoft/playwright-dotnet/issues/2680

> Also, it would be nice to have an image with only chromium (which would significantly reduce the size)

I recommend filing a separate feature request for it, this sounds helpful.

> If not possible, then at least provide guidance how we can strip down the image to remove unnecessary browsers.

So far we recommend building your own image, which you can do either by downloading Playwright manually over nuget and invoking the cli to do: ["install --with-deps"](https://playwright.dev/dotnet/docs/browsers#install-system-dependencies) or you checkout our repository and use our publishing [scripts](https://github.com/microsoft/playwright-dotnet/blob/main/utils/docker/build.sh).

mxschmitt:
Thanks for the suggestion! This indeed makes it 570MB smaller for me: https://github.com/microsoft/playwright-dotnet/pull/2695

PR #2695, opened 2023-09-10
fix: clear nuget cache in Docker image before publishing

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2694#issuecomment-1712741094

This saves us 600 MB.


mxschmitt:
Sounds like doing both is the best:

```
root@e3386b9a4a0b:/# dotnet nuget locals all --list
http-cache: /root/.local/share/NuGet/http-cache
global-packages: /root/.nuget/packages/
temp: /tmp/NuGetScratchroot
plugins-cache: /root/.local/share/NuGet/plugin-cache
root@e3386b9a4a0b:/# 
```

then we delete the whole  /root/.nuget folder as well.

Issue #2408, opened 2022-11-23
[Feature] Support dotnet 7


JHooper-bsl:
In dotnet 7 I get the warning "The project bindings ... could not be discovered. Navigation, step completion and other features are disabled"

PR #2720, opened 2023-10-09
fix: use minimum 6.0.0 System.Text.Json version

mxschmitt:
This does it like the Azure CLI folks: https://github.com/Azure/azure-sdk-for-net/blob/ff337a1ce38c57f770d28a36198d90ad7be27f80/sdk/digitaltwins/Azure.DigitalTwins.Core/samples/DigitalTwinsClientSample/DigitalTwinsClientSample.csproj#L21

But this most likely will regress https://github.com/microsoft/playwright-dotnet/issues/2408 since it reverts https://github.com/microsoft/playwright-dotnet/pull/2410, requires testing.

Fixes https://github.com/microsoft/playwright-dotnet/issues/2696

### DIFF
--- a/src/Playwright/Playwright.csproj
+++ b/src/Playwright/Playwright.csproj
@@ -31,9 +31,9 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
-    <PackageReference Include="System.Text.Json" Version="7.0.3" />
+    <PackageReference Include="System.Text.Json" Version="6.0.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
   </ItemGroup>
   <ItemGroup>
     <None Remove=".drivers\**" />


mxschmitt:
@campersau seems like a bug in .NET Azure Functions to me in the way how they bundle their assemblies. (they don't publish a custom System.Text.Json package).

campersau:
The problem is that the AzureFunctionsTools which hosts the local functions does include the `System.Text.Json.dll` with version `6.0.xx` as you can see in `%LOCALAPPDATA%\AzureFunctionsTools\Releases\4.55.0\cli_x64`. I don't know if the same is true when the function gets actually deployed.

In .NET Framework you can work around that by using [`<bindingRedirect>`](https://learn.microsoft.com/dotnet/framework/configure-apps/file-schema/runtime/bindingredirect-element) and in .NET you can workaround that with this code which tries to load assemblies without doing a version check:
```c#
AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
private Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEventArgs args)
{
    var name = new AssemblyName(args.Name);
    var dll = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, name.Name + ".dll");
    if (File.Exists(dll))
    {
        return Assembly.LoadFrom(dll);
    }
    return null;
}
```
E.g. .NET will try to load `System.Text.Json = 7.0.0` and fails so the custom `CurrentDomain_AssemblyResolve` is called and will resolve `System.Text.Json = 6.0.0` from the AzureFunctionsTools directory and .NET will just use the returned assembly without additional checks and everything works*.

In my opinion libraries should always try to use the lowest dependency version possible since the users can then "override" it by specifying it explicitly in their project.
E.g. if Playwright requires `System.Text.Json >= 6.0.0` and I explicitly include a `PackageReference` to `System.Text.Json = 7.0.0` then Playwright will also use `7.0.0` without any hacks*.

\* This of course only works at runtime if there aren't any breaking changes which Playwirght uses.

I am not sure what https://github.com/microsoft/playwright-dotnet/issues/2408 was about exactly as there isn't enough information for me to reproduce it, maybe there are other solutions for this too.

mxschmitt:
Thanks for your great investigation! 💯 

> I don't know if the same is true when the function gets actually deployed.

From my experiments System.Text.Json was not included in the zip which got uploaded to the Azure Function.

> I am not sure what https://github.com/microsoft/playwright-dotnet/issues/2408 was about exactly as there isn't enough information for me to reproduce it, maybe there are other solutions for this too.

Yeah for me it also wasn't clear anymore, I made a quick .NET 7 test where it looked good.

I guess I'll merge this for now (which is like a revert of https://github.com/microsoft/playwright-dotnet/pull/2410 so it should be fine). Thanks again!

yury-s (approved):

Issue #2719, opened 2023-10-06
[BUG] Modification of Page.Frames


jgilbert2017:
### System info
- Playwright Version: v1.37.1
- Operating System: Linux
- Browser: Chromium
- Other info:

### Source code
**Test file (self-contained)**

```cs
var frames = page.Frames;
var frame = frames.Single(f => f.Url.Contains("framename", StringComparison.Ordinal));
```

**Steps**

**Expected**

No exception from the enumerator.

**Actual**

99% of the time there is no error and my code works fine, however I did observe the following exception which is indicative of unsafe access from multiple threads from within the playwright library:
```
Inner Exception: InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at System.Linq.Enumerable.TryGetSingle[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
   at MyCode.PerformGetRequestAsync(IPage page, String url, CancellationToken token) in C:\src\MyCode.cs:line 100
```

It looks like the issue is that `Page.Frames` returns a read only wrapper around `_frames` which is subject to modification via `Channel_FrameAttached` / `Channel_FrameDetached`.

I can probably limit the probability of this error by calling `page.Frames.ToList()` but the underlying issue should probably be addressed.

mxschmitt:
Can you share a minimal reproducible which surfaces this issue? Are you using Playwright with multiple threads?

jgilbert2017:
I am not using multiple threads.  As I described, I believe it is a race condition where the list returned is being modified by the internal thread of playwright.  I don't have a reliable reproduction but the error should be clear.

There needs to be synchronization around the internal usage of _frames and probably a copied list should be returned by `Frames`.

jgilbert2017:
this does not fix https://github.com/microsoft/playwright-dotnet/issues/2719. Channel_FrameAttached and Channel_FrameDetached require synchronization when altering _frames because they are called from a different thread than .Frames.

also, if a copy is being returned there is not much reason to use AsReadOnly()

jgilbert2017:
@mxschmitt can you please address the synchronization issue i have highlighted?  this issue was closed without a proper fix being put in place.  are you planning on fixing it as part of some other work item?

PR #2724, opened 2023-10-11
fix: return copy of list in IPage.Frames

mxschmitt:
We do that in the other places too:

image

Fixes https://github.com/microsoft/playwright-dotnet/issues/2719


jgilbert2017:
this does not fix #2719.  `Channel_FrameAttached` and `Channel_FrameDetached` require synchronization when altering `_frames` because there are called from a different thread than `.Frames`.

also, if a copy is being returned there is not much reason to use `AsReadOnly()`

yury-s (approved):

PR #2726, opened 2023-10-11
chore(roll): collect stale handles

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2690


src/Playwright/Transport/Channels/BrowserChannel.cs
```diff
@@ -103,7 +103,7 @@ internal override void OnMessage(string method, JsonElement? serverParams)
 
     internal Task<CDPChannel> NewBrowserCDPSessionAsync()
     => Connection.SendMessageToServerAsync<CDPChannel>(
```
yury-s, line 105: Would it make sense to add Channel.SendMessageToServerAsync and call it in all these places? It then can check if the channel has been disposed already.

yury-s (approved):

PR #2740, opened 2023-11-01
chore: Simplify the implementation of the driver environment variables

0xced:
* Since everything is static ("csharp" constant string + environment version + assembly version) the dictionary is only created once and thus changed from a method to a property.
* Using the built-in `Version.ToString(3)` instead of the equivalent custom method `GetSemVerPackageVersion`.


mxschmitt (approved):

mxschmitt:
Great cleanup!

PR #2741, opened 2023-11-02
chore: address .NET 8 linter warnings

mxschmitt:
https://learn.microsoft.com/en-us/aspnet/core/diagnostics/asp0016?view=aspnetcore-8.0
https://learn.microsoft.com/en-us/aspnet/core/diagnostics/asp0019?view=aspnetcore-8.0

### DIFF
--- a/src/Playwright.MSTest/ContextTest.cs
+++ b/src/Playwright.MSTest/ContextTest.cs
@@ -40,7 +40,8 @@ public class ContextTest : BrowserTest
 
     public virtual BrowserNewContextOptions ContextOptions()
     {
-        return new() {
+        return new()
+        {
             Locale = "en-US",
             ColorScheme = ColorScheme.Light,
         };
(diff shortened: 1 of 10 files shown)


yury-s (approved):

src/Playwright.MSTest/ContextTest.cs
```diff
@@ -40,6 +40,10 @@ public async Task ContextSetup()
 
     public virtual BrowserNewContextOptions ContextOptions()
     {
-        return null!;
+        return new()
+        {
+            Locale = "en-US",
```
yury-s, line 45: This seems irrelevant to .net 8

yury-s (approved):

src/Playwright.MSTest/ContextTest.cs
mxschmitt: Discussed offline: Align it with the upstream defaults.

PR #2742, opened 2023-11-02
chore: move to .NET 8 (LTS)

mxschmitt:
This gets merged around November 14-16, 2023, when .NET 8 is the new LTS.

Changes include:
- Do not target `netcoreapp3.1` anymore for the CLI, since its EOL, target .NET 6/ .NET 7/.NET 8 instead.
  - These are the currently supported .NET versions
  - dotnet tools only work under these two target frameworks
- Had to remove `GeneratePackageOnBuild` which made dotnet pack work, seems like not to have other bad side effects

### DIFF
--- a/src/Playwright.TestingHarnessTest/Playwright.TestingHarnessTest.csproj
+++ b/src/Playwright.TestingHarnessTest/Playwright.TestingHarnessTest.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <IsPackable>false</IsPackable>
     <ReleaseVersion>0.0.0</ReleaseVersion>
     <LangVersion>10</LangVersion>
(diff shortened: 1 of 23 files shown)


utils/docker/Dockerfile.focal
yury-s, line 1: According to https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle .NET 6 will be supported until Nov 2024, I believe we could keep this image.

src/Playwright.TestingHarnessTest/Playwright.TestingHarnessTest.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
yury-s, line 4: Will it break those who haven't updated to net8 yet?

mxschmitt: These are our internal tests, its not affecting users.

yury-s (approved):

PR #2744, opened 2023-11-03
fix(trace): remote SaveAsAsync() with fileName only

mxschmitt:


src/Playwright/Core/Artifact.cs
```diff
@@ -67,7 +67,7 @@ public async Task SaveAsAsync(string path)
             await _channel.SaveAsAsync(path).ConfigureAwait(false);
             return;
         }
-        System.IO.Directory.CreateDirectory(Path.GetDirectoryName(path));
+        System.IO.Directory.CreateDirectory(Path.GetDirectoryName(Path.GetFullPath(path)));
```
yury-s, line 70: Does System.IO.Directory.CreateDirectory not work with relative paths?

yury-s (approved):

src/Playwright/Core/Artifact.cs
mxschmitt: It does, but the case I was fixing it was just foo.zip

Issue #2748, opened 2023-11-07
[BUG] Cannot read incomplete UTF-16 JSON text as string with missing low surrogate


xyqlx:
### System info
- Playwright Version: v1.39.0
- Operating System: Windows 11
- Browser: Chromium
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Link to the GitHub repository with the repro**

[https://github.com/your_profile/playwright_issue_title]

or

**Test file (self-contained)**

```cs
using Microsoft.Playwright;

using var playwright = await Playwright.CreateAsync();
await using var browser = await playwright.Chromium.LaunchAsync();
var page = await browser.NewPageAsync();
// this is a page on a Chinese video website
await page.GotoAsync("https://www.bilibili.com/video/BV1yR4y1C7KX");

var collections = await page.QuerySelectorAllAsync(".video-page-card-small");
foreach (var item in collections)
{
    var titleA = await item.QuerySelectorAsync(".info a .title");
    if (titleA != null)
    {
        Console.WriteLine(await titleA.InnerTextAsync());
    }
}
```

**Steps**
- Create the project following <https://playwright.dev/dotnet/docs/library#usage> and dotnet run.

**Expected**

No exceptions or I can handle it.

**Actual**

```text
Unhandled exception. Microsoft.Playwright.PlaywrightException: System.InvalidOperationException: Cannot read incomplete UTF-16 JSON text as string with missing low surrogate.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_ReadIncompleteUTF16()
   at System.Text.Json.JsonReaderHelper.TryUnescape(ReadOnlySpan`1 source, Span`1 destination, Int32 idx, Int32& written)
   at System.Text.Json.JsonReaderHelper.GetUnescapedString(ReadOnlySpan`1 utf8Source)
   at System.Text.Json.JsonDocument.GetString(Int32 index, JsonTokenType expectedType)
   at System.Text.Json.JsonElement.ToString()
   at Microsoft.Playwright.Transport.Channels.ElementHandleChannel.OnMessage(String method, Nullable`1 serverParams) in /_/src/Playwright/Transport/Channels/ElementHandleChannel.cs:line 55
   at Microsoft.Playwright.Transport.Connection.Dispatch(PlaywrightServerMessage message) in /_/src/Playwright/Transport/Connection.cs:line 302
[...]

PR #2749, opened 2023-11-08
Throws exception when preview have wrong utf string

feiyun0112:
#2748 

 when preview have wrong utf string, syste.text.json cannot convert it to object.

However, the preview only use to print, should not affect the normal operation of the program

### DIFF
--- a/src/Playwright/Transport/Channels/ElementHandleChannel.cs
+++ b/src/Playwright/Transport/Channels/ElementHandleChannel.cs
@@ -26,6 +26,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Playwright.Core;
 using Microsoft.Playwright.Helpers;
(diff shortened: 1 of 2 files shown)


src/Playwright/Transport/Channels/ElementHandleChannel.cs
```diff
@@ -49,7 +50,25 @@ internal override void OnMessage(string method, JsonElement? serverParams)
...
             case "previewUpdated":
-                PreviewUpdated?.Invoke(this, serverParams.Value.GetProperty("preview").ToString());
+                var preview = string.Empty;
+                try
+                {
+                    preview = serverParams.Value.GetProperty("preview").ToString();
+                }
+                catch
+                {
+                    var matchResult = Regex.Match(serverParams.Value.ToString(), "\"preview\":\"(.*)\"");
```
kblok, line 60: I think there should be a better way to accomplish this. What's the exception you're getting?

mxschmitt: I went ahead here: https://github.com/microsoft/playwright/pull/28038

Investigation: we trimmed wrongly, it ended up in an invalid utf8 char and the .NET JSON parser didn't like this.

PR #2757, opened 2023-11-20
chore(roll): roll Playwright to 1.40.0-alpha-oct-18-2023

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2690

### DIFF
--- a/src/Playwright.Tests/TestConstants.cs
+++ b/src/Playwright.Tests/TestConstants.cs
@@ -45,5 +45,5 @@ internal static class TestConstants
     internal static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
     internal static readonly bool IsLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
 
-    internal static string TargetClosedErrorMessage = "Target page, context or browser has been closed";
+    internal static readonly string TargetClosedErrorMessage = "Target page, context or browser has been closed";
 }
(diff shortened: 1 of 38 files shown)


src/Playwright.Tests/TestConstants.cs
```diff
@@ -44,4 +44,6 @@ internal static class TestConstants
     internal static readonly bool IsMacOSX = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
     internal static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
     internal static readonly bool IsLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
+
+    internal static string TargetClosedErrorMessage = "Target page, context or browser has been closed";
```
campersau, line 48: Nit:

```suggestion
    internal static readonly string TargetClosedErrorMessage = "Target page, context or browser has been closed";
```

src/Playwright/Core/Waiter.cs
```diff
@@ -122,6 +122,15 @@ internal void RejectImmediately(Exception exception)
...
         PlaywrightException navigationException,
         Func<T, bool> predicate = null)
+    {
+        RejectOnEvent(eventSource, e, () => navigationException, predicate);
+    }
+
+    internal void RejectOnEvent<T>(
+        object eventSource,
+        string e,
+        Func<PlaywrightException> navigationExceptionPredicate,
```
yury-s, line 132: Why can we pass `navigationException` directly instead of a function? Also predicate sounds like a wrong term for this parameter.

yury-s (approved):

Issue #2784, opened 2023-12-04
[BUG] CMD.EXE UNC message when Playwright run from share


harryhobson:
### System info
- Playwright Version: [v1.40.0]
- Operating System: Windows 10, Windows Server 2016
- Browser: Chromium
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

```cs
using Microsoft.Playwright;
using var playwright = await Playwright.CreateAsync();
```

**Steps**
- Run the code as a Windows Console application from a network share (e.g. \\\\someshare\\test)
   - This can be done either via Powershell (using the working directory as the share)
   - Or by double clicking the executable

**Expected**

No output on the console window.

**Actual**

On the console screen the below text appears:

> '\\\\someshare\\test'
> CMD.EXE was started with the above path as the current directory.
> UNC paths are not supported.  Defaulting to Windows directory.

Two reasons I am raising this issue:
- I have not run into any functionality issues even though this message has been appearing. However, I am not sure what implication this has on Playwright functionality.
- This message pollutes console applications, if it is possible to adjust the CMD.EXE calls such that this does not appear on the console screen of the application that would be appreciated.

PR #2786, opened 2023-12-05
fix: get rid of driver wrapper

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2784

Background: Its not possible not run cmd.exe on a UNC share, so we need to get rid of it or switch to another loader.


yury-s (approved):

src/Playwright/Helpers/Driver.cs
```diff
@@ -63,46 +63,68 @@ internal static string GetExecutablePath()
...
         }
         else
         {
             throw new PlaywrightException("Unknown platform");
         }
 
-        return Path.Combine(driversPath, ".playwright", "node", platformId, runnerName);
+        string getArgs(string args)
+        {
+            var cliEntrypoint = Path.Combine(driversPath, ".playwright", "package", "lib/cli/cli.js");
```
campersau, line 122: Combine everything?
```suggestion
            var cliEntrypoint = Path.Combine(driversPath, ".playwright", "package", "lib", "cli", "cli.js");
```

Nit: calculate this once upfront outside of `getArgs`.

PR #2793, opened 2023-12-08
chore: move channel events into impl classes

mxschmitt:


src/Playwright/Transport/Connection.cs
```diff
@@ -305,8 +305,7 @@ internal void Dispatch(PlaywrightServerMessage message)
                 @object.DisposeOwner(message.Params.Value.TryGetProperty("reason", out var reason) ? reason.GetString() : null);
                 return;
             }
-
-            @object.Channel?.OnMessage(message.Method, message.Params);
+            ((ChannelOwnerBase)@object).OnMessage(message.Method, message.Params);
```
yury-s, line 308: Maybe we can change Objects.TryGetValue to return ChannelOwnerBase and avoid this cast?

src/Playwright/Core/CDPSession.cs
```diff
@@ -41,26 +41,29 @@ internal class CDPSession : ChannelOwnerBase, ICDPSession, IChannelOwner<CDPSess
...
-        _channel.CDPEvent += OnCDPEvent;
     }
 
     ChannelBase IChannelOwner.Channel => _channel;
 
     IChannel<CDPSession> IChannelOwner<CDPSession>.Channel => _channel;
 
+    internal override void OnMessage(string method, JsonElement? serverParams)
+    {
+        OnCDPEvent(serverParams!.Value.GetProperty("method").ToString(), serverParams.Value.GetProperty("params"));
```
yury-s, line 52: Check the `method` name, there may be more than one in the future.

yury-s (approved):

PR #2794, opened 2023-12-09
chore: move channels into impl classes

mxschmitt:


src/Playwright/Transport/ChannelOwnerBase.cs
```diff
@@ -136,4 +136,15 @@ private void UpdateEventSubscription(string eventName, bool enabled)
             }),
             true).IgnoreException();
     }
+
+    internal Task<JsonElement?> SendMessageToServerAsync(
```
yury-s, line 140: why do we need this specialization of the template?

yury-s (approved):

src/Playwright/Transport/ChannelOwnerBase.cs
mxschmitt: Some legacy, let me try to follow-up with that.

mxschmitt: Turns out we have it twice:

a) SendMessageToServerAsync without the generic gives you a JsonElement? back.
b) SendMessageToServerAsync with a generic which internally deserialises the JSON for you into some specific data structure T

So no need to follow-up!

maw136:
This broke implementation:
` { "chromiumSandbox", options.ChromiumSandbox },
                { "slowMo", options.ChromiumSandbox },`

PR #2796, opened 2023-12-10
chore: various channel follow-ups

mxschmitt:
~~Gets rebased on https://github.com/microsoft/playwright-dotnet/pull/2795 once landed.~~


src/Playwright/Core/ScriptsHelper.cs
```diff
@@ -87,3 +87,8 @@ internal static string EvaluationScript(string content, string path)
         throw new ArgumentException("Either path or content property must be present");
     }
 }
+
+internal class EvaluateArgumentGuidElement
```
yury-s, line 91: Doesn't C# style gives preference to file per class?

src/Playwright/Transport/Converters/EvaluateArgumentValueConverter.cs
```diff
@@ -252,11 +252,8 @@ private static object ToExpectedType(object parsed, Type t, IDictionary<object,
 
             foreach (var kv in parsedExpando)
             {
-                var property = t.GetProperties().FirstOrDefault(prop => string.Equals(prop.Name, kv.Key, StringComparison.OrdinalIgnoreCase));
-                if (property != null)
-                {
-                    property.SetValue(objResult, ToExpectedType(kv.Value, property.PropertyType, visited));
-                }
+                var property = Array.Find(t.GetProperties(), prop => string.Equals(prop.Name, kv.Key, StringComparison.OrdinalIgnoreCase));
+                property?.SetValue(objResult, ToExpectedType(kv.Value, property.PropertyType, visited));
```
yury-s, line 256: What is the minimal .Net version that supports  `?.` operator, are we confident that we'll not need to build with it?

yury-s (approved):

src/Playwright/Transport/Converters/EvaluateArgumentValueConverter.cs
mxschmitt: .NET takes care of that via the TFM, we can use any newer C# language feature how we like.

PR #2797, opened 2023-12-10
chore: do not expose PlaywrightEvent

mxschmitt:
Looks like this was exposed by accident.


src/Playwright/API/IEvent.cs
```diff
@@ -1,33 +0,0 @@
...
- * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
- * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
- */
-
-namespace Microsoft.Playwright;
-
-public interface IEvent
```
yury-s, line 27: I believe we cannot remove it as there maybe clients doing something like `switch (Event.Name) {` and without a base interface it will fall apart. Or is there a workaround for them?

Issue #2811, opened 2024-01-05
[BUG] Page.RouteFromHARAsync() does not embed information into .har file when setting UpdateContent = Embed


PandaMagnus:
### System info
- Playwright Version: 1.39 and 1.40
- Operating System: Windows 10
- Browser: Chromium, Chrome, Firefox, WebKit
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Test file (self-contained)**

```cs
using Microsoft.Playwright;
using Microsoft.VisualStudio.TestTools.UnitTesting;
using System.IO;
using System.Threading.Tasks;

namespace ReportValidation;

[TestClass]
public class HarTest
{
    [TestMethod]
    public async Task EmbedNotWorking()
    {
        using IPlaywright pw = await Playwright.CreateAsync();
        await using IBrowser browser = await pw.Chromium.LaunchAsync(new BrowserTypeLaunchOptions { Channel = "chrome" });

        await using IBrowserContext context = await browser.NewContextAsync();
        IPage page = await context.NewPageAsync();

        await page.RouteFromHARAsync(Path.Combine("C:\\", "temp", "test", "test.har"),
            new PageRouteFromHAROptions
            {
                Update = true,
                UpdateMode = HarMode.Minimal,
                UpdateContent = RouteFromHarUpdateContentPolicy.Embed
            });

        await page.GotoAsync("https://www.google.com");
        await page.Locator("css=textarea[aria-label='Search']").FillAsync("playwright.dev");
        await page.Keyboard.PressAsync("Enter");
        await page.Locator("css=a[href='https://playwright.dev/']").IsVisibleAsync();
    }
}

```

**Steps**
- Run the test
- Check C:\temp\test

**Expected**

Only the .har file exists with all of the resources embedded in-line

**Actual**

Separate JSON and CSS files, with the .har file referencing out to those files

I may be in a weird and unintended scenario here. I'm not using this for mocking. I'm using it to scrape just specific pages of content. Using BrowserContext.NewContextAsync(new{ RecordHarPath = ... }) works exactly as expected, but in this case I need to scrape just specific pages spread over multiple iterations of a test, so having one gigantic .har file will make things extremely difficult. I understand if this is not an expected use case, and if that's the answer I'm happy to turn this into a feature request!

PR #2815, opened 2024-01-10
fix(page): RouteFromHarAsync embed files

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2811

### DIFF
--- a/src/Playwright/Core/Page.cs
+++ b/src/Playwright/Core/Page.cs
@@ -1353,10 +1353,10 @@ internal class Page : ChannelOwner, IPage
         {
             await Context.RecordIntoHarAsync(har, this, new()
             {
-                NotFound = options?.NotFound,
-                Url = options?.Url,
-                UrlString = options?.UrlString,
-                UrlRegex = options?.UrlRegex,
+                NotFound = options.NotFound,
+                Url = options.Url,
+                UrlString = options.UrlString,
+                UrlRegex = options.UrlRegex,
                 Update = options.Update,
                 UpdateContent = options.UpdateContent,
                 UpdateMode = options.UpdateMode,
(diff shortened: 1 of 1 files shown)


src/Playwright/Core/Page.cs
```diff
@@ -1357,6 +1357,9 @@ await Context.RecordIntoHarAsync(har, this, new()
                 Url = options?.Url,
                 UrlString = options?.UrlString,
                 UrlRegex = options?.UrlRegex,
+                Update = options.Update,
+                UpdateContent = options.UpdateContent,
+                UpdateMode = options.UpdateMode,
```
campersau, line 1362: `options` can be `null`

```suggestion
                Update = options?.Update,
                UpdateContent = options?.UpdateContent,
                UpdateMode = options?.UpdateMode,
```
(not sure if that compiles)

src/Playwright/Core/Page.cs
```diff
@@ -1357,6 +1357,9 @@ await Context.RecordIntoHarAsync(har, this, new()
                 Url = options?.Url,
                 UrlString = options?.UrlString,
                 UrlRegex = options?.UrlRegex,
+                Update = options.Update,
+                UpdateContent = options.UpdateContent,
```
yury-s, line 1361:
```suggestion
                UpdateContent = options?.UpdateContent,
```
here and below?

yury-s (approved):

src/Playwright/Core/Page.cs
campersau: I fell for it as well but above is already a `options?.Update == true` check so `options` can't be `null` here.
Still might be good to make it consistent.

yury-s: Agreed. @mxschmitt let's make it consistent.

mxschmitt: good point! addressed.

PR #2821, opened 2024-01-16
chore: roll driver to 1.41.0-beta

yury-s:
Reference https://github.com/microsoft/playwright-dotnet/issues/2772


src/Playwright/Core/Page.cs
```diff
@@ -48,6 +48,8 @@ internal class Page : ChannelOwner, IPage
     private List<RouteHandler> _routes = new();
     private Video _video;
     private string _closeReason;
+    internal bool _closeWasCalled = false;
```
kblok, line 51: This could by an internal property instead of exposing the variable :)

src/Playwright/Core/Route.cs
```diff
@@ -42,8 +42,9 @@ namespace Microsoft.Playwright.Core;
 internal class Route : ChannelOwner, IRoute
 {
     private readonly RouteInitializer _initializer;
-    private readonly Request _request;
+    internal readonly Request _request;
```
kblok, line 45: Same here. I would expose this using a read-only property.

mxschmitt (approved):

src/Playwright/Core/BrowserContext.cs
```diff
@@ -50,8 +50,9 @@ internal class BrowserContext : ChannelOwner, IBrowserContext
     private List<RouteHandler> _routes = new();
     internal readonly List<Page> _pages = new();
     private readonly Browser _browser;
-    private bool _closeWasCalled;
+    internal bool _closeWasCalled;
     private string _closeReason;
+    private List<HarRouter> _harRouters = new();
```
campersau, line 55:
```suggestion
    private readonly List<HarRouter> _harRouters = new();
```

src/Playwright/Core/Page.cs
```diff
@@ -48,6 +48,8 @@ internal class Page : ChannelOwner, IPage
...
+    private List<HarRouter> _harRouters = new();
```
campersau, line 52:
```suggestion
    private readonly List<HarRouter> _harRouters = new();
```

src/Playwright/Core/Page.cs
yury-s: Done.

src/Playwright/Core/Route.cs
yury-s: Done.

Issue #2827, opened 2024-01-17
[BUG] MSB3073 error with Microsoft.Playwright 1.41.0 on Ubuntu


martincostello:
### System info

- Playwright Version: v1.41.0
- Operating System: Ubuntu
- Browser: N/A
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Link to the GitHub repository with the repro**

https://github.com/martincostello/dotnet-patch-automation-sample/commit/a41590758cadf4bc65b270cf05822d80bfadb0a5 in https://github.com/martincostello/dotnet-patch-automation-sample/pull/419

**Steps**

- Run `build.ps1`

**Expected**

Code compiles and tests pass.

**Actual**

[Build fails with error `MSB3073`](https://github.com/martincostello/dotnet-patch-automation-sample/actions/runs/7555540234/job/20571444376#step:9:87):

```
/home/runner/.nuget/packages/microsoft.playwright/1.41.0/buildTransitive/Microsoft.Playwright.targets(65,5): error MSB3073: The command "chmod +x """ exited with code 1. [/home/runner/work/dotnet-patch-automation-sample/dotnet-patch-automation-sample/tests/TodoApp.Tests/TodoApp.Tests.csproj]
```

It appears that something is going wrong here:

https://github.com/microsoft/playwright-dotnet/blob/e964457aa770c39ecb8f3605da512791b6f527a9/src/Playwright/build/Microsoft.Playwright.targets#L65

martincostello:
You could also use [`EnsureTrailingSlash`](https://learn.microsoft.com/en-us/visualstudio/msbuild/property-functions?view=vs-2022#msbuild-ensuretrailingslash).

campersau:
|  | `OutDir` | `OutputPath` | `PublishDir` |
| --- | --- | --- | --- |
| `dotnet build` | :white_check_mark: `bin/Debug/net8.0/` | :white_check_mark: `bin/Debug/net8.0/` | :white_check_mark: `bin/Debug/net8.0/publish/` |
|  `dotnet build --output bin/Custom` | :white_check_mark: `bin/Custom/` | :x: `bin/Custom` | :interrobang: `bin/Custompublish/` |
| `dotnet publish` | :white_check_mark: `bin/Release/net8.0/` | :white_check_mark: `bin/Release/net8.0/` |  :white_check_mark: `bin/Release/net8.0/publish/` |
[...]

PR #2829, opened 2024-01-17
fix: build with --output without trailing slash

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2827

My first attempt was using `OutDir` but it broke some of the LocalNugetTests. Instead we now call EnsureTrailingSlash.

### DIFF
--- a/src/Playwright/build/Microsoft.Playwright.targets
+++ b/src/Playwright/build/Microsoft.Playwright.targets
@@ -60,7 +60,7 @@
     <Target Name="PlaywrightChmodExecutables" AfterTargets="CopyFilesToOutputDirectory;CopyFilesToPublishDirectory" Condition="!$([MSBuild]::IsOSPlatform('Windows'))">
     <ItemGroup>
       <_PlaywrightChmodItems Include="$([MSBuild]::EnsureTrailingSlash('$(OutputPath)')).playwright\node\*\node" />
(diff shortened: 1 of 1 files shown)


src/Playwright/build/Microsoft.Playwright.targets
```diff
@@ -59,7 +59,7 @@
   </Target>
     <Target Name="PlaywrightChmodExecutables" AfterTargets="CopyFilesToOutputDirectory;CopyFilesToPublishDirectory" Condition="!$([MSBuild]::IsOSPlatform('Windows'))">
     <ItemGroup>
-      <_PlaywrightChmodItems Include="$(OutputPath).playwright\node\*\node" />
+      <_PlaywrightChmodItems Include="$([MSBuild]::EnsureTrailingSlash('$(OutputPath)')).playwright\node\*\node" />
       <_PlaywrightChmodItems Include="$(PublishDir).playwright\node\*\node" />
```
campersau, line 63:
```suggestion
      <_PlaywrightChmodItems Include="$([MSBuild]::EnsureTrailingSlash('$(PublishDir)')).playwright\node\*\node" />
```

campersau:
FYI: you used `OutputDir` instead of `OutDir` in your first commit: https://github.com/microsoft/playwright-dotnet/commit/faf6ea0c3f81196bb3af60e53dfb4cd060b034ef

yury-s (approved):

Issue #2847, opened 2024-01-26
[BUG] Fresh Playwright for .NET Install Throws 'Invalid browser name'


djlawhead:
### System info
- Playwright Version: 1.41.1
- Operating System: Ubuntu 22  (devcontainer dotnet:8.0-jammy)
- Browser: All
- Other info: .NET 8.0.101

### Source code

```csharp
using System.Text.RegularExpressions;
using System.Threading.Tasks;
using Microsoft.Playwright;
using Microsoft.Playwright.NUnit;
using NUnit.Framework;

namespace PlaywrightTests;

[Parallelizable(ParallelScope.Self)]
[TestFixture]
public class Tests : PageTest
{
    [Test]
    public async Task HomepageHasPlaywrightInTitleAndGetStartedLinkLinkingtoTheIntroPage()
    {
        await Page.GotoAsync("https://playwright.dev");

        // Expect a title "to contain" a substring.
        await Expect(Page).ToHaveTitleAsync(new Regex("Playwright"));

        // create a locator
        var getStarted = Page.GetByRole(AriaRole.Link, new() { Name = "Get started" });

        // Expect an attribute "to be strictly equal" to the value.
        await Expect(getStarted).ToHaveAttributeAsync("href", "/docs/intro");

        // Click the get started link.
        await getStarted.ClickAsync();

        // Expects the URL to contain intro.
        await Expect(Page).ToHaveURLAsync(new Regex(".*intro"));
    }
}
```

**Steps**
- Follow installation steps outlined in https://playwright.dev/dotnet/docs/intro
- dotnet test -- NUnit.NumberOfTestWorkers=5 

**Expected**

Playwright test execute successfully in one or more browsers, whether they pass or not.

**Actual**

```
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
  Failed HomepageHasPlaywrightInTitleAndGetStartedLinkLinkingtoTheIntroPage [228 ms]
  Error Message:
   System.ArgumentException : Invalid browser name: /vscode/vscode-server/bin/linux-x64/8b3775030ed1a69b13e4f4c628c612102e30a681/bin/helpers/browser.sh
  Stack Trace:
     at Microsoft.Playwright.TestAdapter.PlaywrightSettingsProvider.ValidateBrowserName(String browserName) in /_/src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs:line 102
[...]

PR #2849, opened 2024-01-29
chore: improve invalid browser name by env exception

mxschmitt:
We could warn as well, but it seems not a helpful warning where the user can act on.

Fixes https://github.com/microsoft/playwright-dotnet/issues/2847.

### DIFF
--- a/src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
+++ b/src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
@@ -42,13 +42,13 @@ public class PlaywrightSettingsProvider : ISettingsProvider
             var browserFromEnv = Environment.GetEnvironmentVariable("BROWSER")?.ToLowerInvariant();
             if (!string.IsNullOrEmpty(browserFromEnv))
             {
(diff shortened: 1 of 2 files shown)


yury-s (commented):
Why is throwing bad here?

src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
```diff
@@ -93,13 +93,15 @@ public static BrowserTypeLaunchOptions LaunchOptions
...
 
-    private static void ValidateBrowserName(string browserName)
+    private static void ValidateBrowserName(string browserName, string fromText)
     {
         if (browserName != BrowserType.Chromium &&
             browserName != BrowserType.Firefox &&
             browserName != BrowserType.Webkit)
         {
-            throw new ArgumentException($"Invalid browser name: {browserName}");
+            throw new ArgumentException($"Invalid browser name from {fromText}.\n" +
```
dgozman, line 102: Perhaps we can recommend unsetting the env in this error message?

dgozman (approved):

Issue #2603, opened 2023-06-05
[Feature] Set Default global LocatorAssertions in .NET.


MWRAUTO:
Hello! as we spoke in this issue: https://github.com/microsoft/playwright/issues/23471

I would like to request for a posibility to set expect global timeout while we're setting up browser context, same as we have default timeout & default navigation timeout in screenshot below.

![image](https://github.com/microsoft/playwright-dotnet/assets/128363781/63dc0c16-ecc8-4062-ab6f-53582140313c)

![image](https://github.com/microsoft/playwright-dotnet/assets/128363781/480e4332-c6f4-436a-a6f0-f07e82ddba78)

Materix:
As workaround I use:
```
        typeof(Playwright).Assembly.GetType("Microsoft.Playwright.Core.AssertionsBase")!
            .GetMethod("SetDefaultTimeout", BindingFlags.Static | BindingFlags.NonPublic)!
            .Invoke(null, new object[] { _defaultTimeout });
```

in any static ctor which are always loaded

SephVin:
Thanks for workaround! I really don't understand why we can't set default timeout for Assertions without using runsettings. This  file works only in VS

Melkor972:
Totally on the same page. Unless the workaround above is applied, Playwright makes pretty much obliges users to use runsettings file while configuration from code/custom config sources might be preferable to someone

mxschmitt:
Folding into #2259

Issue #2259, opened 2022-08-17
[Feature] Expose Ability to Set Expect Timeout


bjharo:
With 1.25, we have the ability to set the default timeout for web-first assertions (Expect) via the runsettings file but only if your testing project is using the Microsoft.Playwright.NUnit package. You can see in `WorkerAwareTest` where the timeout is getting set in the SetUp method:
```
        [SetUp]
        public void WorkerSetup()
        {
            if (!_allWorkers.TryPop(out _currentWorker))
            {
                _currentWorker = new();
            }
            WorkerIndex = _currentWorker.WorkerIndex;
            if (PlaywrightSettingsProvider.ExpectTimeout.HasValue)
            {
                AssertionsBase.SetDefaultTimeout(PlaywrightSettingsProvider.ExpectTimeout.Value);
            }
        }
```

The call to AssertionsBase.SetDefaultTimeout seems to set the default timeout. I've been wanting to set a default timeout for Expect assertions since they first came to Playwright.NET but unfortunately, our testing project doesn't use Microsoft.Playwright.NUnit and it appears that AssertionsBase is an internal class so there is no way we can set the timeout.

It would be great if this functionality was exposed in some way so that those of use not using Microsoft.Playwright.NUnit can set a default.

marcuskrahl:
This setting is not only exposed to Microsoft.Playwright.NUnit. 

We have a .runsettings file like:
``` xml
<?xml version="1.0" encoding="utf-8"?>
<RunSettings>
	<!-- Playwright specific parameters -->
	<Playwright>
		<ExpectTimeout>30000</ExpectTimeout>
	</Playwright>
</RunSettings>
```
The timeout is then correctly applied even for MSTest in our case by running the tests via `dotnet test -s .runsettings`.

So the scope of this issue is probably just to expose this property via C# like `Context.SetDefaultTimeout`

bryanh-bloomerang:
I've tried placing this setting in my .runsettings file as you suggest @marcuskrahl but the timeout for my Expect calls still defaults to 5000ms regardless of what value I put in the setting.

Philip-DeLuca:
I second this.  I have been trying to figure this out for 3 full days now.  When I apply the ExpectTimout parameter to my .runsettings file, I am still getting error messages that say "5000ms" for Expect calls.  My problem may be slightly different though- as I am using only Microsoft.Playwright.NUnit for the projects.

bryanh-bloomerang:
[...]

PR #2851, opened 2024-01-29
feat: expose SetDefaultExpectTimeout()

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2259
Fixes https://github.com/microsoft/playwright-dotnet/issues/2603

~~Maybe we should introduce Expect().Configure like we have in Node.js: https://playwright.dev/docs/test-assertions#expectconfigure~~

### DIFF
--- a/src/Playwright/Assertions.cs
+++ b/src/Playwright/Assertions.cs
@@ -28,6 +28,15 @@ namespace Microsoft.Playwright;
 
 public static class Assertions
 {
+    /// <summary>
+    /// Sets the default timeout for all future <c>Expect</c> calls.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


src/Playwright/Assertions.cs
```diff
@@ -1,9 +1,35 @@
...
+ * SOFTWARE.
+ */
+
 using Microsoft.Playwright.Core;
 
 namespace Microsoft.Playwright;
 
 public static class Assertions
 {
+    public static void SetDefaultExpectTimeout(float timeout) => AssertionsBase.SetDefaultTimeout(timeout);
```
dgozman, line 31: Let's add a brief documentation comment here.

dgozman (approved):

PR #2855, opened 2024-02-07
devops: migrate to 1ES PT

mxschmitt:
See here: https://dev.azure.com/securitytools/SecurityIntegration/_wiki/wikis/Guardian/1385/Suppressions

The `.gdnsuppress` file got obtained during a previous test-run where it was in the artifacts section of the SDL stage. They are like `.eslintignore`, but instead of the file location, they contain the location, type, reason etc. in a hashed form, for security reasons.

When to update? When SDL checks are red in Azure Pipelines.
How to update? See the link, based on the artifact it can be just downloaded / replaced.

---

How is it tested? 1.42.0-beta-1 is published.

### DIFF
--- a/.build/azure-pipeline-nuget.yml
+++ b/.build/azure-pipeline-nuget.yml
@@ -1,260 +1,247 @@
 trigger: none # We don't want CI builds, just a manual release process
-pool: $(PlaywrightPoolName)
 parameters:
-  - name: doRelease
-    displayName: Push the Playwright Release to NuGet.org
-    default: false
-    type: boolean
+- name: doRelease
+  displayName: Push the Playwright Release to NuGet.org
+  default: false
+  type: boolean
 
-  - name: doReleaseCLI
-    displayName: Push the CLI Release to NuGet.org
-    default: false
-    type: boolean
+- name: doReleaseCLI
+  displayName: Push the CLI Release to NuGet.org
+  default: false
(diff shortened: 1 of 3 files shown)


.build/azure-pipeline-nuget.yml
```diff
@@ -1,260 +1,247 @@
...
+  repositories:
+  - repository: 1esPipelines
+    type: git
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1esPipelines
+  parameters:
+    pool:
+      name: VSEngSS-MicroBuild2022-1ES
```
mxschmitt, line 35: Note: This was previously defined in the AzDo web-interfaces, this moves it into the yaml.

yury-s: In the web-interface there was autocomplete. How do we find a new value once this one goes out of fashion? Let's add instructions/link here?

yury-s (approved):

.build/azure-pipeline-nuget.yml
mxschmitt: It was previously hard-coded without auto-complete, a string variable. See https://aka.ms/MicroBuild for the future.

Issue #2874, opened 2024-02-28
[Bug]: ClickOnce deployment error due to mismatched assembly identity in manifest


shunkino:
### Version

1.41.2

### Steps to reproduce

Example steps:
1. Clone my repo at https://github.com/shunkino/PlaywrightDemo/
2. Publish the app and deploy it as ClickOnce application
3. Try installing the app through web or file share
4. Install will fail with System.Deployment.Application.InvalidDeploymentException

You can also use following link to quickly repro the issue. (click 'launch' link to install ClickOnce application)
https://coplaywrightdotnet.z11.web.core.windows.net/publish.htm

### Expected behavior

ClickOnce application will boot without issue even if it references playwright-dotnet.

### Actual behavior

ClickOnce application will fail with System.Deployment.Application.InvalidDeploymentException with following error message.
```
Reference in the manifest does not match the identity of the downloaded assembly Microsoft.Playwright.dll.
```

### Additional context

I investigated the detailed log of ClickOnce and found that when the error occurred, the validation of a manifest called id1Manifest was performed as follows.
```
[23:30:49] : id1Manifest is parsed successfully.
[23:30:49] : _id1Identity = MyApplication.app, Version=1.0.0.0, Culture=neutral
[23:30:49] : Method Call : AssemblyManifest.ExtractIdentityFromCompLibAssembly(C:\Users\shkinosh\AppData\Local\Temp\Deployment\2NZW3QHW.X0T\HYJB9BL0.NBP\Microsoft.Playwright.dll) called.
[23:30:49] : Managed Assembly Identity = Microsoft.Playwright, Version=1.41.2.0, Culture=neutral, PublicKeyToken=db12f80f85d8ba23, processorArchitecture=MSIL
[23:30:49] : _complibIdentity =Microsoft.Playwright, Version=1.41.2.0, Culture=neutral, PublicKeyToken=db12f80f85d8ba23, processorArchitecture=MSIL
```
The error was caused by the mismatch of the assembly identity in this validation. We also debugged the ClickOnce loader and confirmed that the System.Deployment.Application.InvalidDeploymentException with the above message was thrown when the id1Manifest and the manifest embedded in the .NET header were inconsistent.

The id1Manifest here refers to the manifest included in the Application manifests described in this article. Manifest is defined in XML file and can be embedded as a resource in the compiled file.
https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests

[...]

PR #2881, opened 2024-03-11
chore: set <NoWin32Manifest>true</NoWin32Manifest>

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2874

cc @nohwnd - looks like a low-risk change I guess.


nohwnd:
I don't know what that switch does exactly, but from the comments in this thread, and the suggestion from jkotas being that it is disabled for roslyn, I think disabling it (set it to true, as the PR does) is the correct choice here. 

https://github.com/dotnet/sdk/issues/12904#issuecomment-1518242402

nohwnd (approved):

mxschmitt:
Thank you!

PR #2887, opened 2024-03-18
fix: restrict access to _activeInvocations in concurrent RouteHandler methods …

Brimerland:
… because HashSet is not thread-safe

Access to RouteHandle._activeInvocations must be controlled using a lock. Otherwise IndexOutOfBounds exceptions or InvalidOperation exceptions may be triggered during concurrent requests. HashSet is not thread safe.

A test "ShouldNotThrowOnConcurrentRequest" is added which reproduces the issue by having 10 pages requesting a http resource at high load for 5 seconds.

### DIFF
--- a/src/Playwright.Tests/BrowserContextRouteTests.cs
+++ b/src/Playwright.Tests/BrowserContextRouteTests.cs
@@ -204,36 +204,20 @@ public class BrowserContextRouteTests : BrowserTestEx
         });
 
 
-        // subscribe for unobserved exceptions since we can not catch the exception
-        // in RouteHandler.cs directly
-        Exception? unobservedException = null;
-        void _handleUnobservedException(object sender, UnobservedTaskExceptionEventArgs args)
-            => unobservedException = args.Exception;
+        // open 10 pages to generate load on the RouteHandler
+        foreach (int pageNr in Enumerable.Range(0, 10))
+            await (await context.NewPageAsync())
+                .GotoAsync(Server.EmptyPage);
 
-        try
-        {
-            TaskScheduler.UnobservedTaskException += _handleUnobservedException;
+        // let the test run for 5 second
+        await Task.Delay(5000);
 
-            // open 10 pages to generate load on the RouteHandler
-            foreach (int pageNr in Enumerable.Range(0, 10))
-                await (await context.NewPageAsync())
-                    .GotoAsync(Server.EmptyPage);
+        // trigger garbage collection which checks for unorbserved exceptions
+        // in collected tasks
+        GC.Collect();
+        GC.WaitForPendingFinalizers();
 
-            // let the test run for 5 second
-            await Task.Delay(5000);
-
-            // trigger garbage collection which checks for unorbserved exceptions
-            // in collected tasks
-            GC.Collect();
-            GC.WaitForPendingFinalizers();
-
-            // check for exceptions
-            Assert.Null(unobservedException);
-        }
-        finally
-        {
-            TaskScheduler.UnobservedTaskException -= _handleUnobservedException;
-        }
+        // unobserved task exceptions are automatically collected by the PlaywrightTest attribute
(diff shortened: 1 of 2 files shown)


Brimerland:

Brimerland:

mxschmitt:
What do you think about the following approach instead? Thats how we usually deal with such issues:

```diff
diff --git a/src/Playwright/Core/RouteHandler.cs b/src/Playwright/Core/RouteHandler.cs
index 10d27376..42214217 100644
--- a/src/Playwright/Core/RouteHandler.cs
+++ b/src/Playwright/Core/RouteHandler.cs
@@ -23,6 +23,7 @@
  */
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
@@ -32,7 +33,7 @@ namespace Microsoft.Playwright.Core;
 
 internal class RouteHandler
 {
-    private readonly HashSet<HandlerInvocation> _activeInvocations = new HashSet<HandlerInvocation>();
+    private readonly IDictionary<HandlerInvocation, byte> _activeInvocations = new ConcurrentDictionary<HandlerInvocation, byte>();
 
     private bool _ignoreException;
 
@@ -88,7 +89,7 @@ internal class RouteHandler
             Complete = new TaskCompletionSource<bool>(),
             Route = route,
         };
-        _activeInvocations.Add(handlerInvocation);
+        _activeInvocations.Add(handlerInvocation, 0);
         try
         {
             return await HandleInternalAsync(route).ConfigureAwait(false);
@@ -122,7 +123,7 @@ internal class RouteHandler
         else
         {
             var tasks = new List<Task>();
-            foreach (var activation in _activeInvocations)
+            foreach (var activation in _activeInvocations.Keys)
             {
                 if (!activation.Route.DidThrow)
                 {

```

src/Playwright/Core/RouteHandler.cs
```diff
@@ -88,7 +89,10 @@ public async Task<bool> HandleAsync(Route route)
             Complete = new TaskCompletionSource<bool>(),
             Route = route,
         };
-        _activeInvocations.Add(handlerInvocation);
+        lock (_activeInvocationsLock)
```
kblok, line 92: Did you try with some thread-safe list instead of adding locks?

Brimerland: Of course, this can also be solved with concurrent collections, which is also my go-to option for these kinds of problems.
But since we are not in a performance-critical inner loop and the lock is only held for a very short time without user code being called, this is a simple and effective solution with minimal changes to the code.

But I certainly don't insist on this solution. If an alternative solution with the replacement of the type used is desired, then I will do that.

src/Playwright.Tests/BrowserContextRouteTests.cs
```diff
@@ -177,6 +177,65 @@ public async Task ShouldFallBackToContextRoute()
...
+
+        // subscribe for unobserved exceptions since we can not catch the exception
+        // in RouteHandler.cs directly
+        Exception? unobservedException = null;
+        void _handleUnobservedException(object sender, UnobservedTaskExceptionEventArgs args)
+            => unobservedException = args.Exception;
+
+        try
+        {
+            TaskScheduler.UnobservedTaskException += _handleUnobservedException;
```
mxschmitt, line 215: We have a similar logic already in place [here](https://github.com/microsoft/playwright-dotnet/blob/04431afbef20dfffe48022b8e953077526d14850/src/Playwright.Tests/Attributes/PlaywrightTestAttribute.cs#L139) so I'm wondering if we can remove 213-216 and 225-236. Keeping the Task.Delay seems essential but the rest is already covered? For me it fails constantly without the fix, so it seems tested?

mxschmitt (approved):

Issue #2888, opened 2024-03-18
[Bug]: 'The parameter is incorrect' at System.ConsolePal.SetConsoleInputEncoding(Encoding enc)


adamwoods1vitalitycouk:
### Version

1.42.0

### Steps to reproduce

Install Rider version 2022.3.0
Install Rider version 2023.3.4
Within any project, **Debug** (Not run) a test with the following line of code
`var playwright = await Playwright.CreateAsync();`

### Expected behavior

Playwright to initialise in both versions when run in debug mode.

### Actual behavior

Playwright raises an exception but only in Rider 2023.3.4 when in debug mode,

### Additional context

**Exception thrown**
```
System.IO.IOException: The parameter is incorrect.
   at System.ConsolePal.SetConsoleInputEncoding(Encoding enc)
   at System.Console.set_InputEncoding(Encoding value)
   at Microsoft.Playwright.Transport.StdIOTransport.StartProcessWithUTF8IOEncoding(Process process) in /_/src/Playwright/Transport/StdIOTransport.cs:line 175
   at Microsoft.Playwright.Transport.StdIOTransport..ctor() in /_/src/Playwright/Transport/StdIOTransport.cs:line 47
   at Microsoft.Playwright.Playwright.CreateAsync() in /_/src/Playwright/Playwright.cs:line 43
   at SalesforceServicingOrg.Foundation.Hooks.CommonHooks.BeforeScenario() in D:\Users\adam.woods\RiderProjects\salesforceservicingorg.ui.tests\SalesforceServicingOrg.Foundation\Hooks\CommonHooks.cs:line 31
```

**Additional**
We are unable to start any tests with a parameter is incorrect in StdIOTransport.cs:line 175. The issue seems to be encoding from the console. Although despite all Windows and Rider settings pointing to encoding that matches UTF-8 - we still seem to hit this.  We can execute tests in run mode, but not in debug mode. There appears to be no obvious settings that have changed between Rider versions - but it is unclear where Playwright gets the encoding setting from. Out of 6 team members, only 1 can run in the latest Rider version.

### Environment

```shell
System:
    OS: Windows 10 10.0.17763
    CPU: (4) x64 Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz
    Memory: 7.24 GB / 16.00 GB
  Binaries:
    Node: 18.12.1 - C:\Program Files\nodejs\node.EXE
[...]

PR #2901, opened 2024-04-03
fix(CreateAsync): try/catch restoring of Input/Output Encoding

mxschmitt:
See https://github.com/microsoft/playwright-dotnet/issues/2888#issuecomment-2032027194

Fixes https://github.com/microsoft/playwright-dotnet/issues/2888

### DIFF
--- a/src/Playwright/Transport/StdIOTransport.cs
+++ b/src/Playwright/Transport/StdIOTransport.cs
@@ -171,9 +171,17 @@ internal class StdIOTransport : IDisposable
         {
             if (hasConsole)
             {
-                // Restore the original encodings
-                Console.InputEncoding = originalInputEncoding;
-                Console.OutputEncoding = originalOutputEncoding;
+                try
+                {
+                    // Restore the original encodings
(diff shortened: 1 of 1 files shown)


src/Playwright/Transport/StdIOTransport.cs
```diff
@@ -171,9 +171,17 @@ private static void StartProcessWithUTF8IOEncoding(Process process)
...
-                // Restore the original encodings
-                Console.InputEncoding = originalInputEncoding;
-                Console.OutputEncoding = originalOutputEncoding;
+                try
+                {
+                    // Restore the original encodings
+                    Console.InputEncoding = originalInputEncoding;
+                    Console.OutputEncoding = originalOutputEncoding;
+                }
+                catch (System.Exception)
```
nohwnd, line 180:
```suggestion
                catch (System.IO.IOException)
```
I would be more strict on the exception that is being caught to not accidentally hide additional problems.

mxschmitt: sg

Issue #2896, opened 2024-03-29
[Ports]: Backport client side changes


playwrightmachine:
Please backport client side changes:
  - [ ] https://github.com/microsoft/playwright/commit/a950a7b86b3977e98ef9f2bcfc90817039e38808 (feat: Support intervals arg for toPass (#30037))

PR #2905, opened 2024-04-09
feat(roll): roll Playwright to 1.43.0-beta-1712646596000

mxschmitt:
This was added after we rolled the last time.

https://github.com/microsoft/playwright-dotnet/issues/2896

### DIFF
--- a/src/Playwright/Core/BrowserContext.cs
+++ b/src/Playwright/Core/BrowserContext.cs
@@ -44,6 +44,7 @@ internal class BrowserContext : ChannelOwner, IBrowserContext
     private readonly Dictionary<string, Delegate> _bindings = new();
     private readonly BrowserContextInitializer _initializer;
     private readonly Tracing _tracing;
+    internal readonly HashSet<IPage> _backgroundPages = new();
     internal readonly IAPIRequestContext _request;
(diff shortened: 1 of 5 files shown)


src/Playwright/Core/BrowserContext.cs
```diff
@@ -44,6 +44,7 @@ internal class BrowserContext : ChannelOwner, IBrowserContext
     private readonly Dictionary<string, Delegate> _bindings = new();
     private readonly BrowserContextInitializer _initializer;
     private readonly Tracing _tracing;
+    internal readonly HashSet<IPage> _backgroundPages = new();
```
campersau, line 47: Why does this use a HashSet intead of List?

mxschmitt: Hmm, was following [upstream impl](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/browserContext.ts#L61), doesn't really matter probably, Java uses a [List](https://github.com/microsoft/playwright-java/commit/452effbe3f81c596d6f40c02b236efde290c90bf#diff-24e3a8cd1a7d4f1c30ddf3f3550359d3d9ad77c879e1b5f51a5b7b7f3c8d8ad9R50) as well.

dgozman (approved):

Issue #2255, opened 2022-08-15
[BUG] Exception thrown when using playwright in a single-file published program


jeroenpelgrims:
**Context:**
- Playwright Version: 1.24.1
- Operating System: Windows 10
- .NET version: .NET 6
- Browser: All
- Extra: 

**Code Snippet**

See code to reproduce this bug here: https://github.com/resurge/PlaywrightSingleFilePathError 

**Describe the bug**
When publishing a project as single file and running the published executable the following error will be shown:
```
Unhandled exception. System.ArgumentException: The path is empty. (Parameter 'path')
   at System.IO.Path.GetFullPath(String path)
   at System.IO.FileInfo..ctor(String originalPath, String fullPath, String fileName, Boolean isNormalized)
   at System.IO.FileInfo..ctor(String fileName)
   at Microsoft.Playwright.Helpers.Driver.GetExecutablePath() in /_/src/Playwright/Helpers/Driver.cs:line 41
   at Microsoft.Playwright.Transport.StdIOTransport.GetProcess() in /_/src/Playwright/Transport/StdIOTransport.cs:line 119
   at Microsoft.Playwright.Transport.StdIOTransport..ctor() in /_/src/Playwright/Transport/StdIOTransport.cs:line 44
   at Microsoft.Playwright.Playwright.CreateAsync() in /_/src/Playwright/Playwright.cs:line 44
   at Program.<Main>$(String[] args) in C:\Users\jeroe\Documents\repos\PlaywrightSingleFilePathError\Program.cs:line 3
   at Program.<Main>(String[] args)
```
I would expect the program to behave the same way as a regular Debug or Release build.  
But choosing to publish as a single file causes the behaviour of the executable to change. (And not work)

The following tickets were about the same bug but were closed because of no code to reproduce:
- https://github.com/microsoft/playwright-dotnet/issues/1952
- https://github.com/microsoft/playwright-dotnet/issues/1527

EDIT: Some other issues I found relating to this bug:
- https://github.com/microsoft/playwright-dotnet/issues/1069
  Not sure how the solution here works because `Playwright.CreateAsync` doesn´t accept parameters.
- https://github.com/microsoft/playwright-dotnet/issues/1274
   This ticket is closed with comment "not supported", but no reason is given as to why not.

kblok:
My two cents here. I don't think an end-to-end testing library needs to support single-file publishing.

resurge:
[...]

PR #2909, opened 2024-04-15
fix: support self-contained executables

mxschmitt:
Should not be any big logical changes, just safeguards the access to `assembly.CodeBase` and falls-back to `AppContext.BaseDirectory` otherwise which is important for the self-contained executables.

https://github.com/microsoft/playwright-dotnet/issues/2255

### DIFF
--- a/src/Playwright/Helpers/Driver.cs
+++ b/src/Playwright/Helpers/Driver.cs
@@ -26,6 +26,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Reflection;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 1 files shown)


nohwnd (approved):
This looks like it should work, would be nice to have a test that proves it works. But if this is not a main scenario for you it seems okay to keep as is.

mxschmitt:
Tested on macOS and Windows. One thing to note is that `playwright.ps1` does not work, this is because its a single executable and the ps1 is looking for `Microsoft.Playwright.dll`.

Workaround: `Microsoft.Playwright.Program.Main(["install"]);`

Will keep the issue open for now, but this patch should help the users already a lot.

PR #2910, opened 2024-04-15
chore: bundle Node.js per rid via runtime.json

mxschmitt:


mxschmitt:
This is a feature which is not publicly exposed by nuget and might be removed in the future.

Issue #2912, opened 2024-04-16
[Bug]: Custom/Unknown CDP events cause playwright to close


joelgriffith:
### Version

1.43.0

### Steps to reproduce

Difficult to reproduce as it involves creating a custom server that intercepts WebSocket messages and emits custom events in the CDP protocol. Happy to set up something for reproducing as we have our own custom CDP events.

### Expected behavior

Unknown CDP events should be allowed to propagate through to an event handler if there is one, and ignored if not. The connection shouldn't be closed when unknown CDP events are raised.

### Actual behavior

Any unknown CDP events cause playwright to close the connection when connecting over CDP.

### Additional context

We override the CDP event space with custom ones for browserless: https://docs.browserless.io/open-api#section/The-Browserless-CDP-API, and have noticed that most playwright libraries work flawlessy with this, however, the .NET variation closes anytime an unknown CDP message is sent (even when there's an event handler for it). I'm happy to go over this in further detail or provide temporary environments where this behavior is exhibited. Happy to also to try and provide a PR as well if possible.

### Environment

```Text
- Operating System: [macOS 14.4.1]
- CPU: [arm64]
- Browser: [Chromium]
- .NET Version (TFM): [net8.0]
```

mxschmitt:
Do you know where it crashes?

Feel free to send me a repro, either here if public is fine or on Discord, then we can take a look.

joelgriffith:
@mxschmitt I'll work on a reproduction script for you and get established in your Discord group, I don't necessarily think it's a full-on crash but something is closing the underlying WebSocket connection, rather. Thanks for the response!

joelgriffith:
Is there a preferred person or channel to start? I'm in that discord channel

PR #2913, opened 2024-04-17
fix(cdpSession): events without payload

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2912

Aligns it with JavaScript: https://github.com/microsoft/playwright/blob/82aefd24db1ba9977c3f6953347ee266845790ea/packages/playwright-core/src/client/cdpSession.ts#L31

And with Python: https://github.com/microsoft/playwright-python/pull/2411

And with Java: https://github.com/microsoft/playwright-java/pull/1553

### DIFF
--- a/src/Playwright/Core/CDPSession.cs
+++ b/src/Playwright/Core/CDPSession.cs
@@ -45,7 +45,8 @@ internal class CDPSession : ChannelOwner, ICDPSession
         switch (method)
         {
             case "event":
-                OnCDPEvent(serverParams!.Value.GetProperty("method").ToString(), serverParams.Value.GetProperty("params"));
+                serverParams!.Value.TryGetProperty("params", out var cdpParams);
+                OnCDPEvent(serverParams!.Value.GetProperty("method").ToString(), cdpParams);
                 break;
         }
     }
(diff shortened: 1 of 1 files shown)


src/Playwright/Core/CDPSession.cs
```diff
@@ -45,7 +45,8 @@ internal override void OnMessage(string method, JsonElement? serverParams)
         switch (method)
         {
             case "event":
-                OnCDPEvent(serverParams!.Value.GetProperty("method").ToString(), serverParams.Value.GetProperty("params"));
+                serverParams!.Value.TryGetProperty("params", out var cdpParams);
+                OnCDPEvent(serverParams!.Value.GetProperty("method").ToString(), cdpParams);
```
campersau, line 49: Since `JsonElement` is a struct it might be nicer to pass `null` instead of a `default(JsonElement)`? Looks like `OnCDPEvent` accepts `null`.

```suggestion
                OnCDPEvent(serverParams!.Value.GetProperty("method").ToString(), serverParams!.Value.TryGetProperty("params", out var cdpParams) ? cdpParams : null);
```
(not sure if that compiles)

PR #2928, opened 2024-05-01
chore: delete not-exposed test-retries implementation

mxschmitt:
This was never public, we used retries on our internal tests only. 

So lets either re-add it for only our internal tests only (because before it was half public half internal) or stick with NUnit defaults: e.g. https://docs.nunit.org/articles/nunit/writing-tests/attributes/retry.html.


yury-s (approved):
> So lets either re-add it for only our internal tests only (because before it was half public half internal) or stick with NUnit defaults:

If you want to keep it then let's just change the implementation so that it's private?

mxschmitt:
> > So lets either re-add it for only our internal tests only (because before it was half public half internal) or stick with NUnit defaults:
> 
> If you want to keep it then let's just change the implementation so that it's private?

I'd try to live without for now - maybe revisit it in the future.

PR #2943, opened 2024-06-04
chore: update service api signature

Sid200026:


src/Playwright.NUnit/BrowserService.cs
```diff
@@ -56,19 +55,17 @@ private static async Task<IBrowser> CreateBrowser(IBrowserType browserType)
...
-        var caps = new Dictionary<string, string>
-        {
-            ["os"] = Environment.GetEnvironmentVariable("PLAYWRIGHT_SERVICE_OS") ?? "linux",
-            ["runId"] = Environment.GetEnvironmentVariable("PLAYWRIGHT_SERVICE_RUN_ID") ?? DateTime.Now.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ss", CultureInfo.InvariantCulture),
-        };
-        var wsEndpoint = $"{serviceUrl}?cap={JsonSerializer.Serialize(caps)}";
+        var os = Environment.GetEnvironmentVariable("PLAYWRIGHT_SERVICE_OS") ?? "linux";
+        var runId = Environment.GetEnvironmentVariable("PLAYWRIGHT_SERVICE_RUN_ID") ?? DateTime.Now.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ss", CultureInfo.InvariantCulture);
+        var apiVersion = "2023-10-01-preview";
+        var wsEndpoint = $"{serviceUrl}?os={os}&runId={runId}&api-version={apiVersion}";
```
mxschmitt, line 61: Ideally we encode the values here, since user input might contain special characters. Maybe with https://learn.microsoft.com/en-us/dotnet/api/system.uri.escapedatastring?view=net-8.0

mxschmitt (approved):

PR #2949, opened 2024-06-20
chore(roll): roll to Playwright v1.45

mxschmitt:
To be discussed:

1. APIRequestContext.DisposeAsync(reason=...) is not ported yet - I also saw that we don't have BrowserContext.close(reason...) in .NET
2. We should evaluate if:
  ```
        Time = clone.Time;
        TimeDate = clone.TimeDate;
        TimeInt64 = clone.TimeInt64;
        TimeString = clone.TimeString;
  ```
  is expected or if we should call it `TimeLong` instead.

https://github.com/microsoft/playwright-dotnet/issues/2941


src/Playwright/Core/BrowserContext.cs
```diff
@@ -130,6 +132,12 @@ public ITracing Tracing
         set => throw new NotSupportedException();
     }
 
+    public IClock Clock
+    {
+        get => _clock;
+        set => throw new NotSupportedException();
```
campersau, line 138: Is the setter needed?

src/Playwright/Core/Clock.cs
```diff
@@ -0,0 +1,104 @@
...
+
+internal class Clock(BrowserContext browserContext) : IClock
+{
+    public async Task InstallAsync(ClockInstallOptions options = null)
+    {
+        var timeIsGiven = options?.Time != null || options?.TimeInt64 != null || options?.TimeString != null || options?.TimeDate != null;
+        await browserContext.SendMessageToServerAsync("clockInstall", timeIsGiven ? ParseTime(options.Time ?? options.TimeString, options?.TimeDate, options?.TimeInt64) : null).ConfigureAwait(false);
+    }
+
+    private Dictionary<string, object> ParseTime(string timeString, DateTime? timeDate, long? timeInt64)
```
campersau, line 39: Nit static.
```suggestion
    private static Dictionary<string, object> ParseTime(string timeString, DateTime? timeDate, long? timeInt64)
```

src/Playwright/API/Generated/Options/ClockInstallOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        Time = clone.Time;
+        TimeDate = clone.TimeDate;
+        TimeInt64 = clone.TimeInt64;
+        TimeString = clone.TimeString;
+    }
+
+    /// <summary><para>Time to initialize with, current system time by default.</para></summary>
+    [JsonPropertyName("time")]
+    public string? Time { get; set; }
```
yury-s, line 51: Why do we have both Time and TimeDate?

src/Playwright/Core/Clock.cs
yury-s, line 39: why not 3 different overloaded methods?

yury-s (approved):

src/Playwright/Core/BrowserContext.cs
```diff
@@ -124,11 +126,9 @@ internal BrowserContext(ChannelOwner parent, string guid, BrowserContextInitiali
...
     public event EventHandler<IWorker> ServiceWorker;
 
-    public ITracing Tracing
-    {
-        get => _tracing;
-        set => throw new NotSupportedException();
-    }
+    public ITracing Tracing { get => _tracing; }
+
+    public IClock Clock { get => _clock; }
```
campersau, line 131: Nit: Could be even shorter like the other ones:

```suggestion
    public ITracing Tracing => _tracing;

    public IClock Clock => _clock;
```

src/Playwright/Core/Page.cs
```diff
@@ -182,6 +182,11 @@ public IMouse Mouse
         get;
     }
 
+    public IClock Clock
+    {
+        get => Context.Clock;
+    }
```
kblok, line 188:
```suggestion
    public IClock Clock => Context.Clock;
```

PR #2975, opened 2024-08-01
chore: align with upstream exception handling

mxschmitt:
https://github.com/microsoft/playwright-dotnet/issues/2969

### DIFF
--- a/src/Playwright.Tests/PageEvaluateTests.cs
+++ b/src/Playwright.Tests/PageEvaluateTests.cs
@@ -549,7 +549,7 @@ public class PageEvaluateTests : PageTestEx
     [PlaywrightTest("page-evaluate.spec.ts", "should evaluate exception with a function on the stack")]
     public async Task ShouldEvaluateExceptionWithAFunctionOnTheStack()
     {
-        var exception = await Page.EvaluateAsync<PlaywrightException>(@"() => {
+        var exception = await Page.EvaluateAsync<Exception>(@"() => {
             return (function functionOnStack() {
                 return new Error('error message');
             })();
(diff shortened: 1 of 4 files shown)


src/Playwright/Transport/Converters/EvaluateArgumentValueConverter.cs
```diff
@@ -123,6 +123,19 @@ internal static object Serialize(object value, List<EvaluateArgumentGuidElement>
             return new { bi = bigInteger.ToString(CultureInfo.InvariantCulture) };
         }
 
+        if (value is Exception exception)
+        {
+            return new Dictionary<string, object>
+            {
+                ["e"] = new Dictionary<string, object>
```
yury-s, line 130: What's the point in serializing .net error to a js error? I'd not add this at all, especially given that we not parse the error from js and just return a string.

src/Playwright.Tests/PageEvaluateTests.cs
```diff
@@ -547,24 +547,62 @@ public async Task ShouldNotLeakHandles()
...
     {
         var exception = await Page.EvaluateAsync<PlaywrightException>(@"() => {
             return (function functionOnStack() {
                 return new Error('error message');
             })();
         }");
-        Assert.Equals("error message", exception.Message);
-        StringAssert.Contains("functionOnStack", exception.StackTrace);
+        StringAssert.Contains("Error: error message", exception.Message);
+        StringAssert.Contains("functionOnStack", exception.Message);
```
yury-s, line 558: Shouldn't it be in exception.StackTrace ?

src/Playwright.Tests/PageEvaluateTests.cs
```diff
@@ -547,24 +547,62 @@ public async Task ShouldNotLeakHandles()
...
-        StringAssert.Contains("Error: error message", exception);
+        var exception = await Page.EvaluateAsync<object>(@"() => {
+            function innerFunction() {
+                const e = new Error('error message');
+                e.name = 'foobar';
+                return e;
+            }
+            return innerFunction();
+        }");
+        Assert.IsInstanceOf<PlaywrightException>(exception);
```
yury-s, line 572: I'd use a generic Exception instead of PlaywrightException

yury-s (approved):

src/Playwright.Tests/PageEvaluateTests.cs
mxschmitt: We don't re-serialise the JS error into a .NET StackTrace property yet, I can try doing it tho, we only do message as of today.

PR #2976, opened 2024-08-05
chore: remap persistent-context options into context options

mxschmitt:
drive-by: address two other TODO comments.

### DIFF
--- a/src/Playwright.Tests/ClientCertficatesTests.cs
+++ b/src/Playwright.Tests/ClientCertficatesTests.cs
@@ -118,7 +118,6 @@ public class ClientCertificatesTests : BrowserTestEx
     {
         var context = await Browser.NewContextAsync(new()
         {
-            // TODO: Remove this once we can pass a custom CA.
             IgnoreHTTPSErrors = true,
             ClientCertificates =
             [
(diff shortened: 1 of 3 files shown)


src/Playwright.Tests/ClientCertficatesTests.cs
```diff
@@ -118,7 +118,6 @@ public async Task ShouldWorkWithNewContext()
     {
         var context = await Browser.NewContextAsync(new()
         {
-            // TODO: Remove this once we can pass a custom CA.
```
mxschmitt, line 121: drive-by: remove TODO comments as per https://github.com/microsoft/playwright/pull/31973 - since we don't plan to follow-up there.

src/Playwright.Tests/PageRouteTests.cs
```diff
@@ -513,13 +513,9 @@ public async Task ShouldWorkWithEncodedServer2()
             route.ContinueAsync();
             requests.Add(route.Request);
         });
-        var response = await Page.GotoAsync($"data:text/html,<link rel=\"stylesheet\" href=\"{Server.EmptyPage}/fonts?helvetica|arial\"/>");
+        var response = await Page.GotoAsync($"data:text/html,<meta charset=utf-8><link rel=\"stylesheet\" href=\"{Server.EmptyPage}/fonts?helvetica|arial\"/>");
         Assert.Null(response);
-        // TODO: https://github.com/microsoft/playwright/issues/12789
```
mxschmitt, line 518: drive-by: remove TODO by applying their recommended upstream fix.

PR #2985, opened 2024-08-16
chore: change the target closed exception public

shuowpro:
When we make a call to the playwright when the browser is disconnected, the method may throw the TargetClosedException, but due to the TargetClosedException is marked as internal, it is hard to catch this exception as the user. this is a trivial change to make this possible


mxschmitt:
We recommend catching `PlaywrightException` instead and looking at the message if necessary. We don't feel comfortable exposing this exception, since its an internal implementation detail which might change over time (did in the past). Closing by that. Thanks for your understanding!

PR #2992, opened 2024-08-29
devops(tooling): Fix path to `README.md` in driver downloader

RenderMichael:
I followed the directions in `CONTRIBUTING.md` and ran `./build.sh --download-driver`.

When it ran `dotnet run --project ./src/tools/Playwright.Tooling/Playwright.Tooling.csproj -- download-drivers --basepath .` I got this following exception:

```bash
WARNING: Could not update the browser versions in the README file.
This is usually due to the readme file not yet existing for 1.46.0.
Could not find a part of the path 'D:\Code\RenderMichael\playwright\README.md'.
```

This fixes the issue for me. I assume it hasn't been hit before because everyone else also has the upstream `playwright` project as well.

### DIFF
--- a/src/tools/Playwright.Tooling/DriverDownloader.cs
+++ b/src/tools/Playwright.Tooling/DriverDownloader.cs
@@ -71,7 +71,7 @@ internal class DriverDownloader
 
             var regex = new Regex("<!-- GEN:(.*?) -->(.*?)<!-- GEN:stop -->", RegexOptions.Compiled);
 
-            var basePlaywrightDir = Environment.GetEnvironmentVariable("PW_SRC_DIR") ?? Path.Combine(Environment.CurrentDirectory, "..", "playwright");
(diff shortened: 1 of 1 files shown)


mxschmitt:
This would end being a noop, since it would read from the same `README.md` like it writes it to.

We could add a flag like `--update-readme` and only use this flag internally when rolling a new version via `./build.sh --roll 1.46.0`.

RenderMichael:
Ah got it. Sorry I didn't look closely enough at what the code was doing. My bad.

I can add a flag for that, or to avoid disruptions to flow I can add the inverse `--no-update-readme` flag, and add guidance to use that in `CONTRIBUTING.md`. Whatever y'all think is best.

mxschmitt:
I'll close it since it seem stale. Its just a non-user facing warning, so I think we don't need to perform any action. We could change the text of it if you are interested in making it more clear.

PR #3000, opened 2024-09-09
chore(roll): roll Playwright to v1.47.0

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2969

### DIFF
--- a/src/Playwright/Core/APIRequestContext.cs
+++ b/src/Playwright/Core/APIRequestContext.cs
@@ -29,6 +29,7 @@ using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text.Json;
 using System.Threading.Tasks;
+using System.Web;
 using Microsoft.Playwright.Helpers;
 using Microsoft.Playwright.Transport;
 using Microsoft.Playwright.Transport.Protocol;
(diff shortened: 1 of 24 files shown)


src/Playwright/Core/APIRequestContext.cs
```diff
@@ -164,6 +159,28 @@ public async Task<IAPIResponse> FetchAsync(string url, APIRequestContextOptions
         return new APIResponse(this, response?.GetProperty("response").ToObject<Transport.Protocol.APIResponse>());
     }
 
+    private static IEnumerable<NameValue> QueryParamsToProtocol(IEnumerable<KeyValuePair<string, object>> @params, string paramsString)
+    {
+        if (@params != null)
```
dgozman, line 164: Why `@params` and not `params`?

dgozman (approved):

src/Playwright/Core/APIRequestContext.cs
campersau: Reserved keyword I guess: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters#params-modifier

PR #3010, opened 2024-09-27
Update Playwright.csproj to upgrade Microsoft.Bcl.AsyncInterfaces

RaminJolfaei:
upgrade Microsoft.Bcl.AsyncInterfaces to latest version 8.0.0

### DIFF
--- a/src/Playwright/Playwright.csproj
+++ b/src/Playwright/Playwright.csproj
@@ -33,7 +33,7 @@
     <PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
     <PackageReference Include="System.Text.Json" Version="6.0.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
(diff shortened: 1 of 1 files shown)


campersau:
Is there a specific reason for this?
In my opinion libraries should use the lowest version possible since the users can override them by explicitly include the references themselves. This was also done with `System.Text.Json` here https://github.com/microsoft/playwright-dotnet/pull/2720
Also since `System.Text.Json` references `Microsoft.Bcl.AsyncInterfaces` as well, these should be kept in sync.

Meir017:
I think as long as our dependencies support the latest netstandard we should be ok updating, see https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/8.0.0#supportedframeworks-body-tab
image

RaminJolfaei:
My reason for upgrading Microsoft.Bcl.AsyncInterfaces to version 8.0.0 is that our DevOps pipeline throws an error during the post-compile step. The pipeline checks DLL references and fails when it doesn’t find version 8 in the bin folder. To resolve this issue and ensure the pipeline can complete successfully, the upgrade to version 8 is necessary.
Most of the other packages we use are now dependent on version 8.0.0 of Microsoft.Bcl.AsyncInterfaces.

mxschmitt:
Closing as per above - please start with filing an issue with reproduction steps etc. Thank you!

PR #3027, opened 2024-10-18
test: fix NUnit Console.Error and friends to appear in test output

mxschmitt:


src/Playwright.Tests.TestServer/SimpleServer.cs
```diff
@@ -89,11 +93,21 @@ public SimpleServer(int port, string contentRoot, bool isHttps)
         _contentRoot = contentRoot;
 
         _webHost = new WebHostBuilder()
+         .ConfigureLogging(logging =>
```
kblok, line 96: I would add one indentation here.

dgozman (approved):

PR #3029, opened 2024-10-21
chore(roll): roll Playwright to v1.48.0

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2996

### DIFF
--- a/src/Playwright/Core/Tracing.cs
+++ b/src/Playwright/Core/Tracing.cs
@@ -48,26 +48,21 @@ internal class Tracing : ChannelOwner, ITracing
         TracingStartOptions options = default)
     {
         _includeSources = options?.Sources == true;
-        var traceName = await _connection.WrapApiCallAsync(
-            async () =>
+        await SendMessageToServerAsync(
+        "tracingStart",
+        new Dictionary<string, object>
         {
-            await SendMessageToServerAsync(
-            "tracingStart",
-            new Dictionary<string, object>
-            {
-                ["name"] = options?.Name,
-                ["title"] = options?.Title,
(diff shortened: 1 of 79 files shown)


src/Playwright/Core/Tracing.cs
```diff
@@ -48,26 +48,21 @@ public async Task StartAsync(
         TracingStartOptions options = default)
     {
         _includeSources = options?.Sources == true;
-        var traceName = await _connection.WrapApiCallAsync(
-            async () =>
+        await SendMessageToServerAsync(
```
dgozman, line 51: Missing `MarkAsInternalType()` in this file?

src/Playwright/Core/WebSocketRoute.cs
```diff
@@ -0,0 +1,236 @@
...
+
+    internal override void OnMessage(string method, JsonElement? serverParams)
+    {
+        switch (method)
+        {
+            case "messageFromPage":
+                if (_onPageMessage != null)
+                {
+                    var frame = new WebSocketFrame(serverParams.Value.GetProperty("message").GetString(), serverParams.Value.GetProperty("isBase64").GetBoolean());
+                    _onPageMessage(frame);
```
dgozman, line 65: It seems like all the snippets for java/csharp should be updated to use `WebSocketFrame` instead of a string?

dgozman (approved):

src/Playwright/Core/WebSocketRoute.cs
mxschmitt: Filed as https://github.com/microsoft/playwright/issues/33199.

PR #3044, opened 2024-11-08
feat: add xUnit testing harness

mxschmitt:
First stab for adding xUnit testing harness for Playwright since they [changed their parallelism algorithm](https://github.com/microsoft/playwright-dotnet/issues/2977#issue-2451177306) which unblocked us. So far I found the following questions along the way:

- `Microsoft.Playwright.xUnit` vs `Microsoft.Playwright.XUnit`
- Is it normal that `Console.WriteLine` doesn't work in the test method? I had to follow [this](https://xunit.net/docs/capturing-output#:~:text=new%20methods%20instead.-,Capturing%20output%20in%20unit%20tests,-Unit%20tests%20have) guide instead to make it work. Looks like we can use https://github.com/xunit/xunit/issues/1730#issuecomment-2330825763 once its released. 
- Do we stick with `AppDomain.CurrentDomain.FirstChanceException` to determine if a test failed and restart the worker services (mostly Browser) appropriately? Inspired by @SimonCropp from [here](https://github.com/SimonCropp/XunitContext/blob/main/src/XunitContext/Context.cs#L46). We need some hack since there is no TestContext in xUnit.
  - Looks like if we target only xUnit.v3 then we could use the TestContext as per https://github.com/microsoft/playwright-dotnet/pull/3044#issuecomment-2465733703.
- Is targeting `net8.0` fine? `AppDomain` is only available in `net5.0`+. xUnit itself targets `net472;net6.0`. Maybe we should target both as well and provide `AppDomain` support only for the `net6.0` output.

Curious about your inputs @kblok @campersau @nohwnd thank you in advance!
 
Relates https://github.com/microsoft/playwright-dotnet/issues/2977


campersau:
> Is it normal that Console.WriteLine doesn't work in the test method?

Yes.

> Do we stick with AppDomain.CurrentDomain.FirstChanceException to determine if a test failed and restart the worker services appropriately?

If you wait a little xunit.v3 will be released and it contains new features like a [`TestContext`](https://github.com/xunit/xunit/blob/bb04d8f4a62c404d05ae6ec1233ff02e2658633a/src/xunit.v3.core/TestContext.cs#L60) and [`CaptureConsoleAttribute`](https://github.com/xunit/xunit/blob/bb04d8f4a62c404d05ae6ec1233ff02e2658633a/src/xunit.v3.core/CaptureConsoleAttribute.cs#L15) which might help.
See
- https://github.com/xunit/xunit/issues/2133
- https://github.com/xunit/xunit/issues/3031#issuecomment-2460668967

src/Playwright.xUnit/BrowserTest.cs
```diff
@@ -0,0 +1,62 @@
...
+    public override async Task InitializeAsync()
+    {
+        await base.InitializeAsync().ConfigureAwait(false);
+        var service = await BrowserService.Register(this, BrowserType).ConfigureAwait(false);
+        Browser = service.Browser;
+    }
+
+    public override async Task DisposeAsync()
+    {
+        await base.InitializeAsync().ConfigureAwait(false);
```
campersau, line 51:
```suggestion
        await base.DisposeAsync().ConfigureAwait(false);
```

src/Playwright/Core/AssertionsBase.cs
```diff
@@ -32,8 +32,9 @@
...
+[assembly: InternalsVisibleTo("Microsoft.Playwright.MSTest, PublicKey=0024000004800000940000000602000000240000525341310004000001000100059a04ca5ca77c9b4eb2addd1afe3f8464b20ee6aefe73b8c23c0e6ca278d1a378b33382e7e18d4aa8300dd22d81f146e528d88368f73a288e5b8157da9710fe6f9fa9911fb786193f983408c5ebae0b1ba5d1d00111af2816f5db55871db03d7536f4a7a6c5152d630c1e1886b1a0fb68ba5e7f64a7f24ac372090889be2ffb")]
 [assembly: InternalsVisibleToAttribute("Microsoft.Playwright.NUnit, PublicKey=0024000004800000940000000602000000240000525341310004000001000100059a04ca5ca77c9b4eb2addd1afe3f8464b20ee6aefe73b8c23c0e6ca278d1a378b33382e7e18d4aa8300dd22d81f146e528d88368f73a288e5b8157da9710fe6f9fa9911fb786193f983408c5ebae0b1ba5d1d00111af2816f5db55871db03d7536f4a7a6c5152d630c1e1886b1a0fb68ba5e7f64a7f24ac372090889be2ffb")]
```
campersau, line 36:
```suggestion
[assembly: InternalsVisibleTo("Microsoft.Playwright.NUnit, PublicKey=0024000004800000940000000602000000240000525341310004000001000100059a04ca5ca77c9b4eb2addd1afe3f8464b20ee6aefe73b8c23c0e6ca278d1a378b33382e7e18d4aa8300dd22d81f146e528d88368f73a288e5b8157da9710fe6f9fa9911fb786193f983408c5ebae0b1ba5d1d00111af2816f5db55871db03d7536f4a7a6c5152d630c1e1886b1a0fb68ba5e7f64a7f24ac372090889be2ffb")]
```

kblok:
> * `Microsoft.Playwright.xUnit` vs `Microsoft.Playwright.XUnit`

Maybe we could follow their name `xunit`, `Microsoft.Playwright.xunit`. Not that pretty. `Microsoft.Playwright.XUnit` would be my second choice.

> * Is it normal that `Console.WriteLine` doesn't work in the test method? I had to follow [this](https://xunit.net/docs/capturing-output#:~:text=new%20methods%20instead.-,Capturing%20output%20in%20unit%20tests,-Unit%20tests%20have) guide instead to make it work. Looks like we can use [Console / Trace / Debugger capture support xunit/xunit#1730 (comment)](https://github.com/xunit/xunit/issues/1730#issuecomment-2330825763) once its released.

Yes. That was a headache. We were using `XunitLoggerProvider` in the [old days](https://github.com/microsoft/playwright-dotnet/blob/df71ab7fcce12f185710368111def8ba8e45d7b2/src/Playwright.Tests/BaseTests/PlaywrightSharpBaseTest.cs#L20).

bradwilson:
In theory, the only capitalization "rules" are:

* Product name = xUnit.net
* Class name or namespace name = `Xunit`

In practice, people do whatever they want. 😁

The product name being "xUnit.net" is really because back in the day, "xUnit" meant "replace 'x' with something that represents your programming environment". SUnit = Smalltalk, JUnit = Java, NUnit = .NET. Since NUnit was already taken, we opted for "xUnit.net" as the next best name. (And by "we" I mean @jnewkirk who also was part of creating NUnit. 😉)

File names or package names, we don't care about. We use all lowercase for ours, but we've never asked anybody to follow that pattern. Frankly it's like that mostly because I'm an Extra Oldster™️ who likes lowercase file names because of case sensitive file systems.

bradwilson:
I will say, `XUnit` looks wrong to my eyes, but that's a bit of personal preference. I think for package names `Xunit` would be better than `XUnit`; or if you feel strongly about capitalizing the U, then use `xUnit`. We never capitalize the X and U together.

nohwnd:
I don’t feel strongly about the name, but I do prefer I prefer Xunit over xUnit in the package name.

src/Playwright.Xunit/BrowserService.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        var accessToken = Environment.GetEnvironmentVariable("PLAYWRIGHT_SERVICE_ACCESS_TOKEN");
+        var serviceUrl = Environment.GetEnvironmentVariable("PLAYWRIGHT_SERVICE_URL");
+
+        if (string.IsNullOrEmpty(accessToken) || string.IsNullOrEmpty(serviceUrl))
+        {
+            return await browserType.LaunchAsync(PlaywrightSettingsProvider.LaunchOptions).ConfigureAwait(false);
+        }
+
+        var exposeNetwork = Environment.GetEnvironmentVariable("PLAYWRIGHT_SERVICE_EXPOSE_NETWORK") ?? "<loopback>";
```
yury-s, line 59: Do we need all this service code right away? If so, can it at least be shared with other runners?

yury-s (approved):

PR #3056, opened 2024-11-19
chore(roll): roll Playwright to v1.49.0

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/3014

### DIFF
--- a/src/Playwright/Helpers/URLMatch.cs
+++ b/src/Playwright/Helpers/URLMatch.cs
@@ -1,0 +1,82 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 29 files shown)


src/Playwright/Helpers/URLMatch.cs
```diff
@@ -0,0 +1,82 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Playwright.Helpers;
+
+public class URLMatch
```
dgozman, line 30: Are we going to reuse this class for `Route` in addition to `WebSocketRoute`?

mxschmitt: Yes, I was planning to do this in a follow-up.

PR #3063, opened 2024-11-22
fix(xunit): support netstandard2.0

mxschmitt:
`ModuleInitializer` was something which was not supported in older .NET frameworks. This patch starts listening on the constructor instead and resets the TestOK status in TestInitialize.

Background: xUnit 2 does not provide a reliable way of determining if a test was passing or failing and reading it's status in the AfterEach. xUnit 3 allows to read its status which we'll use once its published. We want to support xUnit 2 as well.

https://github.com/microsoft/playwright-dotnet/issues/2977

### DIFF
--- a/src/Playwright.Xunit/WorkerAwareTest.cs
+++ b/src/Playwright.Xunit/WorkerAwareTest.cs
@@ -73,7 +73,6 @@ public class WorkerAwareTest : ExceptionCapturer
 
     public async override Task DisposeAsync()
     {
-        await base.DisposeAsync().ConfigureAwait(false);
         if (TestOk)
         {
             foreach (var kv in _currentWorker.Services)
@@ -90,6 +89,7 @@ public class WorkerAwareTest : ExceptionCapturer
             }
             _currentWorker.Services.Clear();
         }
+        await base.DisposeAsync().ConfigureAwait(false);
     }
 }
 
--- a/src/Playwright.Xunit/WorkerAwareTest.cs
+++ b/src/Playwright.Xunit/WorkerAwareTest.cs
@@ -25,7 +25,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Playwright.Core;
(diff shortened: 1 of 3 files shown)


src/Playwright.Xunit/WorkerAwareTest.cs
```diff
@@ -34,10 +34,10 @@
 
 namespace Microsoft.Playwright.Xunit;
 
-public class WorkerAwareTest : ExceptionCapturer, IAsyncLifetime
+public class WorkerAwareTest : ExceptionCapturer
 {
     private static readonly ConcurrentStack<Worker> _allWorkers = new();
-    private Worker? _currentWorker = null!;
+    private Worker _currentWorker = null!;
```
mxschmitt, line 40: drive-by: Similar to https://github.com/microsoft/playwright-dotnet/commit/7f3b8884efc0f288f504d90a677afbd3187f4fbe.

kblok:
How about targeting netstandard instead?

campersau:
`xunit.runner.visualstudio` and `Microsoft.NET.Test.Sdk` only work with `net462`.
Is there a reason why these packages are referenced?

mxschmitt:
> How about targeting netstandard instead?

I didn't do it because of:

> Playwright.Xunit.csproj : warning NU1701: Package 'xunit.runner.visualstudio 2.8.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' instead of the project target framework '.NETStandard,Version=v2.0'. This package may not be fully compatible with your project.

Since removing https://github.com/microsoft/playwright-dotnet/pull/3063#issuecomment-2493613530 sounds good to me we can target `netstandard2.0` now.

> `xunit.runner.visualstudio` and `Microsoft.NET.Test.Sdk` only work with `net462`. Is there a reason why these packages are referenced?

Addressed.

Thanks both!

src/Playwright.Xunit/WorkerAwareTest.cs
```diff
@@ -69,11 +69,11 @@ public virtual Task InitializeAsync()
         {
             AssertionsBase.SetDefaultTimeout(PlaywrightSettingsProvider.ExpectTimeout.Value);
         }
-        return Task.CompletedTask;
     }
 
-    public virtual async Task DisposeAsync()
+    public async override Task DisposeAsync()
     {
+        await base.DisposeAsync().ConfigureAwait(false);
```
dgozman, line 76: Base dispose is usually called at the end.

PR #3068, opened 2024-11-25
chore: ignore GitHub Codespaces BROWSER env var

mxschmitt:
e.g. `/vscode/bin/linux-x64/f1a4fb101478ce6ec82fe9627c43efbf9e98c813/bin/helpers/browser.sh` which prevents customers from running the getting started on GitHub Codespaces.

We respect the `BROWSER` env var for legacy reasons.

Fixes https://github.com/microsoft/playwright-dotnet/issues/2847

### DIFF
--- a/src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
+++ b/src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
@@ -41,7 +41,8 @@ public class PlaywrightSettingsProvider : ISettingsProvider
         {
             var browserFromEnv = Environment.GetEnvironmentVariable("BROWSER")?.ToLowerInvariant();
             // GitHub Codespaces sets the BROWSER environment variable, ignore it if its bogus.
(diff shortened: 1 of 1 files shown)


src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs
```diff
@@ -40,7 +40,8 @@ public static string BrowserName
         get
         {
             var browserFromEnv = Environment.GetEnvironmentVariable("BROWSER")?.ToLowerInvariant();
-            if (!string.IsNullOrEmpty(browserFromEnv))
+            // GitHub Codespaces sets the BROWSER environment variable, ignore it if its bogus.
+            if (!string.IsNullOrEmpty(browserFromEnv) && !browserFromEnv!.StartsWith("/vscode/"))
```
dgozman, line 44: Ouch! I don't like hardcoding this.

Perhaps we should change the error below into a warning and run `chromium`, if we consider `BROWSER` to be a legacy thing?

yury-s (approved):

PR #3077, opened 2024-12-04
devops: set up docker test and publish for NET 8 and 9

### LINKED ISSUE #3075: [Feature]: Support .NET 9 for docker
### 🚀 Feature Request

Currently the docker image is built with .NET 8, when will there be .NET 9 support?

### Example

_No response_

### Motivation

Makes it possible to do testing in CI after switching my projects to .NET 9.

timbze:
Here is a PR to do docker build testing and publishing for both NET 8 and 9. I have not tested this as it's not trivial to test without setting up docker hub, secrets, etc.

Fixes #3075

- [ ] 9.0-jammy docker image does not exist. I could just change the 2 .sh scripts to exit gracefully when trying to run for that configuration.

### DIFF
--- a/utils/docker/build.sh
+++ b/utils/docker/build.sh
@@ -3,7 +3,7 @@ set -e
 set +x
 
 if [[ ($1 == '--help') || ($1 == '-h') || ($1 == '') || ($2 == '') ]]; then
-  echo "usage: $(basename $0) {--arm64,--amd64} {jammy,noble} playwright:localbuild-noble"
+  echo "usage: $(basename $0) {--arm64,--amd64} {jammy,noble} {8.0,9.0} playwright:localbuild-noble"
   echo
   echo "Build Playwright docker image and tag it as 'playwright:localbuild-noble'."
   echo "Once image is built, you can run it with"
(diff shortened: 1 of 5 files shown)


timbze:

utils/docker/build.sh
```diff
@@ -36,6 +36,13 @@ else
   exit 1
 fi
 
+# Set default .NET version to net8 if not specified
```
timbze, line 39: Need to remove

timbze:
The jammy 8.0 test that failed (only 1) is a timeout and I guess not related to this PR

mxschmitt:
We recommend installing it manually on top of our existing image see the docs [here](https://playwright.dev/dotnet/docs/docker#using-a-different-net-version) since this is a STS version.

timbze:
@mxschmitt thank you, I didn't think of it working that simple. Should work for me then.

Kralizek:
> Thank you for your contribution! We recommend installing it manually on top of our existing image see the docs [here](https://playwright.dev/dotnet/docs/docker#using-a-different-net-version) since this is a STS version.

A release being STS or LTS shouldn't be a deciding factor imho.

Especially since Microsoft is trying to push the idea that STS and LTS are qualitative equivalent.

petrikero:
Just a heads up to others using the recommended approach, the proposed URL https://dot.net/v1/dotnet-install.sh has been quite unreliable and often has downtime. The direct GitHub URL https://raw.githubusercontent.com/dotnet/install-scripts/refs/heads/main/src/dotnet-install.sh is much more reliable.

Another downside of using the install script in your docker files is that it doesn't cache well in CI systems. For example, on GitHub, it takes around 8 seconds. An official .NET 9 image of Playwright would fix this.

Issue #3093, opened 2025-01-02
[Feature]: Playwright.Xunit should have a V3 variant


alexaka1:
### 🚀 Feature Request

With XUnit v3 released (https://xunit.net/releases/v3/1.0.0), Playwright should support it the same way it supports NUnit, MSTest etc.

### Motivation

It will allow Playwright to utilize the new `TestContext` of XUnit v3, such as `TestContext.Current.TestStatus` for enabling tracing on failed tests (currently not possible with xunit v2).

mxschmitt:
Linking https://github.com/xunit/xunit/issues/3146

omni-htg:
Should proposals be set up over a new PR, or continue over what #3097 has achieved so far? Thanks!

alexaka1:
Sorry guys, I kept meaning to post my workaround but completely forgot it. This is what I went with, works perfectly for my project:

<details>

```csharp
using System.Collections.Concurrent;
using System.Globalization;
using System.Text.Json;
using System.Text.Json.Serialization;
using Microsoft.Playwright.TestAdapter;
using Xunit;

// ReSharper disable once CheckNamespace
namespace Microsoft.Playwright.Xunit;

public class PageTest : ContextTest
{
    public IPage Page { get; private set; } = null!;

    public override async ValueTask InitializeAsync()
    {
        await base.InitializeAsync().ConfigureAwait(false);
        Page = await Context.NewPageAsync().ConfigureAwait(false);
    }
}

public class ContextTest : BrowserTest
{
    public IBrowserContext Context { get; private set; } = null!;

    public override async ValueTask InitializeAsync()
    {
        await base.InitializeAsync().ConfigureAwait(false);
        Context = await NewContext(ContextOptions()).ConfigureAwait(false);
    }

    public virtual BrowserNewContextOptions ContextOptions()
    {
        return new BrowserNewContextOptions
        {
            Locale = "en-US",
            ColorScheme = ColorScheme.Light,
        };
    }
}

public class BrowserTest : PlaywrightTest
{
    private readonly List<IBrowserContext> _contexts = new();
    public IBrowser Browser { get; internal set; } = null!;

    /// <summary>
    ///     Creates a new context and adds it to the list of contexts to be disposed.
    /// </summary>
    /// <param name="options"></param>
    /// <returns></returns>
    public async Task<IBrowserContext> NewContext(BrowserNewContextOptions? options = null)
    {
[...]

PR #3097, opened 2025-01-23
feat: add support for xUnit.v3

mxschmitt:
This PR adds support for xUnit.v3. The difference is that by default it uses MTP (Microsoft Testing Protocol) which means test discovery and execution are in two different processes. In order to workaround this we are serialising the settings as an environment variable which reaches the execution process later on. This seems like an acceptable workaround for now as per @nohwnd.

Context:
- https://xunit.net/docs/getting-started/v3/whats-new
- https://xunit.net/docs/getting-started/v3/migration

Fixes https://github.com/microsoft/playwright-dotnet/issues/3093

### DIFF
--- a/src/Playwright.TestingHarnessTest/tests/xunit.v3/basic.spec.ts
+++ b/src/Playwright.TestingHarnessTest/tests/xunit.v3/basic.spec.ts
@@ -247,7 +247,7 @@ test('should be able to parse BrowserName and LaunchOptions.Headless from runset
   expect(result.stdout).not.toContain("Headless")
 });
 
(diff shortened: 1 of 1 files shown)


jonathh21:
Can I plus one this with ever fibre of my being.

Having just looked over the wrap sheet for v3... I'd really like to be able to take screenshots and have them attached to the test results.

dgozman (approved):

connorb08:
> @mxschmitt I noticed the NuGet package for this hasn't been pushed yet.
> 
> Do you have an ETA of when that is taking place?

You should be able to use the pre-release `Microsoft.Playwright.Xunit.v3` package version `1.55.0-beta-4`

PR #3130, opened 2025-03-19
chore: enable Microsoft.Testing.Platform for internal tests

mxschmitt:
Interestingly `Command + C` does not work aka. no way of aborting the test run and there is no visible live output when running

    dotnet test -f net8.0 ./src/Playwright.Tests/Playwright.Tests.csproj --logger:"console;verbosity=detailed"

Looks related to `TestingPlatformDotnetTestSupport` when removing this and its using vstest it seems to work.


nohwnd:
at the moment you will have better time running the exe directly. (or `dotnet exec` on server, and probably `dotnet run` locally to ensure you re-build if there are changes).

image


Or you could update to net10  preview sdks where we are adding a new experience for dotnet test.

PR #3146, opened 2025-04-16
chore: make SimpleServer not return an empty response for 404

dgozman:
[Mirrors js TestServer](https://github.com/microsoft/playwright/blob/main/tests/config/testserver/index.ts#L281-L286).

### DIFF
--- a/src/Playwright.Tests/TestServer/SimpleServer.cs
+++ b/src/Playwright.Tests/TestServer/SimpleServer.cs
@@ -179,6 +179,8 @@ public class SimpleServer
         if (!File.Exists(filePath))
         {
             context.Response.StatusCode = StatusCodes.Status404NotFound;
+            context.Response.ContentType = "text/plain";
+            await context.Response.WriteAsync($"File not found: {filePath}").ConfigureAwait(false);
             return;
         }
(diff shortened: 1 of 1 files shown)


src/Playwright.Tests/TestServer/SimpleServer.cs
```diff
@@ -179,6 +179,11 @@ public async Task ServeFile(HttpContext context)
         if (!File.Exists(filePath))
         {
             context.Response.StatusCode = StatusCodes.Status404NotFound;
+            context.Response.ContentType = "text/plain";
+            var bytes = Encoding.UTF8.GetBytes($"File not found: {filePath}").AsMemory();
+            await context.Response.BodyWriter.WriteAsync(bytes).ConfigureAwait(false);
+            await context.Response.BodyWriter.CompleteAsync().ConfigureAwait(false);
+            context.Response.ContentLength = bytes.Length;
```
mxschmitt, line 186:
```suggestion
            context.Response.ContentType = "text/plain";
            await context.Response.WriteAsync($"File not found: {filePath}").ConfigureAwait(false);
            return;
```

Why does this not work?

mxschmitt (approved):

Issue #3163, opened 2025-05-15
[internal] Enable `<Nullable>enable</Nullable>` in `src/Playwright/Playwright.csproj`


mxschmitt:
### Please do not submit this issue.

- [x] I understand

PR #3165, opened 2025-05-21
chore: enable #nullable iteration 1

mxschmitt:


src/Playwright/Core/BrowserType.cs
```diff
@@ -256,21 +257,21 @@ async Task<IBrowser> CreateBrowserAsync()
         }
         var task = CreateBrowserAsync();
         var timeout = options?.Timeout != null ? (int)options.Timeout : 30_000;
-        return await task.WithTimeout(timeout, _ => throw new TimeoutException($"BrowserType.ConnectAsync: Timeout {options.Timeout}ms exceeded")).ConfigureAwait(false);
+        return await task.WithTimeout(timeout, _ => throw new TimeoutException($"BrowserType.ConnectAsync: Timeout {options!.Timeout}ms exceeded")).ConfigureAwait(false);
```
dgozman, line 260:
```suggestion
        return await task.WithTimeout(timeout, _ => throw new TimeoutException($"BrowserType.ConnectAsync: Timeout {timeout}ms exceeded")).ConfigureAwait(false);
```

src/Playwright/Helpers/URLMatch.cs
```diff
@@ -176,9 +177,9 @@ public static string GlobToRegexPattern(string glob)
         return string.Concat(tokens.ToArray());
     }
 
-    internal static string ToWebSocketBaseURL(string baseURL)
+    internal static string? ToWebSocketBaseURL(string? baseURL)
     {
-        if (string.IsNullOrEmpty(baseURL))
+        if (string.IsNullOrEmpty(baseURL) || baseURL == null)
```
dgozman, line 182: Shouldn't `IsNullOrEmpty()` handle the `null` case?

dgozman (approved):

dgozman (approved):

src/Playwright/Core/BrowserType.cs
```diff
@@ -43,19 +44,19 @@ internal BrowserType(ChannelOwner parent, string guid, BrowserTypeInitializer in
...
-    internal PlaywrightImpl Playwright { get; set; }
+    internal PlaywrightImpl Playwright { get; set; } = null!;
 
     public string ExecutablePath => _initializer.ExecutablePath;
 
     public string Name => _initializer.Name;
 
     [MethodImpl(MethodImplOptions.NoInlining)]
-    public async Task<IBrowser> LaunchAsync(BrowserTypeLaunchOptions options = default)
+    public async Task<IBrowser> LaunchAsync(BrowserTypeLaunchOptions? options = default)
```
yury-s, line 54: Will this break older version .net users? Or do we not care about that?

yury-s (approved):

src/Playwright/Core/BrowserType.cs
mxschmitt: This is not the public interface, the public interface is [this](https://github.com/microsoft/playwright-dotnet/blob/8c6b30bb769908065c09e54d8dd18d6a0b9a93d3/src/Playwright/API/Generated/IBrowserType.cs#L160) which is nullable.

PR #3167, opened 2025-05-22
chore: enable #nullable iteration 2

mxschmitt:


src/Playwright/Transport/ApiZone.cs
```diff
@@ -31,7 +31,7 @@ namespace Microsoft.Playwright.Transport;
 internal class ApiZone
 {
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string ApiName { get; set; }
+    public string? ApiName { get; set; }
 
-    public List<StackFrame> Frames { get; set; }
+    public List<StackFrame> Frames { get; set; } = null!;
```
mxschmitt, line 36: NOTE: `!null` in this context means the same as using the `required` keyword. We can't use the `required` keyword tho since its not available with our TFM we support (netstandard2.0).

src/Playwright/Helpers/ChannelHelpers.cs
```diff
@@ -26,8 +26,6 @@
 using System.Text.Json;
 using Microsoft.Playwright.Transport;
 
-#nullable enable
```
yury-s, line 29: Why did it go away?

src/Playwright/Transport/ApiZone.cs
yury-s: What is `TFM we support` ?

src/Playwright/Transport/Connection.cs
```diff
@@ -276,8 +272,7 @@ internal void Dispatch(PlaywrightServerMessage message)
         {
             if (message.Method == "__create__")
             {
-                var createObjectInfo = message.Params.Value.ToObject<CreateObjectInfo>(DefaultJsonSerializerOptions);
-                CreateRemoteObject(message.Guid, createObjectInfo.Type, createObjectInfo.Guid, createObjectInfo.Initializer);
+                CreateRemoteObject(message.Guid, message.Params.Value.GetProperty("type").ToObject<ChannelOwnerType>(), message.Params.GetString("guid", false)!, message.Params?.GetProperty("initializer"));
```
yury-s, line 275: Same, the previous code was arguably more type-safe.

src/Playwright/Transport/Protocol/InputFilesList.cs
```diff
@@ -21,14 +21,14 @@
...
+#nullable enable
 namespace Microsoft.Playwright.Transport.Protocol;
 
 public class InputFilesList
 {
-    public string Name { get; set; }
+    public string Name { get; set; } = null!;
 
-    public string MimeType { get; set; }
+    public string MimeType { get; set; } = null!;
```
yury-s, line 31: `null!` is a very indirect way to say that the field is required!

src/Playwright/Transport/Connection.cs
mxschmitt: Internally its similar and more explicit without going through the `JsonSerializer` infra + we don't need this extra type. I think for use-cases where we don't need to share a type this is preferred (like we do in other places as well).

src/Playwright/Transport/Connection.cs
```diff
@@ -186,21 +183,20 @@ private async Task<T> InnerSendMessageToServerAsync<T>(
...
-                    Id = id,
-                    Guid = @object == null ? string.Empty : @object.Guid,
-                    Method = method,
-                    Params = sanitizedArgs,
-                    Metadata = metadata,
-                },
-                keepNulls);
+            var message = new Dictionary<string, object>
+            {
+                ["id"] = id,
```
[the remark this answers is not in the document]
mxschmitt: Remove use-less wrapping type - this is similar to Python now.

src/Playwright/Transport/ChannelOwner.cs
```diff
@@ -39,13 +39,13 @@ internal class ChannelOwner
     internal bool _wasCollected;
     internal bool _isInternalType;
 
-    internal ChannelOwner(ChannelOwner parent, string guid) : this(parent, null, guid)
+    internal ChannelOwner(ChannelOwner parent, string guid) : this(parent, parent._connection, guid)
```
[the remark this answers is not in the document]
mxschmitt: not in this scenario. It can be null when using the other overload. I'll follow-up once I touch more places aka. Connection.cs which will automatically yield this.

src/Playwright/Transport/ApiZone.cs
mxschmitt: Thats our TFM: https://github.com/microsoft/playwright-dotnet/blob/70f3aa3c6d63ba886a8334f90887062b89070766/src/Playwright/Playwright.csproj#L10

Its supported only in net7+

src/Playwright/Helpers/ChannelHelpers.cs
mxschmitt: Already in line 24.

Issue #3161, opened 2025-05-13
[Bug]: CookiesAsync gives sometimes `Value cannot be null. (Parameter 'source')` - binary breaking changes from 1.51 -> 1.52


304NotModified:
### Version

1.52

### Steps to reproduce

1. Call `var cookies = await _page.Context.CookiesAsync();` 

Update: it's a binary breaking change, see https://github.com/microsoft/playwright-dotnet/issues/3161#issuecomment-2875780379: 
So compiled a lib to version <=1.51 and compile your project to >=1.52

### Expected behavior

- Don't crash if there are no cookies
- Do throw a ArgumentNullException, we don't send an argument

### Actual behavior

a `System.ArgumentNullException` is thrown

```
System.ArgumentNullException
  HResult=0x80004003
  Message=Value cannot be null. (Parameter 'source')
  Source=System.Linq
  StackTrace:
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.Playwright.Core.BrowserContext.<CookiesAsync>d__106.MoveNext() in /_/src/Playwright/Core/BrowserContext.cs:line 385
   at MyProject.End2EndTests.EndToEndSteps.<LogNavigation>d__19.MoveNext() in /_/src/End2EndTests/EndToEndSteps.cs:line 107

  This exception was originally thrown at this call stack:
    [External Code]
    Microsoft.Playwright.Core.BrowserContext.CookiesAsync(System.Collections.Generic.IEnumerable<string>) in BrowserContext.cs
    MyProject.End2EndTests.EndToEndSteps.LogNavigation(string) in EndToEndSteps.cs
```

FYI there is one page in the context, and it's about:blank

![Image](https://github.com/user-attachments/assets/bd0a8e63-ec03-4642-b2ae-67d3b2f967e9)

This is for a generic logging of the used url and used cookies. Unfortunately we cannot check if Context.Page.Count > 0, as there is a page. 

### Additional context

It seems this bug is introduced in 1.52

### Environment

```Text
- Operating System: Windows 11
- CPU: Intel Core i7
- Browser: Chromium
- .NET Version (TFM): net8.0
- Other info: on multiple machines local/CI
```

304NotModified:
Maybe it's a (binary) breaking change, as I see this:

[...]

PR #3169, opened 2025-05-23
Enabled Baseline Package Validation

304NotModified:
see https://github.com/microsoft/playwright-dotnet/issues/3161#issuecomment-2884342260

- Enabled Package baseline validation, see https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/baseline-version-validator
- Use 1.52.0 as baseline (as discussed below)
- Added CompatibilitySuppressions.xml for breaking changes sinds 1.52 (discussed below)
- CLI package is not in scope
- Added `CP0006 `- `A member was added to an interface without a default implementation`  (see see https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/diagnostic-ids) to `<NoWarn>` on Playwright.csproj, as that is the normal flow?
- Add a new workflow for checking the package results. Another option is to add `<GeneratePackageOnBuild>True</GeneratePackageOnBuild>` to the csproj files
  -   "NuGet Package Smoke Test" is not testing all the packages and have a different goal
- Example of breakage: #3170


mxschmitt (commented):
It looks great but I'm asking myself if its worth using a very old version as a baseline. I'd instead use a more recent one and then continue the journey from there?

304NotModified:
As we do semver, we would like to know the breaking changes when someone upgrades from a old version? 

Some breaking changes are acceptable, and those are in the suppression file :)

304NotModified:
@mxschmitt 
So this is the new breaking change in master (introduced by https://github.com/microsoft/playwright-dotnet/commit/d0f27458b0d44109d65b1bffe617f8150ac95721)


First we had a public empty ctor, now that ctor is gone!

image


Is that intentional?

Anyway, shown by the baseline package validation :)

I've added it to the suppressions by `dotnet pack /p:ApiCompatGenerateSuppressionFile=true`

304NotModified:
So the baseline version should 1.52.0, correct?

mxschmitt:
> So the baseline version should 1.52.0, correct?

Yes

304NotModified:
OK working on it :)

mxschmitt (approved):
Awesome! Yeah the `Ref` property was intentionally removed.

304NotModified:
All green. Feel free to rename the PR before merge :)

kblok:
@304NotModified @mxschmitt are we semver? I think playwright broke compatibility many times.

304NotModified:
@kblok 

I hope this PR will help in getting the API more stable in terms of breaking changes.
Sometime there a good reasons to break compatibility, this is now more transparent and explicit. 

My report of the binary breaking change (
https://github.com/microsoft/playwright-dotnet/issues/3161) was quickly fixed, so thanks for that! I hope we have less need of these fixes in the future  :)

kblok:
@304NotModified The problem is that upstream playwright determines the API, and`playwright-dotnet` matches the upstream version. If upstream breaks the compatibility without changing the major version, there is nothing we can do here.

PR #3172, opened 2025-05-24
chore: enable #nullable iteration 3

mxschmitt:


src/Playwright/Helpers/TaskHelper.cs
```diff
@@ -133,15 +133,10 @@ public static Task<T> WithTimeout<T>(this Task<T> task, Action timeoutAction, in
...
-        if (task == null)
-        {
-            return default;
-        }
-
         if (await TimeoutTask(task, timeout).ConfigureAwait(false) && timeoutAction != null)
         {
             timeoutAction();
-            return default;
+            return default!;
```
yury-s, line 139: Maybe leave a comment that this line is unreachable? Or better yet `throw timeoutAction();` or something like that, so that you don't have to add unreachable return statement.

yury-s (approved):

PR #3181, opened 2025-05-29
fix: make sure Playwright works in DevContainers

### LINKED ISSUE #3174: [Bug]: Throws Invalid browser name from 'BROWSER' environment variable inside VS Code devcontainer
### Version

1.52.0

### Steps to reproduce

1. Clone my repo at https://github.com/CurlyFire/playwrightbug
2. Make sure you have all the requirements for running Dev containers (ex: Docker Desktop)
3. Open Visual Studio Code with Dev Containers extension installed
4. Select from the menu View/Command Palette... :"Dev Containers: Open Folder in Container..."
5. Select the local folder where the repo was cloned and press Open
6. **Let the "Configuring Dev Container" step finish** (It is installing the Playwright dependencies and Playwright browsers.  See the postCreateCommand in .devcontainer/devcontainer.json for more details)
7. Select from the menu Terminal/New Terminal
8. From the newly opened Terminal, run the command :"dotnet test"

### Expected behavior

I expected the test to pass

### Actual behavior

bugplaywright test failed with 1 error(s) (5.8s)
    /_/src/Playwright.TestAdapter/PlaywrightSettingsProvider.cs(99): error TESTERROR: 
      bugplaywright.UnitTest.PlaywrightShouldNotThrow (1ms): Error Message: System.ArgumentException : Invalid browser name from 'BROWSER' environment variable.
      Supported browsers: 'chromium', 'firefox', and 'webkit'
      Actual browser: '/vscode/vscode-server/bin/linux-x64/848b80aeb52026648a8ff9f7c45a9b0a80641e2e/bin/helpers/browser.sh'
      Try to remove 'BROWSER' environment variable for using default browser

### Additional context

The unit test is located in UnitTest.cs and is empty, it only inherits from Microsoft.Playwright.Xunit.PageTest, which initializes the Playwright Page and throws.

This was almost fixed in PR https://github.com/microsoft/playwright-dotnet/pull/3068

However the condition that was added to ignore the BROWSER env variable is too narrow, it only targets GitHub CodeSpaces, and not a standalone DevContainer.

### Environment

```Text
- Operating System: [Linux DevContainer mcr.microsoft.com/devcontainers/dotnet:1-9.0-bookworm]
- CPU: [x64]
- Browser: [All, Chromium, Firefox, WebKit]
- .NET Version (TFM): [net9.0]
```

CurlyFire:
This patch fixes Invalid browser name exception when Playwright is used in a DevContainer context

Fixes #3174


CurlyFire:
@mxschmitt I completed the modification after understand how all the tests and checks work.  However, when I commented out the thrown exception in the ValidateBrowserName method, not a single test that I ran locally failed.  I thought that there would be a test that checked this behavior, so that I could modify it.

I would add one, but I'm not 100% certain what the context of the test should be or if I should even add one for this.

mxschmitt:
We don't have any tests for that - I think we can land as it is?

CurlyFire:
@mxschmitt  Yep, sounds good

Issue #3182, opened 2025-05-30
[Bug]: SuggestedFilename is "download" when it contains umlaute when running in docker container


campersau:
### Version

1.52

### Steps to reproduce

Example steps:
1. Clone my repo at https://github.com/campersau/playwrightdockerdownloadfilenameumlautissue
2. `docker run -it --rm -w /app --mount "type=bind,src=$(pwd),target=/app" mcr.microsoft.com/playwright/dotnet:v1.52.0`
3. `dotnet run --project WebApplication5.csproj`
4. The "Download SuggestedFileName" output should contain umlaute like `testäöü.txt` but it is just `download`

### Expected behavior

I expect to see SuggestedFileName = `testäöü.txt`

### Actual behavior

I see SuggestedFileName = `download`

### Environment

```Text
- Operating System: Windows
- CPU: x64
- Browser: Chromium
- .NET Version (TFM): net8.0
```

campersau:
Fixed by setting:
```
ENV LANG C.UTF-8
```

Could also be passed into docker run like this `--env LANG=C.UTF-8` so there is no need for a Dockerfile.

PR #3183, opened 2025-05-30
fix(docker): set default shell encoding

campersau:
Same as in https://github.com/microsoft/playwright/pull/30288

Prevents issues like https://github.com/microsoft/playwright-dotnet/issues/3182


campersau:
Looks like the same is missing in Java/ Python, it was only done in JS.

PR #3188, opened 2025-06-06
test: refactor TracingTests to use actual Trace Viewer

mxschmitt:
This gives us more confidence in the Trace Viewer tests and should in the future allow us to port more tests over to .NET.

### DIFF
--- a/src/Playwright.Tests/TracingTests.cs
+++ b/src/Playwright.Tests/TracingTests.cs
@@ -23,11 +23,9 @@
  */
 
 using System.Diagnostics;
-using System.IO.Compression;
-using System.Text;
-using System.Text.Json;
 using System.Text.RegularExpressions;
 using Microsoft.Playwright.Helpers;
+using Microsoft.Playwright.Tests.TestServer;
 
 namespace Microsoft.Playwright.Tests;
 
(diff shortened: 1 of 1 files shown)


src/Playwright.Tests/TracingTests.cs
```diff
@@ -353,81 +356,80 @@ public async Task ShouldShowTracingGroupInActionList()
...
     }
 
-    private static (IReadOnlyList<TraceEventEntry> Events, Dictionary<string, byte[]> Resources) ParseTrace(string path)
+    private async Task ShowTraceViewerAsync(string path, Func<TraceViewerPage, Task> callback)
     {
-        Dictionary<string, byte[]> resources = new();
-        using var archive = ZipFile.OpenRead(path);
-        foreach (var entry in archive.Entries)
+        var (executablePath, _) = Driver.GetExecutablePath();
+        var traceViewerPath = Path.GetFullPath(Path.Join(Path.GetDirectoryName(executablePath), "../../package/lib/vite/traceViewer/"));
```
campersau, line 377: Nit

```suggestion
        var traceViewerPath = Path.GetFullPath(Path.Join(Path.GetDirectoryName(executablePath), "..", "..", "package", "lib", "vite", "traceViewer"));
```

dgozman (approved):

src/Playwright.Tests/TracingTests.cs
mxschmitt: I think Path.Join allows slashes.

dgozman (approved):

PR #3210, opened 2025-07-23
feat(mstest): Mark TestContext as virtual

### LINKED ISSUE #3209: [Feature]: MSTest `WorkerAwareTest.TestContext` should be virtual
### 🚀 Feature Request

In the MSTest harness, `WorkerAwareTest` has a `TestContext` property, see https://github.com/microsoft/playwright-dotnet/blob/a878a633f888f3c920d4f057a9afe12368c036cf/src/Playwright.MSTest/WorkerAwareTest.cs#L13. This property should be marked `virtual` so that it can be overridden by inheritors as necessary.

### Example

_No response_

### Motivation

[Verify](https://github.com/VerifyTests/Verify) relies on hooking into `TestContext` to provide snapshot testing. It does this with a source generator to override the TestContext property and storing the property in its own `AsyncLocal` during set.

However, because WorkerAwareTest.TestContext isn't marked virtual, the source generator isn't able to add the setter hook. See https://github.com/VerifyTests/Verify/issues/1482 for the bug on the Verify side as well.

MattKotsenas:
`WorkerAwareTest` is designed to be inherited for testing. As such, the TestContext property, which is injected by the test framework, should be marked `virtual` so that inheritors can override behavior as necessary.

Fixes #3209.

### DIFF
--- a/src/Playwright.MSTest/WorkerAwareTest.cs
+++ b/src/Playwright.MSTest/WorkerAwareTest.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Playwright.MSTest;
 
 public class WorkerAwareTest
 {
-    public TestContext TestContext { get; set; } = null!;
+    public virtual TestContext TestContext { get; set; } = null!;
 
     private static readonly ConcurrentStack<Worker> _allWorkers = new();
     private Worker _currentWorker = null!;


mxschmitt:
2-3 weeks I assume.

PR #3214, opened 2025-07-30
test: do not depend on example.com in integration tests

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/actions/runs/16623930634/job/47035091929?pr=3213#step:9:54

Saw it a couple of more times in the past.


Skn0tt (approved):
we're testing MSTest with Playwright Test? 🤯

PR #3232, opened 2025-08-18
chore: roll to 1.55.0

Skn0tt:


mxschmitt (approved):

Skn0tt:
Mirroring the offline discussion here: the extension tests depend on service workers, but that feature isn't available in Java and Dotnet. So we don't test it.

PR #3234, opened 2025-08-27
chore: set v1.55.0

mxschmitt:


src/Common/Version.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <AssemblyVersion>1.55.0</AssemblyVersion>
-    <PackageVersion>$(AssemblyVersion)-beta-4</PackageVersion>
+    <PackageVersion>$(AssemblyVersion)</PackageVersion>
     <DriverVersion>1.55.0-beta-1756314050000</DriverVersion>
```
rajsite, line 5: @mxschmitt is it intentional that that the 1.55.0 release is using a beta DriverVersion? We try and align our playwright node and dotnet versions to the DriverVersion so we aren't mixing browsers used in different tests.

mxschmitt: Yes, this is intentional, the internal driver version can be a beta which means its still from the `release-*` branch.

PR #3248, opened 2025-10-24
feat: XUnit V3 - Add support for altering browser options using custom method

pomianowski:
I would like to conveniently change the options using an additional method, similar to `ConnectOptionsAsync`, e.g., `LaunchOptionsAsync`. In it, I can add my own code and additionally inject my own environment variables or settings, depending on the needs of my test.

#3249


pomianowski:
@mxschmitt, if you could take a moment to look at the idea of extending API with this PR, I would be very grateful

PR #3256, opened 2025-11-14
Run continuation asynchronously in context of events

nvborisenko:
This PR make Playwright events waiter more friendly with sync-over-async pattern.

### Issue

This piece of code demonstrates deadlock. Real code, you can run it (I ran it in NUnit).

```csharp
await page.RunAndWaitForResponseAsync(async () =>
{
    await page.GotoAsync("https://linked.in");
}, new Regex(".*google.*"));

//await Task.Yield();

Task.Run(async () => await page.Locator(".nav__logo-link").ClickAsync()).GetAwaiter().GetResult();
```

Here `GetAwaiter().GetResult()` is blocked. We understand that it is anti-pattern, `sync-over-async`. To fix it we can add `await Task.Yield()`. But what if Playwright can be more friendly for dummy users like me?

### Solution?
This PR allows tasks continuations asynchronously. Now the code above just works.


Skn0tt:
Hi Nikolay, sorry for not taking a look earlier. I'll ask our colleague Dima for a review, but it'll take a little longer due to holidays.

Skn0tt (approved):
I did some digging on the ramifications of this options, and it seems like a safe change for us. The only negative implication is a slight overhead since the continuation now needs to wait for the scheduler, but that microsecond impact is neglegible given we're dealing with slow browsers.

My understanding of DotNet's async scheduling is limited, so please take another look @dgozman.

nvborisenko:
It is absolutely safe, moreover it is recommended for libraries to prevent deadlocks I demonstrated in the original issue.

dgozman (commented):
The fix looks good to me. Let's add a test though, to avoid regressing this behavior.

dgozman (approved):

nvborisenko:
Stabilized CI build. Actually original simple changes/improvements revealed hidden race conditions. Then I made some changes to "fix" it, but I am afraid some of these "improvements" might be unnecessary (I don't fully understand what I did).

I will try to identify whether my changes are actually necessary to clean up this PR at least. Thank you for your support.

Skn0tt (changes requested):
Thanks for looking into it! I'll mark this as "request changes" to prevent anybody from accidentally merging while we haven't re-reviewed it.

nvborisenko:
Up.

nvborisenko:
Just noticed version 1.60 rolled out but this fix didn't make it in. Is anyone available to take a look?

nvborisenko:
I thought playwright is opensource, what is opensource...Who else should be tagged to review?.. given that one guy approved changes.. and I spent a lot of time to understand your internal unit tests flakiness. This change resolves the topic entirely, and flakiness implicitly. Please review.

dgozman:
@nvborisenko Sorry for missing reviews here, this PR did fall through the cracks.

At first, I thought that the original change that passed `TaskCreationOptions.RunContinuationsAsynchronously` internally was fine. But now, I am not that sure.

> So the flow is evaluate -> prompt opened -> start awaiting. This is misuse, I would say bad. The correct flow should: start awaiting -> evaluate -> prompt opened.

This is (very unfortunately) differs from Playwright for any other language, which is why this pattern is very common in tests. In other languages, Playwright does not run on multiple threads concurrently, so there is no possible timing issue this such an example. While in theory this looks like a bad test, in practice this has never triggered a timing issue, unless you insert an artificial async hop in the middle (like a timeout in your example). I have a feeling that changes in the `Waiter.cs` might actually introduce this problem here.

Also, I'm not sure I understand the value of the new changes with the locks. Playwright is not designed for concurrency. We operate under assumption that all state updates happen synchronously upon receiving a protocol message from the driver. I don't think sprinkling some locks will improve the situation. In fact, looking at the changes in `src/Playwright/Core/Frame.cs`, I start thinking that the `RunContinuationsAsynchronously` could be the actual cause of problems here, because it introduces an async hop in the exact place we are touching there.

Overall, this PR is not being landed quickly because:
- there has been no issue filed;
- the (non-existent) issue did not collect upvotes, so we cannot even understand whether this is a common problem among the users;
- the fix may (or may not - it's not obvious from the first glance) introduce subtle regressions, perhaps similar to the change in `Frame.cs`.
In circumstances like these we prefer to not introduce changes without a clear reason and benefit.

I think the best way to proceed for us would be to start over with an issue.

Again, sorry for not being very transparent and proactive here, and definitely for being slow.

PR #3303, opened 2026-04-30
fix(screencast): correct overlay/chapter channels

### LINKED ISSUE #3302: [Bug]: Screencast chapter and overlay features have bad bindings and do not work
### Version

1.59.0

### Steps to reproduce

To reproduce in any project, make a call to `page.ScreenCast.ShowChapterAsync` or `page.ScreenCast.ShowOverlayAsync.

To reproduce

1. Clone my repo at https://github.com/luisfaceira/playwright-screencast-bindings-repro
2. dotnet run Program.cs
3. You should see the error come up

### Expected behavior

Calls should not throw an Exception and the respective feature should work, adding a chapter message or an overlay.

### Actual behavior

throws the following Exception:

```
Unhandled exception. Microsoft.Playwright.PlaywrightException: Unknown scheme for Params: Page.overlayChapter
   at Microsoft.Playwright.Transport.Connection.InnerSendMessageToServerAsync[T](ChannelOwner object, String method, Dictionary`2 dictionary, Boolean keepNulls) in /_/src/Playwright/Transport/Connection.cs:line 201
   at Microsoft.Playwright.Transport.Connection.WrapApiCallAsync[T](Func`1 action, Boolean isInternal, String title) in /_/src/Playwright/Transport/Connection.cs:line 505
```

### Additional context

I've traced this to be a misconfigured binding between these new C# commands and the corresponding playwright server API commands and I've verified the exact fix needed on my own fork, which I'll link next as suggestion for fixing this issue.

### Environment

```Text
- Operating System: Windows 11
- CPU: arm64
- Browser: Chromium
- .NET Version (TFM): net8.0

I've come across and reproduced in the above environment, though I'm convinced it will happen in any other C# environment using v1.59.0
```

luisfaceira:
- Fix screencast overlay/chapter/visibility APIs to call the `screencast*` channel methods expected by the driver.
- Add regression coverage for overlay and chapter APIs.

Fixes #3302

### DIFF
--- a/src/Playwright.Tests/ScreencastTests.cs
+++ b/src/Playwright.Tests/ScreencastTests.cs
@@ -281,4 +281,45 @@ public class ScreencastTests : BrowserTestEx
         }
         await context.CloseAsync();
     }
+
+    [PlaywrightTest()]
+    public async Task ShowOverlayAsync_ReturnsDisposableThatCompletesOnDispose()
+    {
+        var context = await Browser.NewContextAsync();
+        var page = await context.NewPageAsync();
+
+        await using (var overlay = await page.Screencast.ShowOverlayAsync("<div>overlay</div>"))
+        {
+            Assert.IsNotNull(overlay);
+        }
+
+        await context.CloseAsync();
+    }
+
+    [PlaywrightTest()]
+    public async Task ShowChapterAsync_CompletesWithTitleAndOptions()
+    {
+        var context = await Browser.NewContextAsync();
+        var page = await context.NewPageAsync();
+
+        await page.Screencast.ShowChapterAsync("Chapter title", new()
+        {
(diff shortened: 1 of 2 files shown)


luisfaceira:

dgozman (approved):
Thank you for the fix!

luisfaceira:
Happy to help! In fact, I'm eager to use the screencast features on lots of our projects and have already prepared a significant refactoring across multiple of our testing projects to generate artifacts... but I can't go through with it until the fix is published.
When can we expect a release published in NuGet with this fixed, @dgozman ?

I'm aware of multiple workarounds to getting non-officially-published versions into our project's dependencies, but I would very much prefer not have to deal with all the nasty consequences of those approaches.

Thanks!

PR #3322, opened 2026-06-11
devops(driver): assemble driver from npm package and Node.js builds

dgozman:
## Summary
- `build.sh --download-driver` now assembles `src/Playwright/.drivers` from the `playwright-core` npm package and the official Node.js builds, instead of downloading prebuilt driver archives from the CDN.
- `.drivers` now mirrors the shipped `.playwright` layout (one copy of `package/`, per-platform node binaries), simplifying packaging in `Playwright.csproj` to a single glob.
- The bundled Node.js version is pinned as `DriverNodeVersion` in `Version.props`; `build.sh --roll` keeps it in sync with upstream.
- `api.json` and `protocol.yml` are no longer shipped in the NuGet package; nothing in this repo uses them. The package contents are otherwise identical to the released 1.60.0.

### DIFF
--- a/src/tools/Playwright.Tooling/DriverDownloader.cs
+++ b/src/tools/Playwright.Tooling/DriverDownloader.cs
@@ -24,6 +24,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Formats.Tar;
 using System.IO;
 using System.IO.Compression;
 using System.Net.Http;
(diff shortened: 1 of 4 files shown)


src/tools/Playwright.Tooling/DriverDownloader.cs
```diff
@@ -34,156 +35,283 @@
...
     }
 
-    private async Task<bool> ExecuteAsync()
+    private async Task DownloadPlaywrightPackageAsync(HttpClient client, string driversDirectory)
     {
-        var destinationDirectory = new DirectoryInfo(Path.Combine(BasePath, "src", "Playwright", ".drivers"));
-        var dedupeDirectory = new DirectoryInfo(Path.Combine(BasePath, "src", "Playwright", ".playwright"));
-        string driverVersion = DriverVersion;
+        string url = $"https://registry.npmjs.org/playwright-core/-/playwright-core-{DriverVersion}.tgz";
+        await WithRetriesAsync(url, async () =>
```
Skn0tt, line 191: NPM needs retries?? wow.

src/tools/Playwright.Tooling/DriverDownloader.cs
```diff
@@ -34,156 +35,283 @@
...
+        string archiveDir = $"node-v{NodeVersion}-{nodeSuffix}";
+        string url = $"https://nodejs.org/dist/v{NodeVersion}/{archiveDir}.{(isWindows ? "zip" : "tar.gz")}";
+        string nodeEntry = isWindows ? $"{archiveDir}/node.exe" : $"{archiveDir}/bin/node";
+        string nodeDestination = Path.Combine(driversDirectory, "node", platformId, isWindows ? "node.exe" : "node");
+        // Every Node.js archive carries the same LICENSE; extract it from a single
+        // designated platform to avoid parallel writes to the same file.
+        string licenseDestination = platformId == "linux-x64" ? Path.Combine(driversDirectory, "node", "LICENSE") : null;
 
-        if (!versionFile.Exists)
+        await WithRetriesAsync(url, async () =>
```
Skn0tt, line 227: Same here, were these CDNs unreliable enough that you needed this locally?

Skn0tt (approved):

src/tools/Playwright.Tooling/DriverDownloader.cs
dgozman: Yes, npm needs retries.

src/tools/Playwright.Tooling/DriverDownloader.cs
dgozman: Didn't have any issues with downloading node locally.

PR #3338, opened 2026-07-30
chore(driver): roll to 1.62.0

### LINKED ISSUE #38919: [Bug]: locale is not propagated to workers in firefox
### Version

95a728a037070f31afb7decebbc4d9e2284e26ba

### Steps to reproduce

https://github.com/microsoft/playwright/blob/95a728a037070f31afb7decebbc4d9e2284e26ba/tests/library/browsercontext-locale.spec.ts#L166

This test passes in CI, but not locally. I assume it's because CI locale happens to give the same output as es-MX. If you change the locale to ru-RU, it should also fail in CI.

It seems that workers are not respecting the context locale.

### Expected behavior

Workers should be respecting context locale.

### Actual behavior

Workers are using system locale.

### Additional context

found during https://github.com/microsoft/playwright-python/pull/3026

### Environment

```shell
System:
    OS: macOS 26.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 133.67 MB / 18.00 GB
  Binaries:
    Node: 20.19.0 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/node
    Yarn: 1.22.22 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/yarn
    npm: 9.9.4 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/npm
    pnpm: 10.28.1 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/pnpm
    Deno: 2.5.2 - /opt/homebrew/bin/deno
  IDEs:
    VSCode: 1.108.1 - /opt/homebrew/bin/code
    Claude Code: 2.1.15 - /opt/homebrew/bin/claude
  Languages:
    Bash: 3.2.57 - /bin/bash
```

dcrousso:
Rolls Playwright for .NET to driver `1.62.0` with Chromium `151.0.7922.34`, Firefox `153.0`, WebKit `26.5`, and Node.js `24.18.1`.

### Protocol timeout migration

`1.62` removes per method `timeout` parameters and carries the call timeout in `Metadata.timeout`.  The .NET client still sent `timeout` as a parameter, so the `1.62.0` server ignored every server dispatched operation timeout.

`Connection.InnerSendMessageToServerAsync` now moves the resolved `timeout` parameter into the metadata before sending the message.  `waitForTimeout` remains unaffected because it uses `waitTimeout`.

The follow-up <https://github.com/microsoft/playwright-dotnet/pull/3339> passes the timeout as a separate `SendMessageToServerAsync` argument instead of extracting it from the parameter dictionary.

| | Failed | Passed | Duration |
| :--- | :---: | :---: | :---: |
| Before | 24 | 278 | 12m 15s |
| After | 0 | 302 | 1m 49s |

### Rolling tooling

Upstream removed `utils/build/build-playwright-driver.sh` in <https://github.com/microsoft/playwright/pull/41518>, so `roll_driver` could no longer read the Node.js version.  It now reads the latest Node.js LTS from <https://nodejs.org/dist/index.json>, matching `utils/build/update-playwright-node.mjs`.

### API and browser updates

- Adds the handwritten `IAPIResponse.Timing` implementation required by the generated interface
- Adds the handwritten `ILocator.WaitForFunctionAsync` implementation required by the generated interface
- Removes the Firefox specific worker locale expectation because Firefox `153` fixes <https://github.com/microsoft/playwright/issues/38919>
- Expects the certificate subject `true` from WebKit on Windows, matching the `1.62.0` upstream test
- Aligns `ShouldWorkWhileNavigating` with the upstream full page screenshot test

Upstream marks `timing` and `responseEndTiming` optional and emits them together, while the .NET channel generator emits `ResponseEndTiming` as a non nullable `float`.  `APIResponse.Timing` therefore uses `timing == null` to select the `-1` fallback for `ResponseEnd`.

The viewport version of `ShouldWorkWhileNavigating` passed with `1.61.1-beta` in <https://github.com/microsoft/playwright-dotnet/actions/runs/28052877970>, but waits out its timeout with `1.62.0`.  An explicit `Timeout = 1000` is honored after approximately `1000` milliseconds, so the test now matches the upstream full page scenario.

Validation:

- `dotnet build src/Playwright` succeeds with no warnings or errors
- `dotnet format ./src/ --verify-no-changes` succeeds
- The targeted WebKit security details tests pass
- The Chromium suite passes with `1818` tests and `33` skipped tests
- The only two local failures require `pwsh`, which is not installed on the validation machine

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -36,7 +36,12 @@ function download_driver() {
 function roll_driver() {
   new_driver_version="$1"
   upstream_package_version=$(node -e "console.log(require('${upstream_repo_path}/package.json').version)")
-  new_node_version=$(sed -n 's/^NODE_VERSION="\([^"]*\)".*/\1/p' "${upstream_repo_path}/utils/build/build-playwright-driver.sh")
+  new_node_version=$(curl -fsSL "https://nodejs.org/dist/index.json" \
+    | node -e "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>console.log(JSON.parse(s).find(r=>r.lts).version.slice(1)))")
+  if [[ -z "${new_node_version}" ]]; then
+    echo "Failed to determine the Node.js version for driver ${new_driver_version}"
+    exit 1
+  fi
   echo "Rolling .NET driver to driver ${new_driver_version} (Node.js ${new_node_version}) and upstream version ${upstream_package_version}..."
 
   xml_file_path="./src/Common/Version.props"
--- a/src/Playwright/Transport/Connection.cs
+++ b/src/Playwright/Transport/Connection.cs
@@ -181,6 +181,11 @@ internal class Connection : IDisposable
             ["internal"] = isInternal,
             ["wallTime"] = DateTimeOffset.Now.ToUnixTimeMilliseconds(),
         };
+        if (sanitizedArgs.TryGetValue("timeout", out var timeout))
+        {
+            sanitizedArgs.Remove("timeout");
+            metadata["timeout"] = timeout;
+        }
         if (!string.IsNullOrEmpty(title))
         {
             metadata["title"] = title;
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 
 |          | Linux | macOS | Windows |
 |   :---   | :---: | :---: | :---:   |
-| Chromium <!-- GEN:chromium-version -->149.0.7827.55<!-- GEN:stop --> | ✅ | ✅ | ✅ |
+| Chromium <!-- GEN:chromium-version -->151.0.7922.34<!-- GEN:stop --> | ✅ | ✅ | ✅ |
 | WebKit <!-- GEN:webkit-version -->26.5<!-- GEN:stop --> | ✅ | ✅ | ✅ |
-| Firefox <!-- GEN:firefox-version -->151.0<!-- GEN:stop --> | ✅ | ✅ | ✅ |
+| Firefox <!-- GEN:firefox-version -->153.0<!-- GEN:stop --> | ✅ | ✅ | ✅ |
 
 Playwright for .NET is the official language port of [Playwright](https://playwright.dev), the library to automate [Chromium](https://www.chromium.org/Home), [Firefox](https://www.mozilla.org/en-US/firefox/new/) and [WebKit](https://webkit.org/) with a single API. Playwright is built to enable cross-browser web automation that is **ever-green**, **capable**, **reliable** and **fast**.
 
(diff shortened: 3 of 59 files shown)


src/Playwright/Transport/Connection.cs
```diff
@@ -181,6 +181,13 @@ private async Task<T> InnerSendMessageToServerAsync<T>(
             ["internal"] = isInternal,
             ["wallTime"] = DateTimeOffset.Now.ToUnixTimeMilliseconds(),
         };
+
+        // The protocol carries the call timeout in the metadata instead of as a parameter.
```
dgozman: Useless comment detected!

src/Playwright/Transport/Connection.cs
```diff
@@ -181,6 +181,13 @@ private async Task<T> InnerSendMessageToServerAsync<T>(
...
+        if (sanitizedArgs.TryGetValue("timeout", out var timeout))
```
dgozman, line 184: I think we should follow up and pass timeout as a separate argument of `SendMessageToServerAsync`, similar to js and java clients.

build.sh
```diff
@@ -36,7 +36,22 @@ function download_driver() {
 function roll_driver() {
   new_driver_version="$1"
   upstream_package_version=$(node -e "console.log(require('${upstream_repo_path}/package.json').version)")
-  new_node_version=$(sed -n 's/^NODE_VERSION="\([^"]*\)".*/\1/p' "${upstream_repo_path}/utils/build/build-playwright-driver.sh")
+  # The Node.js version used to be pinned in the upstream driver build script. That script was
```
dgozman: Another useless comment detected!

build.sh
```diff
@@ -36,7 +36,22 @@ function download_driver() {
...
+  # removed in microsoft/playwright#41518, so for newer versions follow the policy it had, which
+  # upstream utils/build/update-playwright-node.mjs still uses: the latest Node.js LTS.
+  legacy_node_version_script="${upstream_repo_path}/utils/build/build-playwright-driver.sh"
```
dgozman: Let's drop the support for the "legacy script".

dgozman (approved):

src/Playwright/Transport/Connection.cs
dcrousso: <https://github.com/microsoft/playwright-dotnet/pull/3339>

PR #3344, opened 2026-08-11
fix(transport): store expect errorDetails as serializable string

### LINKED ISSUE #3342: [Bug]: .NET Framework: any failing Expect() assertion hangs forever and permanently kills the driver connection (ArgumentException: Argument passed is not serializable in Connection.Dispatch) — regression in 1.61
### Version

1.61.0

### Steps to reproduce

Minimal console repro, no test framework involved — 100% deterministic:

**pwrepro.csproj**
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net472</TargetFramework>
    <LangVersion>latest</LangVersion>
    <PlatformTarget>x64</PlatformTarget>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Playwright" Version="1.61.0" />
  </ItemGroup>
</Project>
```

**Program.cs**
```csharp
using Microsoft.Playwright;
using System;
using System.Diagnostics;
using System.Threading.Tasks;

class Program
{
    static async Task Main()
    {
        using var playwright = await Playwright.CreateAsync();
        await page.SetContentAsync("<html><body>empty</body></html>");

        var sw = Stopwatch.StartNew();
        // Any FAILING web-first assertion triggers the bug:
        var assertion = Assertions.Expect(page.Locator("#does-not-exist"))
            .ToHaveCountAsync(1, new LocatorAssertionsToHaveCountOptions { Timeout = 3000 });

        if (await Task.WhenAny(assertion, Task.Delay(15000)) != assertion)
        {
            Console.WriteLine($"HANG: assertion still pending after {sw.ElapsedMilliseconds} ms.");
            try { await page.TitleAsync(); }
            catch (Exception ex) { Console.WriteLine($"Connection is dead: {ex.GetType().FullName}: {ex.Message}"); }
            return;
        }
        try { await assertion; }
        catch (Exception ex) { Console.WriteLine($"OK: failed cleanly after {sw.ElapsedMilliseconds} ms with {ex.GetType().FullName}."); }
    }
}
```

1. `dotnet build -c Release` then run the exe (Windows, .NET Framework 4.7.2+).
2. With **1.61.0**, output is:
   ```
   HANG: assertion still pending after 15012 ms.
   Connection is dead: Microsoft.Playwright.TargetClosedException: The argument passed is not serializable. Parameter name: value
   ```
3. Change the package to **1.60.0**, rebuild, rerun:
   ```
   OK: failed cleanly after 3038 ms with Microsoft.Playwright.PlaywrightException.

### Expected behavior

The failing assertion throws `PlaywrightException` after ~3000 ms (as in 1.60.0), and the driver connection stays usable.

### Actual behavior

On .NET Framework, the assertion task **never completes** (hangs until the process dies / an external test-framework timeout aborts it), and the shared driver connection is **permanently closed**: every subsequent Playwright call throws `TargetClosedException: The argument passed is not serializable. Parameter name: value`. In an NUnit/MSTest suite (where `PlaywrightTest` shares one static `IPlaywright` per process), a single ordinary assertion failure therefore takes down the entire remaining run in a cascade of `TargetClosedException`s.

First-chance capture at the throw site (`AppDomain.FirstChanceException` + `Environment.StackTrace`):

```
System.ArgumentException: The argument passed is not serializable. Parameter name: value
   at System.Collections.ListDictionaryInternal.set_Item(Object key, Object value)
   at Microsoft.Playwright.Transport.Connection.Dispatch(PlaywrightServerMessage message) in /_/src/Playwright/Transport/Connection.cs:line 283
   at Microsoft.Playwright.Playwright.<>c__DisplayClass0_0.<CreateAsync>b__0(Object _, Byte[] message) in /_/src/Playwright/Playwright.cs:line 49
   at Microsoft.Playwright.Transport.StdIOTransport.ProcessStream(CancellationToken token) in /_/src/Playwright/Transport/StdIOTransport.cs:line 246
   at Microsoft.Playwright.Transport.StdIOTransport.<GetResponseAsync>d__30.MoveNext() in /_/src/Playwright/Transport/StdIOTransport.cs:line 220
   ...
```

## Root cause analysis

Introduced in 1.61 (not present in 1.60). In `Connection.Dispatch`:

```csharp
_callbacks.TryRemove(message.Id.Value, out ConnectionCallback value);   // callback removed FIRST
...
Exception ex = ParseException(message.Error.Error, FormatCallLog(message.Log));
ex.Data["playwright.errorDetails"] = message.ErrorDetails;              // ← throws on .NET Framework
ex.Data["playwright.log"] = message.Log;
value.TaskCompletionSource.TrySetException(ex);                         // never reached
```

- `PlaywrightServerMessage.ErrorDetails` is a `JsonElement?`. `System.Text.Json.JsonElement` is **not** `[Serializable]`.
- On .NET Framework, `Exception.Data` is a `ListDictionaryInternal` whose indexer **requires keys and values to be serializable** (`FEATURE_SERIALIZATION` check, see [reference source](https://referencesource.microsoft.com/#mscorlib/system/collections/listdictionaryinternal.cs)) and throws `ArgumentException("Argument passed is not serializable", "value")` otherwise. .NET (Core) has no such check, which is why the bug only affects .NET Framework TFMs.
- `ErrorDetails` is only non-null for **failed `expect` responses**, so ordinary action/navigation timeouts still work — only failing web-first assertions hit the bug.
- Two independent consequences:
  1. the callback was already removed from `_callbacks` and `TrySetException` is never reached, so the awaiting assertion task is **orphaned forever** (it is not even completed by the later connection-close purge);
  2. the exception escapes into the transport read loop, which closes the connection with this `ArgumentException` as the close reason — hence the `TargetClosedException: The argument passed is not serializable` on every subsequent call.

## Suggested fix

Any of (ideally the first two):

- Store a serializable representation instead of the raw `JsonElement`, e.g. `ex.Data[ErrorDetailsDataKey] = message.ErrorDetails?.GetRawText()` (string), and parse it back where it is consumed (`Frame.ExpectAsync`).
- Harden `Dispatch` so enriching/completing a callback can never throw past the `TryRemove` (try/catch around the `Data` assignments, and guarantee `TrySetException`/`TrySetResult` is always called once the callback has been removed).

### Additional context

_No response_

### Environment

```Text
- OS: Windows 11 Pro (10.0.26200)
- .NET: target net472, runtime .NET Framework 4.8.x
- Playwright: Microsoft.Playwright 1.61.0 (repro), 1.60.0 (works); also reproduced through Microsoft.Playwright.NUnit 1.61.0 in a real NUnit 3 suite
- Browser: bundled Chromium
- Note: the exception message above is localized ("L'argument passé n'est pas sérialisable. Nom du paramètre : value" on a French OS)
```

kblok:
## Summary
- Fixes a .NET Framework regression where any failing web-first `Expect()` hung forever and permanently killed the shared driver connection.
- `Connection.Dispatch` now stores expect `errorDetails` as raw JSON text (serializable) instead of `JsonElement`, and always completes the callback even if exception enrichment fails.
- `Frame.ExpectAsync` parses that JSON string when building `FrameExpectResult`.

Fixes #3342

## Root cause
On .NET Framework, `Exception.Data` rejects non-serializable values. Storing `JsonElement` threw after the callback was already removed from `_callbacks`, so `TrySetException` never ran (orphaned hang) and the transport loop closed the connection.

## Test plan
- [x] Added `ConnectionTests.Dispatch_ExpectErrorDetails_MustBeSerializableForExceptionData` (portable unit test; fails on main when ErrorDetails is `JsonElement`)
- [x] Confirmed red on unfixed main, green after fix (`net8.0`)
- [x] Spot-checked `LocatorAssertionsTests.ShouldSupportToBeChecked` and `ShouldSupportToHaveCount`
- [ ] CI on Windows/.NET Framework (original hang scenario)


src/Playwright/Transport/Connection.cs
```diff
@@ -286,8 +286,18 @@ internal void Dispatch(PlaywrightServerMessage message)
...
+                // Store ErrorDetails as raw JSON text (not JsonElement) so expect
+                // failures cannot throw here, orphan the callback, and kill the connection.
+                try
+                {
+                    exception.Data[ErrorDetailsDataKey] = message.ErrorDetails?.GetRawText();
+                    exception.Data[LogDataKey] = message.Log;
+                }
+                catch
+                {
+                    // Best-effort enrichment only; completing the callback is mandatory.
```
dgozman: Let's not do this, or we won't be able to catch issues like this one. The earlier we know something went wrong, the better the chances for a fix.

src/Playwright.Tests/ConnectionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.Playwright.Transport;
+
+namespace Microsoft.Playwright.Tests;
+
+/// <summary>
+/// Regression coverage for expect errorDetails attached to Exception.Data.
+/// On .NET Framework, Exception.Data rejects non-serializable values (e.g. JsonElement),
+/// which used to orphan the callback and kill the driver connection (issue #3342).
+/// </summary>
+public class ConnectionTests
```
dgozman: Let's drop this test, I don't like that we mostly test the test code itself. Ideally, we would get some .net framework coverage, but I'm not it's worth it at this point.

dgozman (commented):
Hi Darío!

Thank you for the PR. I think the string serialization is the right solution, but I have some comments about the rest.

Thank you!

src/Playwright.Tests/ConnectionTests.cs
kblok: Makes sense!

src/Playwright/Transport/Connection.cs
```diff
@@ -286,7 +286,9 @@ internal void Dispatch(PlaywrightServerMessage message)
             if (message.Error != null && message.Result == null)
             {
                 var exception = ParseException(message.Error.Error, FormatCallLog(message.Log));
-                exception.Data[ErrorDetailsDataKey] = message.ErrorDetails;
+                // Exception.Data values must be serializable on .NET Framework.
+                // Store ErrorDetails as raw JSON text (not JsonElement).
```
kblok:
```suggestion
```

src/Playwright/Transport/Connection.cs
kblok: Done!

dgozman (approved):
Looks great, thank you!

PR #3346, opened 2026-08-19
feat(assertions): allow asserting JS undefined in ToHaveJSPropertyAsync

### LINKED ISSUE #3299: [Feature]: Represent ToHaveJSPropertyAsync for JS value undefined
### 🚀 Feature Request

Would like to be able to assert that an element has a property with JS value `undefined` using `ToHaveJSPropertyAsync`. It's not clear from documentation or tests in playwright-dotnet how that is done:

https://github.com/microsoft/playwright-dotnet/blob/9dd8feeae24be5be9aff9f53b835dd89790ddcb0/src/Playwright.Tests/Assertions/LocatorAssertionsTests.cs#L600-L618

Seems to be possible in the JS library based on tests:
https://github.com/microsoft/playwright/blob/dcd350445226ec3c54cc16e8cad1189c61bf07c8/tests/page/expect-misc.spec.ts#L193-L197

### Example

_No response_

### Motivation

Better align playwright-dotnet with capabilities of playwright JS

kblok:
## Summary
- Add `JSUndefined.Value` so `ToHaveJSPropertyAsync` can assert a JavaScript `undefined` property, matching the JS matcher.
- Serialization is isolated to this assertion; `EvaluateAsync` still sends C# `null` as JS `null`.

Fixes #3299

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Playwright.Tests/Assertions/LocatorAssertionsTests.cs
+++ b/src/Playwright.Tests/Assertions/LocatorAssertionsTests.cs
@@ -635,6 +635,25 @@ public class LocatorAssertionsTests : PageTestEx
         await Expect(locator).ToHaveJSPropertyAsync("itsNull", null);
     }
 
+    [PlaywrightTest("playwright-test/playwright.expect.misc.spec.ts", "toHaveJSProperty pass undefined")]
+    public async Task ShouldSupportToHaveJSPropertyUndefined()
+    {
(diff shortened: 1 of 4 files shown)


kblok:
Opened on the kblok fork instead: https://github.com/kblok/playwright-sharp/pull/3

PR #3356, opened 2026-09-12
fix: forward boxes option in aria snapshots

### LINKED ISSUE #3355: [Bug]: AriaSnapshotAsync ignores Boxes in Page and Locator
### Version

1.62.0; also reproduced on `main` at bda416e.

### Steps to reproduce

[Reproduction commit](https://github.com/GalayM/playwright-dotnet/commit/91b2050554fdf5e1e095644e76226f32f16fb1e4): four failing tests, without the fix.

```sh
git clone https://github.com/GalayM/playwright-dotnet.git
cd playwright-dotnet
git checkout 91b2050554fdf5e1e095644e76226f32f16fb1e4
dotnet run --project src/tools/Playwright.Tooling -- download-drivers --basepath .
dotnet build src/Playwright.Tests/Playwright.Tests.csproj -f net8.0
pwsh src/Playwright/bin/Debug/netstandard2.0/playwright.ps1 install chromium
dotnet test src/Playwright.Tests/Playwright.Tests.csproj -f net8.0 --filter ShouldIncludeBoundingBoxes
```

The tests call both `Page.AriaSnapshotAsync()` and `Locator.AriaSnapshotAsync()` with `Boxes = true`, in `Default` and `Ai` modes.

### Expected behavior

The button's snapshot includes `[box=20,30,100,40]`, as described in the [.NET API documentation](https://playwright.dev/dotnet/docs/api/class-page#page-aria-snapshot-option-boxes).

### Actual behavior

All four tests fail: the button appears, but its box is missing. In AI mode the snapshot is:

```text
- button "Submit" [ref=e2]
```

### Additional context

The option classes expose `Boxes`, but both methods omit it from the request sent to the driver.

### Environment

Windows 11 Pro, x64; .NET 8; Chromium.

GalayM:
Fixes #3355.

`Boxes = true` is currently ignored by both `Page.AriaSnapshotAsync()` and `Locator.AriaSnapshotAsync()`. Forward the option to the driver so snapshots include the documented bounding boxes.

The tests cover both methods in `Default` and `Ai` modes, check exact geometry, and verify that omitted or `false` options leave boxes out, including omission after a call with `Boxes = true`.

Also move `ResolveNpmExecutable` above the instance methods to fix the existing SA1204 failure blocking the code-style check. Its implementation is unchanged.

Tested on Windows / .NET 8:

- All four new cases fail before the fix.
- All 12 `PageAriaSnapshotTests` pass in Chromium, Firefox, and WebKit after the fix.
- `dotnet build ./src --no-restore` and `dotnet format ./src/ --verify-no-changes --no-restore` pass.


GalayM:

PR #3359, opened 2026-09-15
chore(driver): roll to 1.63.0

dgozman:
## Summary
- Roll driver to 1.63.0
- Multiple HTTP credentials (`HttpCredentialsList`)
- `Page.DialogClosed` / `BrowserContext.DialogClosed` events
- `FrameLocator()` without a selector (any frame), `Locator.Visible`
- `Opfs` option for storage state; `AriaSnapshots` / `ScreenSnapshots` tracing options
- Tracing: separate stream for API requests, recover after failed stop, HAR save failure no longer stalls context close


src/Playwright/Transport/Protocol/Generated/OPFSEntry.cs
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Playwright.Transport.Protocol;
+
+internal class OPFSEntry
```
yury-s, line 29: Is it actually used?

yury-s (approved):

src/Playwright/Transport/Protocol/Generated/OPFSEntry.cs
dgozman: No, but it is generated from the protocol automatically, so we keep it.

PR #3366, opened 2026-09-18
fix(pipelines): allow public feed access for NuGet.org publish

dgozman:
## Summary
- The release pipeline (build 15376627) failed pushing to NuGet.org: 1ES.PublishNuget@1 could not reach api.nuget.org because 1ES Network Isolation now blocks public feed endpoints by default (the "Stop Network Isolation" step reported api.nuget.org as a blocked connection).
- Set networkIsolationPolicy: Permissive on the pipeline so the release stage can still push to NuGet.org.


yury-s (approved):
I doubt it's gonna work

